PR #11, opened 2021-10-05
Bearer tokens

betolink:
Making sure we use the new bearer tokens for CMR queries.


betolink:
merging bearer token into main

PR #59, opened 2022-02-22
Fix netrc

andypbarrett:
Changed _netrc in auth.py to use netrc library
Added docstring to login


betolink:
Great work Andy! merging and resuming work in the dev branch.

PR #119, opened 2022-08-10
EDL and store improvements

betolink:
earthdata can now persist user's credentials into a `.netrc` file 

```python
from earthdata import Auth, DataCollections, DataGranules, Store

auth = Auth().login(strategy="netrc")
# are we authenticated?
if not auth.authenticated:
    # ask for credentials and persist them in a .netrc file
    auth.login(strategy="interactive", persist=True)
```

We can also renew our CMR token to make sure our authenticated queries work:

```python
auth.refresh_token()
collections = DataCollections(auth).concept_id("c-some-restricted-dataset").get()

```

and probably most importantly is that we can get authenticated `fsspec` file sessions. 

```python
store = Store(auth)

fs = store.get_https_session()
# now we can use fsspec to get any granule from any DAAC!
fs.get("https://DAAC/granule", "./data")
```

Lastly, we can stream certain datasets directly into xarray (even if we are not in AWS) 

```python
%%time 
import xarray as xr


query_results =  DataGranules().concept_id("C2036880672-POCLOUD").temporal("2018-01-01", "2018-12-31").get()
ds = xr.open_mfdataset(store.open(query_results))
ds
```


CONTRIBUTING.md
```diff
@@ -5,6 +5,40 @@ email, or any other method with the owners of this repository before making a ch
 
 Please note we have a code of conduct, please follow it in all your interactions with the project.
 
+
+## Development environment
+
+
+`earthdata` is a Python library that uses Poetry to build and publish the package to PyPi, the defacto Python repository. In order to develop new features or patch bugs etc. we need to set up a virtual environment and install the library locally. We can accomplish this with both Poetry or/and Conda.
```
jhkennedy, line 12:
```suggestion
`earthdata` is a Python library that uses Poetry to build and publish the package to PyPI, the defacto Python repository. In order to develop new features or patch bugs etc. we need to set up a virtual environment and install the library locally. We can accomplish this with both Poetry or/and Conda. 
```

Total nitpick, but this is how PyPI styles their name, see: https://pypi.org/

CONTRIBUTING.md
```diff
@@ -5,6 +5,40 @@ email, or any other method with the owners of this repository before making a ch
...
+>poetry install
+```
+
+After activating our environment and installing the library with Poetry we can run Jupyter lab and start testing the local distribution or we can use the scripts inside `scripts` to run the tests and linting. 
+Now we can create a feature branch and push those changes to our fork!
+
+
+### Using Poetry
+
+If we want to use Poetry, fir we need to [install it](). After installing Poetry we can use the same workflow we used for Conda, first we install the library locally:
```
jhkennedy, line 30:
```suggestion
If we want to use Poetry, first we need to [install it](https://python-poetry.org/docs/#installation). After installing Poetry we can use the same workflow we used for Conda, first we install the library locally:
```

earthdata/store.py
```diff
@@ -2,132 +2,266 @@
...
-        hosted on S3 by third party libraries like xarray
+        returns a list of fsspec file-like objects that can be used to access files
+        hosted on S3 or HTTPS by third party libraries like xarray.
+
         :param granules: a list of granules(DataGranule) instances or list of files
-        :param provider: cloud provider code, used when the list is made of s3 links
         :returns: a list of s3fs "file pointers" to s3 files
         """
+        print("granules need to be a list of URLs or a list of DataGranule instances")
+        return []
```
jhkennedy, line 185: I typically see an exception raised here, does returning an empty list make sense?

earthdata/store.py
```diff
@@ -2,132 +2,266 @@
...
+        total_size = round(sum([granule.size() for granule in granules]) / 1024, 2)
+        print(f" Opening {len(granules)} granules, approx size: {total_size} GB")
+
+        if self.auth is None:
             print(
-                f" Getting {len(granules)} granules, approx download size: {total_size} GB"
+                "A valid Earthdata login instance is required to retrieve credentials"
             )
-        elif isinstance(granules[0], str) and granules[0].startswith("s3"):
+            return []
```
jhkennedy, line 214: Oh, looks like in general you're not raising exceptions and returning empty for bad paths. I would assume some times an empty list is a valid return, so `None` or throwing an exception might make distinguishing the two easier.

jhkennedy (approved):
Overall this looks good to me. A couple super minor suggestions, and some thoughts on exception handling which very much could be future work and not done here.

earthdata/store.py
betolink: Yeah, it should be a `NotImplementedError` exception. Updating it, I usually don't like to rise exceptions but this one is valid.

earthdata/store.py
betolink: Changed to return `None` instead.

PR #182, opened 2022-12-06
LOVE the art badge! Changed color to blue @betolink

jules32:
Did this because red looks like fail :)


jhkennedy (approved):
Ooh I like it!

image

Issue #187, opened 2022-12-13
Document exactly why `curl` does not work by default when called from inside us-west-2 with data in AWS


yuvipanda:
If you make a request *from* inside us-west-2, the OAuth2 login process for earthdata login + redirect is a bit like this:

Original URL -> Earthdata Login OAuth2 -> Original URL (with a cookie for auth) -> Cloudfront -> *Signed S3 URL*.

The issue here is that *some clients* - notably curl, and probably aiohttp - do *something* that means S3 will reject the request with a 403 forbidden, and not much of an explanatory message. Information from Joseph H Kennedy and Luis (and others) is that the *signed S3 URL* is being rejected by AWS because it has additional authentication information (in the form of a cookie or headers) it does not understand.

So the issue is that the eventual signed s3 url *somtimes* gets two bits of authentication:

1. The signed S3 bits, part of the *query string*
2. The earthdata cookie (maybe? Or is it a header? But cookies are also headers soooo)

And S3 rejects this. This isn't manifest in all http clients, it is in some. We should use curl as a reference, and use it.

So we should figure out and document a sequence of `curl` commands, that when executed outside us-west-2, produce the data - and when executed *inside* us-west-2, do not produce the data. We can then go from here.

yuvipanda:
The end goal here is to make sure that *end users* can *start* working on data hosted on AWS without needing to hear the words 'us-west-2' or learn about the internals how how the data is stored. This will have a performance ceiling (due to not using the s3 protocol), but that's alright for a start.

yuvipanda:
Closing in favor of https://github.com/nsidc/earthaccess/issues/188, which has a ton of relevant info

Issue #167, opened 2022-10-21
Implement simpler API signatures


betolink:
The concept of collections vs granules and instantiating those classes can be confusing, especially for new users. Perhaps having a static method that can simplify things would be simpler to use e.g. (using the upcoming name)

```python
import earthaccess
import xarray as xr

auth = earthaccess.login(strategy="netrc")

granules = earthaccess.get_granules(short_name="ATL06",
                                    cloud_hosted=True,
                                    polygon=((1,2),(3,4)... ),
                                    temporal=("2020-01-01","2020-12-31"))

ds = xr.open_mfdataset(earthaccess.open(granules, auth=auth))

```

This would be better suited for regional use cases since we'll be downloading the metadata from CMR in one go. 
Workflows that may require bulk downloads could potentially use an iterator like

```python
import earthaccess

auth = earthaccess.login(strategy="netrc")
query = earthaccess.search(short_name="ATL06",
                           cloud_hosted=True,
                           page_size=2000,
                           polygon=((1,2),(3,4)... ),
                           temporal=("2020-01-01","2020-12-31"))

for granules in query.items():
    # here granules is a resultset of up to 2000 granules.
    earthaccess.download(granules, "./data/ATL06/", auth=auth)
```

andypbarrett:
I like this approach.  However, this methods requires knowing the concept id or short name first.  This is fine if you know your dataset or know to go to EarthData Search but we need to explain that step.  Otherwise getting the shortname is "magic".

betolink:
I totally agree @andypbarrett, we need to explain where this `short_name` is coming from and what's more, expose the other ways of finding data (search by DOI for example). This new methods would also abstract the fact that at granule level there is no way of querying using `cloud_hosted`, a new user needs to know the provider to make that distinction, i.e. `NSIDC_ECS` vs `NSIDC_CPRD`

scottyhq:
I was wondering if the current search keyword arguments are based on the earthdata search API? Perhaps it'd be possible to use STAC-API standards for more consistency across libraries (at least for the main spatiotemporal ones?).

I know it's nitpicky and annoying to have breaking changes like this, but the user experience is nice coming from other libraries like pystac_client / nasa-cmr-stac. and perhaps the standard names and acceptable formats could have the benefit of re-using already implemented parsers in those other libraries.

https://github.com/radiantearth/stac-api-spec/tree/main/item-search#query-parameter-table
e.g.:
```
temporal -> datetime
[...]

PR #202, opened 2023-02-16
v0.5.0

### LINKED ISSUE #195: Login fails if we use the netrc strategy and there is no netrc file present.
```python
import earthaccess 
print(f"using earthaccess version {earthaccess.__version__}")

auth = earthaccess.login(strategy="netrc")
```

The code in earthaccess/api just passes the null auth object to the store and then it breaks. A fix is in progress.

betolink:
This PR will bring some bug fixes and new capabilities to the top level API

```python
import earthaccess

auth = earthaccess.login()
```
will automatically try all strategies, there is no need to specify one, if our credentials are not found it will ask the user to provide them interactively.

 ```python
s3_credentials = earthaccess.get_s3_credentials(daac="PODAAC")
# use them with your fav library, e.g. boto3
```

```python
# another thing we can do with our auth instance is to refresh our EDL tokens
auth.refresh_tokens()
```

We can also get authenticated [fsspec](https://filesystem-spec.readthedocs.io/en/latest/api.html) sessions:

```python
url = "https://data.lpdaac.earthdatacloud.nasa.gov/lp-prod-protected/EMITL2ARFL.001/EMIT_L2A_RFL_001_20220903T163129_2224611_012/EMIT_L2A_RFL_001_20220903T163129_2224611_012.nc"

fs = earthaccess.get_fsspec_https_session()
with fs.open(lpcloud_url) as f:
    data = f.read(10)
data
```
or we can use them in tandem with xarray/rioxarray

```python 
import xarray as xr

ds = xr.open_mfdataset(earthaccess.open([url]))
ds
```

This PR will fix #195 #187 and completes #167


docs/tutorials/demo.ipynb
```diff
@@ -43,35 +43,23 @@
...
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "id": "2acb1926-3442-446a-8a35-fcbba8868d4d",
    "metadata": {},
    "outputs": [],
    "source": [
-    "from earthaccess import Auth, Store, DataCollections, DataGranules\n",
-    "auth = Auth()"
+    "auth = earthaccess.login(strategy=\"environment\")"
```
asteiker, line 62: Could we move this below the next markdown block? That way we can give the full explanation of all of the different auth strategies and then actually execute them (otherwise it's a bit out of order to start with the environment strategy, then explanation, then netrc.

asteiker, line 62: I get `Using environment variables for EDL` after executing this, but I don't believe I have those environment variables set. Is this checking that the variables exist or not?

docs/tutorials/authenticate.md
```diff
@@ -17,7 +16,7 @@ If your Earthdata Login credentials are set as environment variables: EDL_USERNA
 auth = earthaccess.login(strategy="environment")
 ```
 
-If you wish to enter your Earthdata Login credentials when prompted 
+If you wish to enter your Earthdata Login credentials when prompted
```
asteiker, line 19:
```suggestion
If you wish to enter your Earthdata Login credentials when prompted with optional persistence to .netrc
```

docs/tutorials/demo.ipynb
```diff
@@ -469,11 +432,11 @@
     "# If we get an error with direct_access=True, most likely is because\n",
```
asteiker, line 432: `direct_access=True` isn't in this code block though, so where is that coming from?

docs/tutorials/edl.ipynb
```diff
@@ -0,0 +1,98 @@
+{
```
asteiker, line 1: This is a really helpful notebook for testing. Will this get exposed in the documentation? If so, we may want to add more explanation here (how this differs from authenticate.md for example.

asteiker (changes requested):
The updated ux for the auth strategy is much improved! The only confusion I have is in the response to the generic:

`auth = earthaccess.login()`

when a netrc or interactive input is used. Because it is checking for env vars first, I receive this response even though I know the .netrc is used in my case:

```Using environment variables for EDL
You're now authenticated with NASA Earthdata Login
Using token with expiration date: 04/21/2023```

Instead, we should say (@betolink 's suggestion!):  "Using .netrc" or "Using user's input, persisting it to netrc=False" instead of "Using environment variables for EDL" which is misleading.

docs/tutorials/demo.ipynb
betolink: definitely, now the login() method will tell us which of the strategies is using for authentication.

docs/tutorials/demo.ipynb
betolink: I'm refactoring the notebook to use the top level API and this should go away!

docs/tutorials/edl.ipynb
betolink: I was thinking of adding it but maybe all these examples should be in a gallery project so they are not tied to releases.

asteiker (approved):

PR #214, opened 2023-03-10
Auth improvements

### LINKED ISSUE #212: auth.get_s3_credentials() requires a .netrc
I was experimenting with *only* using a Bearer token to authorize all requests (https://github.com/nsidc/earthaccess/issues/188) but it seems s3 credentials endpoints like `https://data.nsidc.earthdatacloud.nasa.gov/s3credentials` do not recognize this so you get re-directed to URS servers that expect basic user:password authentication:

(on a machine *without* a ~/.netrc)
```python
import earthaccess #0.5.0
auth = earthaccess.login('interactive')
auth.get_s3_credentials('NSIDC')
#You're now authenticated with NASA Earthdata Login
#Using token with expiration date: 04/22/2023
#Using user provided credentials for EDL
#https://data.nsidc.earthdatacloud.nasa.gov/s3credentials
#Authentication with Earthdata Login failed with:
#HTTP Basic: Access denied.
```

betolink:
This PR will fix #212 and implements more testing for Auth and `S3Credentials` endpoints. Eventually they are going to support bearer tokens but only ASF does at the moment. 

* Fix call to `S3Credentials`
* Fix readthedocs
* Updated example notebooks to use the new top level API
* Support EARTHDATA_USERNAME and EARTHADATA_PASSWORD same as in IcePyx (work in progress with @JessicaS11) 
* once logged in we can access our profile (and email) with 
```python
auth = earthaccess.login()

profile = auth.user_profile
email = profile["email_address"]
```


betolink:
@JessicaS11 I was thinking that we can test this branch with Icepyx, perhaps refactoring the Earthdata class to do something like

```python
import earthaccess

class Earthdata:
    """
    Initiate an Earthdata session for interactingwith the NSIDC DAAC.
    Returns
    -------
    Earthdata session object after a successful login
    """

    def __init__(self):
        self.session= None

    def login(self):
    # earthaccess will try environment variables, a netrc and lastly will ask our users for their credentials
    auth = earthaccess.login()
    if auth.authenticated:
        self.session = auth.get_requests_https_session()
        self.uid = auth.username
        self.email = auth.email
        self.pswd = auth.password
        return self.session


```


and if we don't want any breaking change we may keep the `__init__` parameters (even if we don't use them) what do you think?

earthaccess/auth.py
```diff
@@ -235,6 +264,9 @@ def _get_credentials(
                 return False
             print("You're now authenticated with NASA Earthdata Login")
             self._credentials = (username, password)
+            self.username = username
+            self.password = password
```
JessicaS11, line 268: is this to make it easier for users to get/see the username+password (given they're stored via `self._credentials`)?

betolink: I think `self._credentials` should probably go away, yes this way is to make it clearer.

earthaccess/auth.py
```diff
@@ -210,8 +235,14 @@ def _netrc(self) -> bool:
...
+            username = os.getenv("EDL_USERNAME")
+        if "EARTHDATA_USERNAME" in os.environ:
+            username = os.getenv("EARTHDATA_USERNAME")
+        if "EDL_PASSWORD" in os.environ:
+            password = os.getenv("EDL_PASSWORD")
+        if "EARTHDATA_PASSWORD" in os.environ:
+            password = os.getenv("EARTHDATA_PASSWORD")
         authenticated = self._get_credentials(username, password)
         if authenticated:
             print("Using environment variables for EDL")
```
JessicaS11, line 248: Recommend we update line 251 to EARTHDATA_USERNAME and EARTHDATA_PASSWORD, if that's the convention we'd like to encourage.

tests/integration/test_auth.py
JessicaS11, line 1: Are we still planning to add the icepyx test_Earthdata tests to this PR? It looks like it would go in integration/test_api.py, but I didn't want to start messing with that file if you already had it in the works somewhere.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+* bug fixes:
+    * get_s3_credentials() only worked when a netrc file was present, bug reported by @scottyhq and @JessicaS11
+    * including tests for all DAAC S3 endpoints
+    * Notebooks updated to use the new top level API
+
+* CI changes:
+    * documentation for readthedocs fixed by including poetry as teh default tool
```
JessicaS11, line 11:
```suggestion
    * documentation for readthedocs fixed by including poetry as the default tool
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+    * injected new secrets to test Auth using the IcePyx convention (EARTHDATA_USERNAME)
```
JessicaS11, line 12:
```suggestion
    * injected new secrets to test Auth using the icepyx convention (EARTHDATA_USERNAME)
```

earthaccess/api.py
```diff
@@ -197,7 +199,7 @@ def get_s3_credentials(daac: str, provider: str) -> Dict[str, Any]:
     return earthaccess.__auth__.get_s3_credentials(daac=daac, provider=provider)
 
 
-def collection_query(cloud_hosted: bool = True) -> Type[CollectionQuery]:
+def collection_query() -> Type[CollectionQuery]:
     """Returns a query builder instance for NASA collections (datasets)
 
     Parameters:
```
JessicaS11, line 205: suggest we remove this part of the docstring (GH won't let me do so as a suggestion)

tests/integration/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+    activate_environment()
+    auth = earthaccess.login(strategy="environment")
+    logger.info(f"Current username: {auth.username}")
+    logger.info(f"earthaccess version: {earthaccess.__version__}")
+
+    assertions.assertIsInstance(auth, earthaccess.Auth)
+    assertions.assertIsInstance(earthaccess.__auth__, earthaccess.Auth)
+    assertions.assertTrue(earthaccess.__auth__.authenticated)
+
+
```
JessicaS11, line 61:
```suggestion
```
Now that we've removed EDL_USERNAME env variables

tests/integration/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+        assertions.assertRaises(FileNotFoundError)
+        print(e_info)
+
+
+def test_auth_populates_attrs():
+    activate_environment()
+    auth = earthaccess.login(strategy="environment")
+    assertions.assertIsInstance(auth, earthaccess.Auth)
+    assertions.assertIsInstance(earthaccess.__auth__, earthaccess.Auth)
+    assertions.assertTrue(earthaccess.__auth__.authenticated)
```
JessicaS11, line 83: This test is currently identical to test_auth_can_read_edl_env_variables (line 40).

tests/integration/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_auth_returns_valid_s3_credentials():
+    activate_environment()
+    auth = earthaccess.login(strategy="environment")
+    assertions.assertIsInstance(auth, earthaccess.Auth)
+    assertions.assertIsInstance(earthaccess.__auth__, earthaccess.Auth)
+    assertions.assertTrue(earthaccess.__auth__.authenticated)
```
JessicaS11, line 91: perhaps I'm missing something - this also appears to be the same test to me.

tests/unit/test_auth.py
```diff
@@ -3,24 +3,16 @@
...
 
 
 class TestCreateAuth(unittest.TestCase):
-    @mock.patch("builtins.input")
-    @mock.patch("getpass.getpass")
-    def test_create_auth_wrong_credentials(self, user_input, user_password) -> bool:
-        user_input.return_value = "user"
-        user_password.return_value = "pass"
-        auth = Auth().login(strategy="interactive")
-        self.assertEqual(auth.authenticated, False)
```
JessicaS11, line 16: Do we not want to test a bad credentials case? Or is that effectively done in the tests below where `auth.authenticated = False`?

betolink: I just added a case for that, to test that when we get a 401 response we are not authenticated.

JessicaS11 (approved):

PR #216, opened 2023-03-16
add that requires python 3.8 or higher

jules32:


jules32:
This was noted by Andy Barrett while trying to live-install on Michele Thorton's machine during NASA Openscapes coworking (with Chris Battisto, Cassie Nickles, Celia Ou)

betolink (approved):
merging, there is a section down on the README about it but should probably be more prominent!

jules32:
Ah, sorry I'd missed that. Thank you Luis!

Issue #219, opened 2023-03-22
Consider not depending on benedict


psarka:
The python-benedict dependency is pretty heavy in terms of sub-dependencies it pulls in:

```
> conda-tree depends python-benedict
python
requests
pyyaml
python-dateutil
ftfy
mailchecker
phonenumbers
python-fsutil
python-slugify
toml
xmltodict
openpyxl
xlrd
```

I don't think you are relying on most of these, so it would be nice to replace this by something more lightweight. Should you be open to this, I may find some time to make a PR, let me know :)

betolink:
Originally benedict provided some handy methods when accessing very nested JSON objects like the ones returned by NASA's CMR but I'm not apposed to just use the regular Python dict. In fact just this week we were talking about how the results should be represented... it seems like using Pandas to create a table out of the entries from CMR resuklts would be a better way and just printing the raw json.

psarka:
In the same spirit as the benedict case, I would suggest not adding a heavy dependency (such as pandas) just for presentation.

Personally and subjectively I'm fine with raw json. The value proposition of earth access is a nice and simple API to auth and find data.

fabiocaccamo:
@psarka @betolink FYI `python-benedict` supports [optional requirements](https://github.com/fabiocaccamo/python-benedict/tree/main#optional-requirements) since version `0.29.0`.

*(I'm the author of `python-benedict`)*

betolink:
closing this as we already implemented it.

PR #229, opened 2023-04-01
Remove benedict dependency

psarka:
Following the short discussion in #219 I removed the benedict dependency, as it's features are not used (except in four lines), but it has 12 dependencies.


betolink:
Awesome! thanks for the PR, I think Iĺl have tome to review it tonight (or tomorrow).

betolink (approved):
Very clean PR, thanks for contributing! I agree with the premise, simpler is better.

Issue #253, opened 2023-06-26
Better support with remote clusters for computation


jrbourbeau:
When analyzing NASA data with [Dask](https://www.dask.org/), it's common to have a client Python session (e.g. a scientists' laptop) connect to a remote cluster that's close to where the dataset that's being analyzed is located (e.g. on AWS). In this case, file access is typically needed both on the scientists' laptop and on the cloud-based cluster (where the actual heavy computations are run). 

Here's an example notebook https://gist.github.com/jrbourbeau/44a70ee11038c0d9532b83c0a0b7bdc0 where I use `earthaccess` + `xarray` + `dask` / `coiled` to drive a computation from my laptop on a cluster that's located on AWS in `us-west-2`. This felt slow, so I then shifted my setup so that both the client and cluster were located on AWS in `us-west-2` and saw an ~10x speedup for this particular computation. 

My guess is this is due to `earthaccess`'s automatic HTTP / S3 file object switching. Today, if I use `earthaccess` to get file object from my laptop, and then forward those file objects to my cluster on AWS, they still remain HTTP file objects, even though they're being accessed directly on S3. 

One possible approach would be to have a special fsspec-compatible `EarthAccessFile` (or some other name) that better handles this HTTP / S3 file object dispatching though a custom `__reduce__` method. 

cc @betolink, who I spoke with about this offline

PR #259, opened 2023-07-08
Improved remote cluster performance

jrbourbeau:
This PR adds an `fsspec`-compatible `EarthAccessFile` object that's a very light wrapper around the current https / S3 file objects being used. The only difference is that when a file is serialized (as happens when working on a remote cluster), if we find that we are running in region on AWS and the underlying file is an https file instead of an S3 file, we reopen the file using the existing `eathaccess.open(...)` logic to get a S3 file. 

Anecdotally, I see a computation performance boost of 2x when running the notebook in https://github.com/nsidc/earthaccess/issues/253 with the changes in this PR 

Closes https://github.com/nsidc/earthaccess/issues/253

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -568,12 +568,12 @@ class Store(object):
         return results
 
     def _open_urls_https(
-        self, urls: List[str] = [], granuales=[], n_jobs: int = 8
+        self, urls: List[str] = [], granules=[], n_jobs: int = 8
     ) -> List[fsspec.AbstractFileSystem]:
         https_fs = self.get_fsspec_session()
         if https_fs is not None:
             try:
-                fileset = _open_files(urls, granuales, https_fs)
+                fileset = _open_files(urls, granules, https_fs)
             except Exception:
                 print(
                     "An exception occurred while trying to access remote files via HTTPS: "
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 from uuid import uuid4
 
+import earthaccess
 import fsspec
 import requests
 import s3fs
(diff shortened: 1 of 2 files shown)


earthaccess/store.py
```diff
@@ -538,17 +568,12 @@ def _download_onprem_granules(
         return results
 
     def _open_urls_https(
-        self, urls: List[str] = [], n_jobs: int = 8
+        self, urls: List[str] = [], granuales=[], n_jobs: int = 8
```
mrocklin, line 571: Maybe typo

```suggestion
        self, urls: List[str] = [], granules=[], n_jobs: int = 8
```

earthaccess/store.py
```diff
@@ -538,17 +568,12 @@ def _download_onprem_granules(
...
         https_fs = self.get_fsspec_session()
         if https_fs is not None:
             try:
-
-                def multi_thread_open(url: str) -> Any:
-                    return https_fs.open(url)
-
-                fileset = pqdm(urls, multi_thread_open, n_jobs=8)
-
+                fileset = _open_files(urls, granuales, https_fs)
```
mrocklin, line 576:
```suggestion
                fileset = _open_files(urls, granules, https_fs)
```

mrocklin (commented):
I can't judge the earthaccess code well, but it's cool to see that this doesn't appear to be a crazy hack 🙂

betolink:
This is amazing! @jrbourbeau I'll take a look today. On the tests failing, I have some WIP to only run integration tests on main. They are not deterministic and some datasets require EULAs, also when you run them from a fork secrets cannot be shared.

jrbourbeau:
Okay, thanks @betolink. I figured it was something like that based of the failures over in https://github.com/nsidc/earthaccess/pull/260. FWIW https://github.com/nsidc/earthaccess/pull/261 should fix the documentation build

MattF-NSIDC:
This looks ready to merge to us; @betolink did some local testing and things worked great. We added a comment about an issue with data from the CMR API that we currently don't know how to resolve: The first "data link" is not always the "right" data link.

We're going to punt on the failing tests because Luis found the unit tests are passing, but issues with secrets are causing the integration tests to fail. Another PR for skipping integration tests on forks is coming soon :)

@jrbourbeau as this PR still has [WIP] in the title, are you planning to do more work on work on it? If not we can merge it tomorrow!

mrocklin:
@jrbourbeau is on vacation this week.  Last I spoke with him I recall him saying positive things about this PR though, and saying that mostly it was waiting review.  I wouldn't be surprised if he just forgot to remove the WIP.

MattF-NSIDC (approved):

MattF-NSIDC:
Great! Thank you for your patience and contribution :)

PR #260, opened 2023-07-09
Test CI

jrbourbeau:
I'm seeing what I think are unrelated CI failures over in https://github.com/nsidc/earthaccess/pull/259. This PR runs CI against the current `main` branch to confirm whether or not that's the case.


jrbourbeau:
Turns out I needed to make a small change in order to trigger the test / documentation GHA workflows to run. Just added a space in an exception message that's raised.

Issue #262, opened 2023-07-21
`earthaccess.download` fails for on-prem access


andypbarrett:
If I follow the example in the readme to search for NSIDC-0051 granules and download them, I get an `AttributeError: 'NoneType' object has no attribute 'get'`.

```
In [1]: import earthaccess
In [2]: auth = earthaccess.Auth()

In [3]: results = earthaccess.search_data(short_name='NSIDC-0051', temporal=('1978-11', '1980-12'), bounding_box=(-180, 0, 180, 90))
Granules found: 788

In [5]: files = earthaccess.download(results, local_path='/home/apbarret/Data/seaice/nsidc-0051/netcdf')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 files = earthaccess.download(results, local_path='/home/apbarret/Data/seaice/nsidc-0051/netcdf')

File ~/anaconda3/envs/test-seaice-regrid/lib/python3.11/site-packages/earthaccess/api.py:167, in download(granules, local_path, provider, threads)
    147 def download(
    148     granules: Union[List[earthaccess.results.DataGranule], List[str]],
    149     local_path: Optional[str],
    150     provider: Optional[str] = None,
    151     threads: int = 8,
    152 ) -> List[str]:
    153     """Retrieves data granules from a remote storage system.
    154 
    155        * If we run this in the cloud, we will be using S3 to move data to `local_path`
   (...)
    165         List of downloaded files
    166     """
--> 167     results = earthaccess.__store__.get(granules, local_path, provider, threads)
    168     return results

AttributeError: 'NoneType' object has no attribute 'get'
```

```
In [10]: earthaccess.__version__
Out[10]: '0.5.2'
```

andypbarrett:
Solved.  I missed that `earthaccess.download` requires `earthaccess.login` to be executed and not `earthaccess.Auth()`.

`earthaccess.login` sets credentials and instantiates a Store object.

Suggest that a try, except is used
```
try:
      results = earthaccess.__store__.get(granules, local_path, provider, threads)
except AttributeError as err:
      print(err)
      print("earthaccess.login() must be called before attempting to download data")
      return None
```

PR #263, opened 2023-07-21
add try block to return useful error message for failed download

andypbarrett:
This PR address #262.

Following the api workflow in the readme, if `earthaccess.login()` is not executed before download, download fails with an ambiguous `AttributeError`.  To fix this, I added a `try-except` block to `download` in `api.py`, so that the user is advised to run `earthdata.login()` before attempting to download and `None` is returned.

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -164,7 +164,13 @@ def download(
     Returns:
         List of downloaded files
     """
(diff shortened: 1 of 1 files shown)


betolink (approved):
Thanks Andy! maybe down the road we should handle the error in the store class so when that happens is not the API that sounds the alarm.

Issue #299, opened 2023-09-13
Automatically run `earthaccess.login()`?


jrbourbeau:
I've run into a few times where either I, or a user I was working with, forgot to run `earthaccess.login()` and get an error like

```python
File "/opt/coiled/env/lib/python3.10/site-packages/earthaccess/api.py", line 184, in open
    results = earthaccess.__store__.open(granules=granules, provider=provider)

AttributeError("'NoneType' object has no attribute 'open'")
```

I think a better user experience would be to automatically try `earthaccess.login()` with the `netrc` and `environment` strategies. That way, at least for users who have a `.netrc` file or have intentionally set the right environment variables, `earthaccess.login()` isn't needed. This removes the "I forgot to run `earthaccess.login()`" issue and cuts down on boilerplate code (at least by 1 line 🙂).

@betolink @MattF-NSIDC I'm curious what you think about this

betolink:
Sounds like a good idea, are you thinking on executing it at import time? I'm planning on working on a PR for this paired with "too verbose" output that you mentioned on #283

jrbourbeau:
> are you thinking on executing it at import time?

My thinking is avoiding a long import time would be nice, as along as the code needed to do things on-demand isn't super complicated

> I'm planning on working on a PR for this paired with "too verbose" output that you mentioned on https://github.com/nsidc/earthaccess/issues/283

Ah, that'd be great. FWIW I played around a little with the automatic login stuff already. This seems to work for on-demand login:

```diff
diff --git a/earthaccess/__init__.py b/earthaccess/__init__.py
index ef4fdfb..dd248b0 100644
--- a/earthaccess/__init__.py
+++ b/earthaccess/__init__.py
@@ -36,7 +36,28 @@ __all__ = [
     "Store",
 ]

-__auth__ = Auth()
-__store__: Any = None
-
 __version__ = version("earthaccess")
+
+_auth = None
+_store = None
+
+def __getattr__(name):
+
+    global _auth, _store
+
+    if name == "__auth__" or name == "__store__":
+        if _auth is None or not _auth.authenticated:
+            for strategy in ["environment", "netrc"]:
+                try:
+                    _auth = Auth().login(strategy=strategy)
+                except Exception:
+                    continue
+                else:
+                    if not _auth.authenticated:
+                        continue
+                    else:
+                        _store = Store(_auth)
+                        break
+        return _auth if name == "__auth__" else _store
+    else:
[...]

Issue #283, opened 2023-08-18
`earthaccess.login()` has lots of output


jrbourbeau:
Running `earthaccess.login()` has the following output:

```python
In [1]: import earthaccess

In [2]: earthaccess.login()
EARTHDATA_USERNAME and EARTHDATA_PASSWORD are not set in the current environment, try setting them or use a different strategy (netrc, interactive)
You're now authenticated with NASA Earthdata Login
Using token with expiration date: 10/01/2023
Using .netrc file for EDL
Out[2]: <earthaccess.auth.Auth at 0x10769cfd0>
```

I could be totally off, but this seems like a lot of information for a successful login. In particular the 

```
EARTHDATA_USERNAME and EARTHDATA_PASSWORD are not set in the current environment, try setting them or use a different strategy (netrc, interactive)
```

could be possibly concerning for someone who's new to `earthaccess`. Maybe they see this message and think something went wrong. 

I'm curious what folks here think about having less verbose output in this case like

```python
In [1]: import earthaccess

In [2]: earthaccess.login()
You have successfully authenticated with NASA Earthdata
Out[2]: <earthaccess.auth.Auth at 0x10769cfd0>
```

or maybe even no output if the login is successful 

This is a relatively minor issue, though one I thought was still worth raising.

MattF-NSIDC:
>no output if the login is successful

:100: 

The rest feels like it should be debug output! And the order seems kind of weird. "Using netrc" seems like it should have been printed before "Success".

Issue #231, opened 2023-04-13
AWS us-west-2 checking method


battistowx:
In some of our DAAC notebooks, we include the following Boto3 snippet to check if the notebook is being executed inside us-west-2, and throws a ValueError (with emojis) if you are not, preventing the notebook from being fully executed:

```
if (boto3.client('s3').meta.region_name == 'us-west-2'):
    display(Markdown('### us-west-2 Region Check: &#x2705;'))
else:
    display(Markdown('### us-west-2 Region Check: &#10060;'))
    raise ValueError('Your notebook is not running inside the AWS us-west-2 region, and will not be able to directly access NASA Earthdata S3 buckets')
```

It may be useful to include a method that the user can call to check if they are in us-west-2 for direct S3 access and will throw an error like this, possibly using an fsspec transitive dependency, or the existing authorization checks.

battistowx:
Found methods to use requests to grab the current instance's region from the locally-linked Instance Identity Document, and it works great! https://gist.github.com/doublenns/7e3e4b72df4aaeccbeabf87ba767f44e

Wondering if this should be its own function that a user could call anywhere, or if it should be called only when an S3 granule is opened, which would print a detailed exception statement.

JessicaS11:
I'd be interested in helping implement this in some fashion - we're updating related functionality in icepyx and I'd love to just put it upstream and then use it.

betolink:
I think it would be great if we can expose `earthaccess.__store__.in_region` to `earthaccess.in_region(default="us-west-2")` and perhaps using an nicely formatted __repr__ output besides the boolean value?

JessicaS11:
Could someone clarify for #424:
> was the spirit to check specifically for us-west-2, or to enable the user to see what region they are running in?

Thanks!

chuckwondo:
Perhaps a silly question, but why do we need to make such an explicit check? I ran across another thread somewhere (I can't find it at the moment) where it seems that this is a hard problem (impossible in some cases?) to get such a check to work reliably/confidently in various environments.

What specifically is such a check to help a user do or not do or avoid?

jhkennedy:
A lot of this has been discussed in #444. 

[...]

PR #300, opened 2023-09-14
Support automatic authentication

jrbourbeau:
Closes https://github.com/nsidc/earthaccess/issues/299, closes https://github.com/nsidc/earthaccess/issues/283


betolink (commented):
I like this! I wonder if we should have a __repr__ method to show the users if they are authenticated and if they are in us-west-2 to incorporate what @battistowx suggested in #231

jrbourbeau:
I think that's a reasonable thing to add. Where would the repr method go? Maybe a new `earthaccess.info` or some better name 

FWIW, while I like the repr idea, my sense is it's probably separate from the automatic login logic here (though let me know if you think otherwise)

earthaccess/auth.py
```diff
@@ -231,29 +234,29 @@ def _netrc(self) -> bool:
...
         except FileNotFoundError as err:
-            print(f"No .netrc found in {os.path.expanduser('~')}")
-            raise err
+            raise FileNotFoundError(
+                f"No .netrc found in {os.path.expanduser('~')}"
+            ) from err
         except NetrcParseError as err:
-            print("Unable to parse .netrc")
-            raise err
+            raise NetrcParseError("Unable to parse .netrc") from err
```
mfisher87, line 241: Are we re-raising because the messages from the library aren't that great? If we don't need to do any handling, can we remove the try/except?

Or was the idea to have log messages for cases when tracebacks are disabled? In which case, can we `logger.error` these messages?

mfisher87:
>FWIW, while I like the repr idea, my sense is it's probably separate from the automatic login logic here (though let me know if you think otherwise)

I'm inclined to agree here!

earthaccess/__init__.py
```diff
@@ -36,7 +37,30 @@
...
-__store__: Any = None
-
 __version__ = version("earthaccess")
+
+_auth = Auth()
+_store = None
+_lock = threading.Lock()
+
+
+def __getattr__(name):  # type: ignore
```
mfisher87, line 47: I like this. I think we could use a docstring.

earthaccess/auth.py
mfisher87: These exceptions would be swallowed [here](https://github.com/nsidc/earthaccess/pull/300/files#diff-28208126471c4332a472c11d7d8d595053abc2031e4b38b660753a28f79ba7acR56-R57) so I think  we should `logger.error` these before re-raising. What do you all think?

earthaccess/__init__.py
```diff
@@ -36,7 +37,30 @@
...
+        with _lock:
+            if not _auth.authenticated:
+                for strategy in ["environment", "netrc"]:
+                    try:
+                        _auth.login(strategy=strategy)
+                    except Exception:
+                        continue
+                    else:
+                        if not _auth.authenticated:
+                            continue
```
mfisher87, line 60: The "environment" strategy will currently log at debug level in this case, which would be triggered if envvars are not set. The "netrc" strategy, however, [will not](https://github.com/jrbourbeau/earthaccess/blob/641eafcf1ae7807cd15323502eb7dab47133ee11/earthaccess/auth.py#L248-L250). Should we have more debug logging on this functionality, since it's happening implicitly?

Maybe we should debug log here "Automatic login attempted with {strategy=} but did not succeed"?  It couldn't hurt to debug log in the `except` case, too. And maybe also in the successfully-authenticated with `{strategy}` case?

mfisher87:
I think this is my last thought: how should this change impact the documentation?

earthaccess/auth.py
jrbourbeau: I don't fully recall, but yeah, I think this was to make the message that was actually raised to the user more meaningful

earthaccess/__init__.py
jrbourbeau: Added 👍

earthaccess/__init__.py
jrbourbeau: Went ahead and added more debug logging here. I think there's enough information now to help figure out where things go wrong (if they do go wrong) but let me know if there are some cases I missed

jrbourbeau:
> how should this change impact the documentation?

I went ahead and updated the authentication section of the docs -- let me know what you think

earthaccess/__init__.py
MattF-NSIDC: I forget when module-level getattrs were added to Python, but I vaguely feel like it wasn't that long ago. My sense of time is often a complete mess :) I think the answer is "no", but just want to double-check: are there any Python versions (that are not EOL) that don't support this?

earthaccess/__init__.py
MattF-NSIDC: I think it looks great, thanks James :)

earthaccess/__init__.py
jrbourbeau: No worries. It was added in Python 3.7 https://peps.python.org/pep-0562 so we should be good

earthaccess/auth.py
MattF-NSIDC: >These exceptions would be swallowed [here](https://github.com/nsidc/earthaccess/pull/300/files#diff-28208126471c4332a472c11d7d8d595053abc2031e4b38b660753a28f79ba7acR56-R57) so I think we should logger.error these before re-raising. What do you all think?

The link "here" is a perma-link, but it doesn't look like it points to what I thought it would. I thought these exceptions were going to be swallowed at the time I wrote the comment, likely I meant to link to [here](https://github.com/nsidc/earthaccess/pull/300/files#diff-28208126471c4332a472c11d7d8d595053abc2031e4b38b660753a28f79ba7acR65-R68), which didn't have debug logging at the time. I think this is resolved, please un-resolve if you catch any misunderstandings :)

MattF-NSIDC (approved):
Looks great to me! I don't have the same level of familiarity with the library as @betolink so you may want a second opinion ;)

mfisher87:
:100:

PR #303, opened 2023-09-20
Add DOI to README and CITATION.cff

mfisher87:


betolink (approved):
another cool badge, looks great!

Issue #307, opened 2023-09-26
Downloads for AU_SI12_NRT_R04 incorrect


trey-stafford:
I am trying to download granules of [AU_SI12_NRT_R04](https://cmr.earthdata.nasa.gov/search/concepts/C1886605827-LANCEAMSR2.html#) using `earthaccess.download` but the results are incorrect. Files are created on disk but they do not seem to contain the data.

```
import earthaccess

results = earthaccess.search_data(short_name='AU_SI12_NRT_R04')
results = sorted(results, key=lambda x: x['meta']['revision-date'], reverse=True)
earthaccess.login()
files = earthaccess.download(results, "/tmp/test")
```
```
Granules found: 14
You're now authenticated with NASA Earthdata Login
Using token with expiration date: 11/25/2023
Using environment variables for EDL
 Getting 14 granules, approx download size: 0.0 GB
QUEUEING TASKS | : 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 28/28 [00:00<00:00, 2188.52it/s]
PROCESSING TASKS | : 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 28/28 [00:02<00:00, 10.15it/s]
COLLECTING RESULTS | : 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 28/28 [00:00<00:00, 160877.41it/s]
```

This results in files of ~4.1K in size in the indicated `/tmp/test` directory. I expect files ~126M in size:
```
$ ls -lah /tmp/test/
-rw-rw-r--  1 trst2284 trst2284 4.1K Sep 26 13:15 AMSR_U2_L3_SeaIce12km_P04_20230926.he5
-rw-rw-r--  1 trst2284 trst2284 4.1K Sep 26 13:15 AMSR_U2_L3_SeaIce12km_R04_20230913.he5
-rw-rw-r--  1 trst2284 trst2284 4.1K Sep 26 13:15 AMSR_U2_L3_SeaIce12km_R04_20230914.he5
-rw-rw-r--  1 trst2284 trst2284 4.1K Sep 26 13:15 AMSR_U2_L3_SeaIce12km_R04_20230915.he5
-rw-rw-r--  1 trst2284 trst2284 4.1K Sep 26 13:15 AMSR_U2_L3_SeaIce12km_R04_20230916.he5
[...]

PR #308, opened 2023-09-26
Small store fixups

trey-stafford:
Couple of minor fixups that I noticed while looking at `earthaccess.store`:

* Fixup spelling mistake in `access` variable assignment (`direc` -> `direct`)
* Pass `threads` arg to `_open_urls_https` in `_open_urls`, replacing hard-coded value of 8.

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -426,7 +426,7 @@ class Store(object):
                     "We cannot open S3 links when we are not in-region, try using HTTPS links"
                 )
                 return None
-            fileset = self._open_urls_https(data_links, granules, 8)
+            fileset = self._open_urls_https(data_links, granules, threads)
             return fileset
 
     def get(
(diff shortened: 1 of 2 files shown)


earthaccess/store.py
```diff
@@ -533,7 +533,7 @@ def _get_granules(
         provider = granules[0]["meta"]["provider-id"]
         endpoint = self._own_s3_credentials(granules[0]["umm"]["RelatedUrls"])
         cloud_hosted = granules[0].cloud_hosted
-        access = "direc" if (cloud_hosted and self.running_in_aws) else "external"
+        access = "direct" if (cloud_hosted and self.running_in_aws) else "external"
```
MattF-NSIDC, line 536: Thinking out loud: I think it'd be really useful to add type safety here. We could type `access` as Enum or as a union of literal string types.

trey-stafford:
I got sidetracked last week and didn't have a chance to come back to this and my other problem that's described in https://github.com/nsidc/earthaccess/issues/307 until now.

I thought I might come across other items that may end up in this PR, but since I'm just getting re-started and I'm not certain of that, I want to make sure these small fixes get merged.

jrbourbeau (approved):
Generally this looks good to me. @betolink @MattF-NSIDC any additional thoughts?

MattF-NSIDC:
:rocket:

PR #316, opened 2023-10-13
Add utility for getting authentication environment variables

jrbourbeau:
When working with multiple machines, it's often easier to set authentication environment variables than it is to ship a `.netrc` file around. This PR adds a small utility for programmatically retrieving `EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`

### DIFF
--- a/earthaccess/__init__.py
+++ b/earthaccess/__init__.py
@@ -13,6 +13,7 @@ from .api import (
     open,
     search_data,
     search_datasets,
+    auth_environ,
 )
 from .auth import Auth
 from .search import DataCollections, DataGranules
@@ -34,6 +35,7 @@ __all__ = [
     "DataCollections",
     "Auth",
     "Store",
+    "auth_environ",
 ]
 
 __auth__ = Auth()
--- a/earthaccess/__init__.py
+++ b/earthaccess/__init__.py
@@ -2,6 +2,7 @@ from importlib.metadata import version
 from typing import Any
 
 from .api import (
+    auth_environ,
     collection_query,
     download,
     get_fsspec_https_session,
@@ -13,7 +14,6 @@ from .api import (
     open,
     search_data,
     search_datasets,
-    auth_environ,
 )
 from .auth import Auth
 from .search import DataCollections, DataGranules
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -331,8 +331,11 @@ def get_edl_token() -> str:
     token = earthaccess.__auth__.token
     return token
 
+
 def auth_environ() -> Dict[str, str]:
     auth = earthaccess.__auth__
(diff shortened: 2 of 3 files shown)


MattF-NSIDC:
Thinking out loud: I think in this case, the envvar names make sense because Earthdata credentials are likely the same regardless of which application uses them, but I have a habit of namespacing my envvars by prefixing them with an application name, e.g. `EARTHACCESS_EDL_USERNAME` or something.

This is different than the case of an envvar like `PORT` which unequivocally should be `MY_SERVICE_PORT` instead, because many applications need _different_ ports set; I'm not suggesting we do this, just looking for others' thoughts on the interface.

MattF-NSIDC (approved):
Looks great.

betolink:
Looks neat, how is this different from what `earthaccess.login()` does? the first thing this method does is trying to authenticate with environment variables. If we are using a cluster and we send the `__store__` instance, it's already authenticated no?

jrbourbeau:
> Looks neat, how is this different from what earthaccess.login() does?

This utility is for, after a Python session is already authenticated, easily grabbing environment variables needed to authenticate other machines. For example, if I have a `.netrc` file locally, then `earthaccess.login()` works great. If I then want to provision a bunch of remote machines (that won't have my local `.netrc` file) and call `earthaccess.download()` on them, I'll get authentication errors. Often you can pretty easily set environment variables when provisioning remote machines, so having a little utility for getting the necessary env vars `earthaccess` needs can be useful. 

Here's a specific example of how I'd like to use this with Coiled (though the code being added here isn't Coiled-specific, it could be used by any other library):

```python
import tempfile
import earthaccess
import coiled

# I have a `~/.netrc` file locally on my laptop where this code is running
earthaccess.login()

@coiled.function(
    region="us-west-2",
    environ=earthaccess.auth_environ(),  # This makes sure the remove VMs being spun up here have the right `EARTHDATA_*` variables set 
)
def process(granule):
    with tempfile.TemporaryDirectory() as tmpdir:
        files = earthaccess.download(granule, tmpdir)
        ... # Do things with these files

granules = earthaccess.search_data(...)
results = process.map(granules)
```

> If we are using a cluster and we send the __store__ instance, it's already authenticated no?

That's correct, we just don't do that in all cases (e.g. calling `earthaccess.download` on a list of granules doesn't result in us sending the `__store__` instance)

betolink:
Ooops I merged in the wrong order @jrbourbeau

jrbourbeau:
No worries -- just resolved the merge conflict

betolink:
I think we should merge this one, what do you think @MattF-NSIDC?

MattF-NSIDC:
:rocket:

mfisher87 (approved):

PR #317, opened 2023-10-13
Minor `earthaccess.download` updates

jrbourbeau:
This PR contains a few ergonomic updates to `earthaccess.download`. Specifically:

1. `earthaccess.download` now accepts a single granule as input in addition to a list of granules
2. The files returned by `earthaccess.download` are now the fully qualified local paths. This makes it easier to do things with the paths that are returned (currently there is some extra `os.path.join` that needs to happen)
```python
files = earthaccess.download(...)
for file in files:
    with open(file) as f:
        ...
```
3. Technically `local_path` can be `None` when using `earthaccess.download`. We weren't accounting for that possibility in all the possible `download` cases (e.g. downloading from S3). This PR moves the `local_path=None` logic higher up in the call stack to account for these cases.

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -9,6 +9,7 @@ from .auth import Auth
 from .search import CollectionQuery, DataCollections, DataGranules, GranuleQuery
 from .store import Store
 from .utils import _validation as validate
+from .results import DataGranule
 
 
 def search_datasets(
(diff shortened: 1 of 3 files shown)


betolink:
This is great! one of the things that confuses new users is that we only worked with list of results, thanks @jrbourbeau !!

jrbourbeau:
> one of the things that confuses new users is that we only worked with list of results

Me too : )

betolink (approved):

PR #320, opened 2023-10-17
Fix ``test_auth_environ_raises integration`` test

jrbourbeau:
This is currently failing on `main` with `FAILED tests/integration/test_api.py::test_auth_environ_raises - Failed: DID NOT RAISE <class 'RuntimeError'>`  (for example, see [this CI build](https://github.com/nsidc/earthaccess/actions/runs/6538911512/job/17763073681)) due to the new automatic authentication update (xref https://github.com/nsidc/earthaccess/pull/300). 

This PR update this test accordingly so it passes while still keeping the original intent of the test

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -91,8 +91,9 @@ def test_auth_environ():
 
 
 def test_auth_environ_raises(monkeypatch):
-    # Temporarily ensure `earthaccess.__auth__` always returns a
-    # new, unauthenticated `earthaccess.Auth`` instance
+    # Ensure `earthaccess.__auth__` always returns a new,
+    # unauthenticated `earthaccess.Auth` instance, bypassing
+    # automatic auth behavior
     monkeypatch.setattr(earthaccess, "__auth__", earthaccess.Auth())
 
     # Ensure `earthaccess.auth_environ()` raises an informative error
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -91,6 +91,16 @@ def test_auth_environ():
 
 
 def test_auth_environ_raises(monkeypatch):
+    # Temporarily remove any existing authentication and make it so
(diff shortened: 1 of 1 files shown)


tests/integration/test_api.py
```diff
@@ -91,6 +91,11 @@ def test_auth_environ():
 
 
 def test_auth_environ_raises(monkeypatch):
-    monkeypatch.setattr(earthaccess.__auth__, "authenticated", False)
+    # Temporarily ensure `earthaccess.__auth__` always returns a
+    # new, unauthenticated `earthaccess.Auth`` instance
```
MattF-NSIDC, line 95:
```suggestion
    # Ensure `earthaccess.__auth__` always returns a new,
    # unauthenticated `earthaccess.Auth`` instance, bypassing
    # automatic auth behavior
```

MattF-NSIDC (approved):
LGTM, but I felt the comment could use a touch more detail :)

jrbourbeau:
> Any idea why this didn't fail on the feature branch prior to merge?

Integration tests aren't run on PRs, only pushes to `main`

https://github.com/nsidc/earthaccess/blob/d0386aebcb0715395c883c3c30930df00c96df3b/.github/workflows/integration-test.yml#L3-L11

MattF-NSIDC:
OT: Any opinions on setting up integration tests to also run on `workflow_dispatch` or in response to a comment in a PR? Would be good for certain changes to be able to manually trigger the integration tests on a feature branch.

MattF-NSIDC:
Great points! To be clear, the event type which protects secrets from forks is `pull_request`. If we trigger on a PR comment, the event type would be `issue_comment`, and we should be very careful that our workflow checks the commenter is a repository member before running the workflow. The triggering user would also have to be aware of the importance of reviewing the PR for malicious code before triggering the workflow. That last part makes this far from an ideal choice :)

Issue #274, opened 2023-08-06
Documentation builds not triggered when README.md, CONTRIBUTING.md, LICENCE.txt altered


mfisher87:
These files are symlinked into the `docs` directory, and our workflow is triggered by changes in the `docs` directory. GitHub Actions does not recognize a change to `README.md` as a change to the symlink `docs/index.md`, so if the README is changed, docs don't build. I tested this on a fork.

https://github.com/mfisher87/earthaccess/pull/4

jrbourbeau:
Circling back here. I think the recent readthedocs work closes this issue as now we get a documentation build + preview on all PRs. I'm going to close this issue as fixed, but feel free to re-open is that's not the case @MattF-NSIDC

mfisher87:
I think we need to clean up and simplify some GHA YAML:

- Ignore [Pull Request event](
https://github.com/nsidc/earthaccess/blob/6fbc4ae4146c399b14b9ec7e7b2333d94cb12830/.github/workflows/documentation.yml#L14-L21) - RTD does this
- Update `paths` list for push events - we want GHA to match RTD when `README.md` is changed, for example.
- Simplify [this conditional](https://github.com/nsidc/earthaccess/blob/6fbc4ae4146c399b14b9ec7e7b2333d94cb12830/.github/workflows/documentation.yml#L60-L63), should just be checking if `branch == main|documentation` - the rest is vestigial IIRC

mfisher87:
I'm opening a quick PR so we don't have to open anything else to track that :)

PR #324, opened 2023-10-21
Cleanup vestigial GH Actions YAML for documentation

mfisher87:
See #274 for more

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,14 +11,8 @@ on:
       - docs/**
       - notebooks/**
       - .github/workflows/documentation.yml
-  pull_request:
-    paths:
-      - earthaccess/**
-      - tests/**
-      - docs/**
-      - notebooks/**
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks! I'm curious if we need this workflow at all anymore? Maybe we just stick with readthedocs for all documentation builds?

mfisher87:
Normally I'd say let's put it all in one place, but in this case I'm not sure. We've already shared out links to the earthaccess documentation. @betolink @asteiker thoughts?

betolink (commented):
If I'm reading this correctly, we are not deploying the documentation for PRs?

mfisher87:
ReadTheDocs is deploying docs for PRs. If you click "Show all checks" for this PR you can see the successful build. We didn't need PR previews on both RTD _and_ GitHub Pages, so the original intent of this PR was to remove the redundancy.

Now that we've decided that GH Pages will redirect to RTD, we no longer need this PR.

Issue #341, opened 2023-11-09
Replace Flake8 with Ruff


mfisher87:

jhkennedy:
:+1: I'll second this one; I'm in the process of switch over over most of our repositories from flake8 to ruff.

jrbourbeau:
+1 from me too -- @jhkennedy do you have bandwidth to switch over here too?

jhkennedy:
@jrbourbeau I'm out tomorrow but could probably hit it next week. I'll assign this to me for now :-D

Issue #282, opened 2023-08-16
Consider using ``pre-commit`` hooks for linting


jrbourbeau:
Sometimes it's easy to forget to lint code (I do this all the time!) and then CI builds fail because, for example, `black` isn't happy. We might consider using `pre-commit` hooks here to handle code formatting / linting. That way things like `mypy`, `black`, etc. all run automatically when a developer runs `git commit ...`

@betolink @MattF-NSIDC thoughts? 

### TODO

- [x] Remove `.scripts/lint.sh` and references to it
- [x] Add Ruff (or Flake8) hook to `.pre-commit-config.yaml` (add Ruff config to pyproject.toml)
- [x] Enable pre-commit.ci (I may do this independently)
- [x] Move isort config from script into a file (pyproject.toml; or migrate to Ruff) _Note: now using ruff and config is in `pyproject.toml`_
- [x] Move Black config from script to file _Note: now using ruff and config is in `pyproject.toml`_
- [x] Move Mypy config from script to file. _Note: now in `pyproject.toml`_
- [x] Should we run Mypy with pre-commit? It can be slow (probably not yet on this project?), and pre-commit is meant to not slow you down. _Resolution: No, it'ss run as part of our unit tests now._
- [x] ~~More pre-commit hooks?~~ Lets open specific issues/PRs if we want them.
- [x] Make sure `pre-commit` is a dev dependency (if it isn't already -- haven't checked) 
- [x] Update `CONTRIBUTING.md` to include `pre-commit install`

MattF-NSIDC:
1000 times yes. Pre-commit gets my tools out of my way and I'm never going back :heart_eyes:  Not everyone will use it, so I think we should keep the CI checks... and maybe run them in pre-commit.ci?

mfisher87:
I think we should go forward with this. While we're there, I'd like to remove Flake8 and isort to replace with Ruff. Ruff also has a Black-compatible formatter that I haven't tried yet, so I would probably still leave Black in place for now?

I think we need to:

- [ ] Remove `.scripts/lint.sh` and references to it
- [ ] Enable pre-commit.ci (I may do this independently)
- [ ] Move isort config from script into a file (pyproject.toml; or migrate to Ruff)
- [ ] Move Black config from script to file
- [ ] Move Mypy config from script to file
- [ ] ??

jrbourbeau:
- [ ] Make sure `pre-commit` is a dev dependency (if it isn't already -- haven't checked) 
[...]

PR #342, opened 2023-11-09
Migrate tool configs to pyproject.toml, configure pre-commit.ci

mfisher87:
Pytest and Mypy can both be configured by `pyproject.toml`. 

[Flake8 can't](https://github.com/PyCQA/flake8/issues/234), but Ruff can. I prefer Ruff anyway, so I opened #341.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,17 @@
+ci:
+  autoupdate_schedule: "monthly"  # Like dependabot
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autofix_prs: false  # Comment "pre-commit.ci autofix" on a PR to trigger
+
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
-    hooks:
-    -   id: check-yaml
-    -   id: trailing-whitespace
-    -   id: check-toml
-    -   id: check-json
--   repo: https://github.com/psf/black
-    rev: 20.8b1
-    hooks:
-    -   id: black
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+    - id: check-yaml
+    - id: trailing-whitespace
+    - id: check-toml
+    - id: check-json
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+    - id: black
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,17 @@ build-backend = "poetry.masonry.api"
 filterwarnings = ["error::UserWarning"]
 
 
+[tool.mypy]
+disallow_untyped_defs = false
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = [
+  "tests.*",
+]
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -78,6 +78,21 @@ requires = ["poetry>=0.12"]
 build-backend = "poetry.masonry.api"
 
 
+[tool.pytest]
+filterwarnings = ["error::UserWarning"]
```
jrbourbeau, line 82: I realize this is just moving an existing setting around, but I'm curious what this is actually doing. Maybe causing `UserWarning`s to be elevated to errors?

jrbourbeau (commented):
Thanks @mfisher87! Overall the changes here look sensible to me. It looks like linting / testing didn't run here due to various restrictions we have in those GHA workflows. We should probably update those to include changes to more files. FWIW I would actually be okay always running linting and tests.

pyproject.toml
mfisher87: It's based on [warnings.simplefilter](https://docs.python.org/3/library/warnings.html#warnings.simplefilter) it seems. Your speculation was spot on :)

mfisher87:
>It looks like linting / testing didn't run here due to various restrictions we have in those GHA workflows. We should probably update those to include changes to more files

Hm, yeah... good eye. I wonder when did this start happening? It looks like we're not running pre-commit checks in CI at all? Maybe for another thread, but I've had a good user experience with and would like to use pre-commit.ci!

jrbourbeau:
> It looks like we're not running pre-commit checks in CI at all?

We're running linting as part of the test script here

https://github.com/nsidc/earthaccess/blob/e9c68ed6e496cead17bf5403542bbdf10c16088e/scripts/test.sh#L6-L7

But I agree, this isn't obvious at first 

> Maybe for another thread, but I've had a good user experience with and would like to use pre-commit.ci!

I'm also +1 for pre-commit (xref https://github.com/nsidc/earthaccess/issues/282)

mfisher87:
Ah, thanks! I expected that they would be run through pre-commit so didn't think to search for direct invocations. I'll continue the conversation about linter changes over there.

.pre-commit-config.yaml
```diff
@@ -1,12 +1,17 @@
+ci:
+  autoupdate_schedule: "monthly"  # Like dependabot
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autofix_prs: false  # Comment "pre-commit.ci autofix" on a PR to trigger
+
```
jrbourbeau, line 5: I've not used pre-commit.ci before. I'm assuming this is correct and you'll handle whatever signup / authentication is needed to hook this repo into pre-commit.ci

jrbourbeau (approved):
Thanks @mfisher87

.pre-commit-config.yaml
mfisher87: Yes, need to flip a switch on my other GitHub account. It will be just a minute...

MattF-NSIDC:
It may not be watching this PR yet, but let's try...

pre-commit.ci run

MattF-NSIDC:
Let's call fixes out of scope. New PR incoming.

PR #347, opened 2023-11-10
Allow single file URL inputs for ``earthaccess.download``

jrbourbeau:
Similar to https://github.com/nsidc/earthaccess/pull/317 but for a single `https://...` URL instead of a single data granule

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Type, Union, cast
+from typing import Any, Dict, List, Optional, Type, Union
 
 import earthaccess
 import requests
@@ -170,8 +170,10 @@ def download(
     Returns:
         List of downloaded files
     """
-    if isinstance(granules, (DataGranule, str)):
-        granules = cast(Union[List[DataGranule], List[str]], [granules])
+    if isinstance(granules, DataGranule):
+        granules = [granules]
+    elif isinstance(granules, str):
+        granules = [granules]
     try:
         results = earthaccess.__store__.get(granules, local_path, provider, threads)
     except AttributeError as err:
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Type, Union, cast
 
 import earthaccess
 import requests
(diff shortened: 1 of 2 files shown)


earthaccess/api.py
```diff
@@ -162,16 +162,16 @@ def download(
...
         provider: if we download a list of URLs we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
 
     Returns:
         List of downloaded files
     """
-    if isinstance(granules, DataGranule):
-        granules = [granules]
+    if isinstance(granules, (DataGranule, str)):
+        granules = cast(Union[List[DataGranule], List[str]], [granules])
```
jrbourbeau, line 174: This `cast` is a little messy, but is to make `mypy` happy. I'm not a typing expert -- definitely open to other suggestions that make this less involved.

MattF-NSIDC: This one feels like it should be easy :laughing: It's such a common pattern for a function to accept one or many of a thing. I'm sure there's a term-of-art of this, but my memory and google skills are failing me.

This method avoids telling the typechecker "trust us" so I think it's a bit better, but it's also still pretty messy. With the combined `isinstance` check, Mypy sees `thing` as `str | Thing`, but then after listing it Mypy `list[str | Thing]` instead of what I expected expected, `list[str] | list[thing]`.

```python
class Thing:
    pass


def foo(things: Thing | str | list[Thing] | list[str]) -> list[str] | list[Thing]:
    thing_list: list[str] | list[Thing]
    if isinstance(things, Thing): 
        thing_list = [things]
    elif isinstance(things, str):
        thing_list = [things]

    return thing_list
```

MattF-NSIDC: I'm going to share this out and see if anyone wants to play code golf.

MattF-NSIDC: An entry using a generic function and switch statement: https://gist.github.com/kbeamnsidc/4ffa3708677fa5b36623071443a194c7

tests/integration/test_api.py
```diff
@@ -69,13 +69,18 @@ def test_granules_search_returns_valid_results(kwargs):
...
+@pytest.mark.parametrize("use_url", [True, False])
+def test_download(tmp_path, selection, use_url):
     results = earthaccess.search_data(
         count=2,
         short_name="ATL08",
         cloud_hosted=True,
         bounding_box=(-92.86, 16.26, -91.58, 16.97),
     )
+    if use_url:
+        # Download via file URL instead of granule
```
MattF-NSIDC, line 81:
```suggestion
        # Download via file URL string instead of DataGranule object
```

jrbourbeau:
I'll plan to merge this PR in a few hours if there are no further comments. It should be a relatively non-controversial change and I'm happy to own any follow-up work, if there is any.

mfisher87:
I'd like to avoid the `cast`. What do you think of the suggestion above? More verbose than I'd like, but I think avoiding `cast`s is more important than succinctness.

jrbourbeau:
I don't really care either way. Just updated to avoid the cast

PR #351, opened 2023-11-10
Raise exceptions more

jrbourbeau:
There are a few places where we're currently printing a message and returning `None`, `[]`, etc. when something goes wrong. It's nice that there's typically an informative message, but it relatively easy to ignore these messages (guilty myself), and can lead to unexpected downstream errors (e.g. downstream code is expecting a list of strings, but you give it `None` instead).

Here's an example

```python
In [5]: result = earthaccess.open(["notavalidprotocol://myfile.nc"])
Schema for notavalidprotocol://myfile.nc is not recognized, must be an HTTP or S3 URL

In [6]: type(result)
Out[6]: NoneType
```

This PR proposes we raise exceptions in these cases (still with an informative message) to properly signal "Hey, something went wrong here". This should make it more clear where things are going awry. 

With the changes in this PR, the above example looks like this instead 

```python
In [2]: result = earthaccess.open(["notavalidprotocol://myfile.nc"])
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 result = earthaccess.open(["notavalidprotocol://myfile.nc"])

File ~/projects/nsidc/earthaccess/earthaccess/api.py:197, in open(granules, provider)
    184 def open(
    185     granules: Union[List[str], List[earthaccess.results.DataGranule]],
    186     provider: Optional[str] = None,
    187 ) -> List[AbstractFileSystem]:
    188     """Returns a list of fsspec file-like objects that can be used to access files
    189     hosted on S3 or HTTPS by third party libraries like xarray.
    190
   (...)
    195         a list of s3fs "file pointers" to s3 files.
    196     """
--> 197     results = earthaccess.__store__.open(granules=granules, provider=provider)
    198     return results

File ~/projects/nsidc/earthaccess/earthaccess/store.py:281, in Store.open(self, granules, provider)
    272 """Returns a list of fsspec file-like objects that can be used to access files
    273 hosted on S3 or HTTPS by third party libraries like xarray.
    274
   (...)
    278     a list of s3fs "file pointers" to s3 files.
    279 """
    280 if len(granules):
--> 281     return self._open(granules, provider)
    282 return []

File ~/mambaforge/envs/earthaccess-dev/lib/python3.9/site-packages/multimethod/__init__.py:315, in multimethod.__call__(self, *args, **kwargs)
    313 func = self[tuple(func(arg) for func, arg in zip(self.type_checkers, args))]
    314 try:
--> 315     return func(*args, **kwargs)
    316 except TypeError as ex:
    317     raise DispatchError(f"Function {func.__code__}") from ex

File ~/projects/nsidc/earthaccess/earthaccess/store.py:383, in Store._open_urls(self, granules, provider, threads)
    381     data_links = granules
    382 else:
--> 383     raise ValueError(
    384         f"Schema for {granules[0]} is not recognized, must be an HTTP or S3 URL"
    385     )
    386 if self.auth is None:
    387     raise ValueError(
    388         "A valid Earthdata login instance is required to retrieve S3 credentials"
    389     )

ValueError: Schema for notavalidprotocol://myfile.nc is not recognized, must be an HTTP or S3 URL
```

Which seems more like what I'd expect.

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -230,10 +230,9 @@ class Store(object):
                 self.initial_ts = datetime.datetime.now()
             return deepcopy(self.s3_fs)
         else:
-            print(
+            raise ValueError(
                 "A valid Earthdata login instance is required to retrieve S3 credentials"
             )
-            return None
 
     @lru_cache
     def get_fsspec_session(self) -> fsspec.AbstractFileSystem:
@@ -269,7 +268,7 @@ class Store(object):
         self,
         granules: Union[List[str], List[DataGranule]],
         provider: Optional[str] = None,
-    ) -> Union[List[Any], None]:
+    ) -> List[Any]:
         """Returns a list of fsspec file-like objects that can be used to access files
         hosted on S3 or HTTPS by third party libraries like xarray.
 
@@ -280,15 +279,14 @@ class Store(object):
         """
         if len(granules):
             return self._open(granules, provider)
-        print("The granules list is empty, moving on...")
-        return None
+        return []
 
     @singledispatchmethod
     def _open(
         self,
         granules: Union[List[str], List[DataGranule]],
         provider: Optional[str] = None,
-    ) -> Union[List[Any], None]:
+    ) -> List[Any]:
         """Returns a list of fsspec file-like objects that can be used to access files
         hosted on S3 or HTTPS by third party libraries like xarray.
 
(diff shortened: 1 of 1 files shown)


MattF-NSIDC:
:100: :100: :100:

I'm going to have to look closer on non-work time. Thank you for digging in to this. This is much needed.

MattF-NSIDC:
I wanted to look over the weekend, but still struggling to find the time. We could probably use another maintainer :laughing:

betolink (approved):
These changes look good to me!

Issue #353, opened 2023-11-14
Not sensitive to temporary S3 credentials expiring


jrbourbeau:
I tried to download some granules in a Python session that had been up for a while (over an hour)

```python
files = earthaccess.download(granules, "test_data")
```

and got this error due to the temporary S3 credentials that `earthaccess` was using being expired (full traceback further below):

```python
PermissionError: The provided token has expired.
```

<details>
<summary>Full traceback:</summary>

```python
---------------------------------------------------------------------------
ClientError                               Traceback (most recent call last)
File /opt/coiled/env/lib/python3.11/site-packages/s3fs/core.py:706, in S3FileSystem._lsdir(self, path, refresh, max_items, delimiter, prefix, versions)
    705 files = []
--> 706 async for c in self._iterdir(
    707     bucket,
    708     max_items=max_items,
    709     delimiter=delimiter,
    710     prefix=prefix,
    711     versions=versions,
    712 ):
    713     if c["type"] == "directory":

File /opt/coiled/env/lib/python3.11/site-packages/s3fs/core.py:756, in S3FileSystem._iterdir(self, bucket, max_items, delimiter, prefix, versions)
    749 it = pag.paginate(
    750     Bucket=bucket,
    751     Prefix=prefix,
   (...)
    754     **self.req_kw,
    755 )
--> 756 async for i in it:
    757     for l in i.get("CommonPrefixes", []):

File /opt/coiled/env/lib/python3.11/site-packages/aiobotocore/paginate.py:30, in AioPageIterator.__anext__(self)
     29 while True:
---> 30     response = await self._make_request(current_kwargs)
     31     parsed = self._extract_parsed_response(response)

File /opt/coiled/env/lib/python3.11/site-packages/aiobotocore/client.py:383, in AioBaseClient._make_api_call(self, operation_name, api_params)
    382     error_class = self.exceptions.from_code(error_code)
--> 383     raise error_class(parsed_response, operation_name)
    384 else:

ClientError: An error occurred (ExpiredToken) when calling the ListObjectsV2 operation: The provided token has expired.

The above exception was the direct cause of the following exception:

PermissionError                           Traceback (most recent call last)
Cell In[29], line 1
----> 1 files = earthaccess.download(granules, "test_data")
      2 print(f"{files = }")

File /opt/coiled/env/lib/python3.11/site-packages/earthaccess/api.py:176, in download(granules, local_path, provider, threads)
    174     granules = [granules]
    175 try:
[...]

Issue #349, opened 2023-11-10
Error downloading S3 file from SEA_SURFACE_HEIGHT_ALT_GRIDS_L4_2SATS_5DAY_6THDEG_V_JPL2205 dataset


jrbourbeau:
Downloading HTTPS files with 

```python
import earthaccess

results = earthaccess.search_data(
    short_name="SEA_SURFACE_HEIGHT_ALT_GRIDS_L4_2SATS_5DAY_6THDEG_V_JPL2205", count=10
)
files = results[0].data_links(access="indirect")
out = earthaccess.download(files, "testdownload")
print(f"{out = }")
```
works as expected with output 

```
Granules found: 2207
QUEUEING TASKS | : 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 1156.09it/s]
PROCESSING TASKS | : 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:03<00:00,  3.45s/it]
COLLECTING RESULTS | : 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 19972.88it/s]
out = ['testdownload/ssh_grids_v2205_1992101012.nc']
```

When I then switch to downloading S3 URLs directly (making sure to run in `us-west-2` too)

```python
import earthaccess

results = earthaccess.search_data(
    short_name="SEA_SURFACE_HEIGHT_ALT_GRIDS_L4_2SATS_5DAY_6THDEG_V_JPL2205", count=10
)
files = results[0].data_links(access="direct")
out = earthaccess.download(files, "testdownload")
print(f"{out = }")
```

I get the following output (and no files downloaded):

```
Granules found: 2207
earthaccess can't yet guess the provider for cloud collections, we need to use one from earthaccess.list_cloud_providers()
out = None
```

it looks like we're having trouble inferring the data provider, so if I then manually specify `provider="PODAAC"`

```python
import earthaccess

results = earthaccess.search_data(
    short_name="SEA_SURFACE_HEIGHT_ALT_GRIDS_L4_2SATS_5DAY_6THDEG_V_JPL2205", count=10
)
files = results[0].data_links(access="direct")
[...]

PR #354, opened 2023-11-14
Handle S3 credential expiration more gracefully

jrbourbeau:
Closes https://github.com/nsidc/earthaccess/issues/353


earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
+                "A valid Earthdata login instance is required to retrieve S3 credentials"
+            )
+        if not any([concept_id, daac, provider, endpoint]):
+            raise ValueError(
+                "At least one of the concept_id, daac, provider or endpoint"
+                "parameters must be specified. "
+            )
+
+        # Get existing S3 credentials if we already have them
+        location = (concept_id, daac, provider, endpoint)
```
jrbourbeau, line 213: Previously we were always returning the same `s3fs.S3FileSystem` instance, regardless of if the `provider=`, etc. has been changed. We now cache S3 credentials based on a unique `(concept_id, daac, provider, endpoint)` combination, which should be more correct.

tests/unit/test_store.py
```diff
@@ -74,40 +85,39 @@ def test_store_can_create_s3_fsspec_session(self):
...
-            self.assertEqual(type(s3_fs), type(fsspec.filesystem("s3")))
+            assert isinstance(s3_fs, s3fs.S3FileSystem)
+            assert s3_fs.storage_options == expected_storage_options
 
         for provider in [
             "NSIDC_CPRD",
             "POCLOUD",
             "LPCLOUD",
-            "ORNLCLOUD",
+            "ORNL_CLOUD",
```
jrbourbeau, line 114: The previous `ORNLCLOUD` provider should actually be `ORNL_CLOUD` but was covered up in the previous implementation `get_s3fs_session` implementation always returning the same `s3fs.S3FileSystem` instance.

jrbourbeau:
@MattF-NSIDC @betolink I confirmed the changes here work as expected in practice (i.e. no longer getting the permissions errors reported in the https://github.com/nsidc/earthaccess/issues/353)

MattF-NSIDC:
It may realistically be a while before I can review this! :cry:

earthaccess/store.py
```diff
@@ -97,8 +96,9 @@ def __init__(self, auth: Any, pre_authorize: bool = False) -> None:
         """
         if auth.authenticated is True:
             self.auth = auth
-            self.s3_fs = None
-            self.initial_ts = datetime.datetime.now()
+            self._s3_credentials: Dict[
+                Tuple, Tuple[datetime.datetime, Dict[str, str]]
+            ] = {}
```
mfisher87, line 101: I'd like to specify the key `Tuple` a little more strictly if possible. Is it always a 4-tuple? Might be worth having a type alias (`Location = Tuple[Something, SomethingElse, AnotherSomething, SomethingAgain]`). Or maybe better: an annotated namedtuple?

```python
class Location(typing.NamedTuple):
   concept_id: Something
   daac: SomethingAgain
   provider: AnotherSomething
   endpoint: SomethingElse
```

Opinion, not request: I feel giving names to the remaining types may also help with readability, i.e. what can we name `Dict[str, str]` to make it more clear what its purpose is?

earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
+
+        # Get existing S3 credentials if we already have them
+        location = (concept_id, daac, provider, endpoint)
+        need_new_creds = False
+        try:
+            dt_init, creds = self._s3_credentials[location]
+        except KeyError:
+            need_new_creds = True
+        else:
+            delta = datetime.datetime.now() - dt_init
```
mfisher87, line 220:
```suggestion
            # If cached credentials are expired, invalidate the cache
            delta = datetime.datetime.now() - dt_init
```

earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
+            raise ValueError(
+                "At least one of the concept_id, daac, provider or endpoint"
+                "parameters must be specified. "
+            )
+
+        # Get existing S3 credentials if we already have them
+        location = (concept_id, daac, provider, endpoint)
+        need_new_creds = False
+        try:
+            dt_init, creds = self._s3_credentials[location]
```
mfisher87, line 216:
```suggestion
            # Check for cached credentials
            dt_init, creds = self._s3_credentials[location]
```

earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
+            if round(delta.seconds / 60, 2) > 55:
+                need_new_creds = True
+                self._s3_credentials.pop(location)
+
+        if need_new_creds:
+            # Don't have existing valid S3 credentials, so get new ones
+            now = datetime.datetime.now()
             if endpoint is not None:
-                s3_credentials = self.auth.get_s3_credentials(endpoint=endpoint)
+                creds = self.auth.get_s3_credentials(endpoint=endpoint)
```
mfisher87, line 229: Weak nit: I kind of prefer longer variable names like `s3_credentials` :) What was the rationale for the name change?

earthaccess/store.py
mfisher87: A brief comment about why all four of these things is needed would be useful here, I think. 

Naively, `concept_id`'s presence is a bit confusing to me since we're deriving the value of `provider` from it below (actually getting the credentials doesn't depend on `concept_id`). Could we reduce the size of the tuple (and elminate some of the if...elif chain below) if we did that derivation in advance if `provider is None`?

earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
-            # TODO: test this mocking the time or use https://github.com/dbader/schedule
-            # if we exceed 1 hour
-            if (
-                self.s3_fs is None or round(delta_minutes.seconds / 60, 2) > 59
-            ) and s3_credentials is not None:
-                self.s3_fs = s3fs.S3FileSystem(
-                    key=s3_credentials["accessKeyId"],
-                    secret=s3_credentials["secretAccessKey"],
-                    token=s3_credentials["sessionToken"],
-                )
```
mfisher87, line 229: I really like the new `try`/`catch`/`else` alternative to this logic. Much more readable.

earthaccess/store.py
```diff
@@ -200,40 +199,49 @@ def get_s3fs_session(
...
-                self.initial_ts = datetime.datetime.now()
-            return deepcopy(self.s3_fs)
-        else:
-            print(
-                "A valid Earthdata login instance is required to retrieve S3 credentials"
-            )
-            return None
```
mfisher87, line 236: I'm curious... how was this previously passing type checking?

tests/unit/test_store.py
```diff
@@ -74,40 +85,39 @@ def test_store_can_create_s3_fsspec_session(self):
...
             status=200,
         )
 
         store = Store(self.auth)
         self.assertTrue(isinstance(store.auth, Auth))
         for daac in ["NSIDC", "PODAAC", "LPDAAC", "ORNLDAAC", "GES_DISC", "ASF"]:
             s3_fs = store.get_s3fs_session(daac=daac)
-            self.assertEqual(type(s3_fs), type(fsspec.filesystem("s3")))
+            assert isinstance(s3_fs, s3fs.S3FileSystem)
+            assert s3_fs.storage_options == expected_storage_options
```
mfisher87, line 103: Nice! Do we still need to inherit from `unittest.TestCast` now that we're using assert statements?

mfisher87 (dismissed):
Great work! I have a couple thoughts about readability, but no concerns about the behavior at the moment. 

I don't think the suggestions I made are significant enough to warrant a second review, so if you're ready to merge after going through them: :rocket:

earthaccess/store.py
mfisher87: Hmm. Is there a better way to write suggestions to avoid the false deletion? The raw suggestion markdown is:

``````
```suggestion
            # If cached credentials are expired, invalidate the cache
            delta = datetime.datetime.now() - dt_init
```
``````

**Comment by betolink** [comment]
related to this, just for reference, Alex Mandel from DevSeed mentioned this piece of code that handles more or less the same for the MAAP project https://github.com/MAAP-Project/gedi-subsetter/blob/58c87ab907bc3b121510eeabdeb2b8e63c7df999/src/gedi_subset/maapx.py#L55-L61

File: earthaccess/store.py (hunk 5 above)
Comment (jrbourbeau) [reply earthaccess/store.py:216]: I think this context is already provided with the 

```python
# Get existing S3 credentials if we already have them
```

comment a few lines above

earthaccess/store.py
jrbourbeau: I also like more descriptive variable names when there's not much other context around, but when there is sufficient context, more succinct names tend to reduce the cognitive load on readers (at least in my personal opinion). 

In this case this method is all about S3, and this variable is coming from a method called `get_s3_credentials`, so my sense is there's enough context to know that `creds` are S3 credentials.

earthaccess/store.py
jrbourbeau: No idea 🤷 But I definitely see that we're returning `None` here but the type signature says this should be an `s3fs` file system

tests/unit/test_store.py
jrbourbeau: We're still using `unittest` style code elsewhere, for example `self.assertTrue(isinstance(store.auth, Auth))` a few line above

earthaccess/store.py
jrbourbeau: Given this is (at least currently) only used in one place, and connected to a private attribute, adding additional typing infrastructure like this seems like overkill to me. I'd prefer to wait until this is used in multiple places before adding extra objects that code readers need to know about/keep track of. (Maybe this point is loosely related to the `daac` vs `provider` conversation over in https://github.com/nsidc/earthaccess/issues/349)

jrbourbeau:
> mentioned this piece of code that handles more or less the same for the MAAP project

Nice, thanks for pointing us to that @betolink. Good to know for future reference.

jrbourbeau:
@mfisher87 I've addressed or responded to all your review comments (thank you for taking the time to review this PR). There are a few minor, subjective points where I think we have differing opinions / styles -- let me know if any of those are blocking comments.

earthaccess/store.py
MattF-NSIDC: Makes total sense to me, thanks for reasoning that out.

MattF-NSIDC:
@jrbourbeau thanks! I don't feel any of those differing opinions / styles should be blockers, marked all as resolved! I think this is ready to go :100:

MattF-NSIDC (approved):

Issue #357, opened 2023-11-21
Zero granules are being reported for datasets with restricted access


danielfromearth:
### Challenge
When using `earthaccess` to search for "restricted" datasets — i.e., those that require an Earthdata login (EDL) to be part of an Access Control List (ACL) — zero granules are always being reported (`"granules found: 0"`) even though there are multiple granules in NASA's Common Metadata Repository (CMR).

### Desired functionality
When using `earthaccess` to search for "restricted" datasets, the correct number of granules should be reported when the user's EDL belongs to the appropriate ACL.

### Example
I encountered this challenge when searching for TEMPO data that is restricted based on Early Adopter and Science Team group memberships, and @betolink informed me that the same issue was encountered when searching for test ICESat-2 data. It is not possible to show a reproducible example that would work for everyone, because it depends on the EDL's access settings.

### Benefit
If restricted datasets can be made to work with `earthaccess`, its usability would be expanded to include early dataset testing.

betolink:
Thanks @danielfromearth ! I think this has been a known issue for `earthaccess` and the fix would be to override the `hit()` method in the DataGranule class to use the authenticated sessions. I can work on this or if you like you can also open a PR. Thanks for reporting this!

mfisher87:
Resolved by #358

PR #358, opened 2023-11-21
Feature/issue 357

danielfromearth:
Addresses #357.

## Description

This change adds a `hits()` method to the `DataGranule` class, which uses the authenticated session. This overrides the `hits()` method from the `Query` base class, which does not use an authenticated session.

The reason for this change is to enable `earthaccess` to correctly determine and report the number of granules for "restricted" datasets, which are collections that only permit access if a user's authenticated Earthdata login (EDL) belongs to an Access Control List (ACL).

## Local test steps
Ran `earthaccess.search()` on a TEMPO data collection that is not public but for which my EDL belonged to an ACL, and confirmed that an appropriate non-zero number of granules was returned. And I was able to subsequently work with the data via `earthaccess.open()` and `xarray` functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [unreleased]
+* Fix zero granules being reported for restricted datasets
+
 ## [v0.7.1] 2023-11-08
 * Bug Fixes:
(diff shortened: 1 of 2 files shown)


betolink (approved):
Really nice! Thanks for the PR @danielfromearth !

Issue #356, opened 2023-11-20
boto3 1.21.32 requires botocore<1.25.0,>=1.24.32, but you have botocore 1.31.64 which is incompatible.


yitai000:
install result:
Installing collected packages: botocore, earthaccess
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
boto3 1.21.32 requires botocore<1.25.0,>=1.24.32, but you have botocore 1.31.64 which is incompatible.
Successfully installed botocore-1.31.64 earthaccess-0.7.0

there is an error when import earthaccess:
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Input In [7], in <cell line: 1>()
----> 1 import earthaccess

File ~\AppData\Roaming\Python\Python39\site-packages\earthaccess\__init__.py:6, in <module>
      3 from importlib.metadata import version
      4 from typing import Any
----> 6 from .api import (
      7     auth_environ,
      8     collection_query,
      9     download,
     10     get_fsspec_https_session,
     11     get_requests_https_session,
     12     get_s3_credentials,
     13     get_s3fs_session,
     14     granule_query,
     15     login,
     16     open,
     17     search_data,
     18     search_datasets,
     19 )
     20 from .auth import Auth
     21 from .search import DataCollections, DataGranules

File ~\AppData\Roaming\Python\Python39\site-packages\earthaccess\api.py:5, in <module>
      3 import earthaccess
      4 import requests
----> 5 import s3fs
      6 from fsspec import AbstractFileSystem
      8 from .auth import Auth

File ~\AppData\Roaming\Python\Python39\site-packages\s3fs\__init__.py:1, in <module>
----> 1 from .core import S3FileSystem, S3File
      2 from .mapping import S3Map
      4 from ._version import get_versions

File ~\AppData\Roaming\Python\Python39\site-packages\s3fs\core.py:29, in <module>
     27 import aiobotocore
     28 import botocore
---> 29 import aiobotocore.session
     30 from aiobotocore.config import AioConfig
     31 from botocore.exceptions import ClientError, HTTPClientError, ParamValidationError

File ~\AppData\Roaming\Python\Python39\site-packages\aiobotocore\session.py:12, in <module>
      3 from botocore.session import (
      4     EVENT_ALIASES,
      5     ServiceModel,
   (...)
      8     copy,
      9 )
     11 from . import retryhandler
[...]

PR #361, opened 2023-11-22
Update issue manager with "feedback requested" behavior

mfisher87:
We weren't using an "answered" label, so I just replaced it.

Triggers updated per documentation!

### DIFF
--- a/.github/workflows/issue-manager.yml
+++ b/.github/workflows/issue-manager.yml
@@ -17,7 +17,7 @@ jobs:
   issue-manager:
     runs-on: "ubuntu-latest"
     steps:
-    - uses: "tiangolo/issue-manager@master"
+    - uses: "tiangolo/issue-manager@0.4.0"
       with:
         token: "${{ secrets.GITHUB_TOKEN }}"
         config: >
--- a/.github/workflows/issue-manager.yml
+++ b/.github/workflows/issue-manager.yml
@@ -2,20 +2,28 @@ name: "Issue Manager"
 
 on:
   schedule:
-  - cron: "0 0 * * *"
+    - cron: "0 0 * * *"
+  issue_comment:
+    types:
+      - "created"
+  issues:
+    types:
+      - "labeled"
+  pull_request_target:
+    types:
+      - "labeled"
 
 jobs:
   issue-manager:
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-latest"
     steps:
-    - uses: tiangolo/issue-manager@master
+    - uses: "tiangolo/issue-manager@master"
       with:
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: "${{ secrets.GITHUB_TOKEN }}"
         config: >
             {
-              "answered": {
-                "users": ["betolink"],
+              "feedback requested": {
                 "delay": 864000,
-                "message": "Assuming the original issue was solved, it will be automatically closed now. But feel free to add more comments or create new issues."
+                "message": "Closing after 10 days of waiting for feedback. If you feel this was in error, please re-open, `@` a maintainer, or create new issues."
               }
             }


mfisher87:
I specifically want to use this functionality for #356

.github/workflows/issue-manager.yml
```diff
@@ -2,20 +2,28 @@ name: "Issue Manager"
...
+    types:
+      - "labeled"
+  pull_request_target:
+    types:
+      - "labeled"
 
 jobs:
   issue-manager:
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-latest"
```
mfisher87, line 18: I feel strongly that quoting YAML strings is important. There's too many weird things that can happen if you don't. Just like bash :) Is there a pre-commit validator that can check for this? :)

jhkennedy: GitHub's own documentation doesn't quote simple YAML strings, except for `${{}}` contexts, e.g.,
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on

Other than a few special cases (number-y strings and special characters) I've not had any issues with unquoted strings in YAML. So, I feel it's entirely unnecessary and would prefer to match GitHub's documentation examples as much as possible.

mfisher87: I don't think this is something with an objectively correct answer, but I also don't really consider GitHub an authority on YAML quality, so I'm not terribly swayed in my opinion by that :) One of the strange thing YAML does is convert the value `no` (the word "no" or the country code for Norway) to `False`. String rules in YAML are very complicated. I think "explicit is better than implicit", and "simple is better than complex" are my guiding principles in this preference.

If you feel strongly against adding the quotes, I don't think we're going to resolve that. I'm happy to restore this to a mix of quoted and unquoted stings as it was before so we can get this merged.

mfisher87: Following up on the earlier thought about validation, there is a yamllint rule that can enforce always-quoted strings (single or double), never enforce quotes, or permit quotes only when needed (for all the special cases like `no`, `yes`, `true`, `false`, `#`, `[]`, `{}`, `\d.*`... probably more I'm not aware of). If/when we have consensus on this issue, I think it'd be great to record that decision as a linting rule.

https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings

jhkennedy: I don't think my preference is strong enough to warrant a standoff :rofl: 

I've just rarely hit this issue in practice, and almost always with trailing zeros on number-y things. So I see a molehill here, but if you've climbed a mountain and planted a flag, that's cool with me -- we'll go with that.

I do like the idea of making a decision and letting the linter-bot be the sheriff, though!

jhkennedy (dismissed):

.github/workflows/issue-manager.yml
```diff
@@ -2,20 +2,28 @@ name: "Issue Manager"
...
     steps:
-    - uses: tiangolo/issue-manager@master
+    - uses: "tiangolo/issue-manager@master"
```
jhkennedy, line 20:
```suggestion
    - uses: "tiangolo/issue-manager@0.4.0"
```
It's probably worth pinning to the latest release and using dependabot to keep the action up to date by adding `github-actions`  to it like:
https://github.com/ASFHyP3/hyp3-sdk/blob/develop/.github/dependabot.yml

.github/workflows/issue-manager.yml
mfisher87: >I do like the idea of making a decision and letting the linter-bot be the sheriff, though!

Yes! Having software solve these problems is much better than dealing with competing preferences. There are lots of usually-not-a-big-deal things people can have strong opinions about, and for me quoting strings in YAML and Bash is one of them. I've been bitten too many times, and "just quote strings" keeps my cognitive load low :) But I don't want things that help me to cause harm to anyone else, so... yeah, let's make this a discussion about linters/formatters :)

.github/workflows/issue-manager.yml
mfisher87: Agree, good catch.

mfisher87:
Looks to be working:

```
INFO:root:Not clossing issue: #356 as the delay hasn't been reached: 10 days, 0:00:00
INFO:root:Finished
```

https://github.com/nsidc/earthaccess/actions/runs/6965375261/job/18953956336

PR #380, opened 2023-11-30
Roll up dependabot GHA updates into one PR

MattF-NSIDC:
If anyone approves, please also merge :)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,10 +10,14 @@ updates:
     schedule:
       interval: "monthly"
     groups:
-      all-dependencies:
+      pip-dependencies:
         patterns:
           - "*"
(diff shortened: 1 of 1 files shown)


jhkennedy (approved):
I *didn't* do this because I find dependency issues much easier to catch and work through in isolated PRs. With a monthly cadence and the lower number of actions compared to pip dependencies, I didn't think this would be a significant burden outside of when we first turn it on. 

That said, I'm probably going to be the least-frequent dependabot reviewer/merger^ so I'm going to approve and leave it up to you. 

---

^My team likes daily cadence and individual PRs and that's **entirely too much** so I've effectively been trained now to ignore dependabot PRs and let *them* handle it.

PR #391, opened 2023-12-02
Enable AWS check with IMDSv2

### LINKED ISSUE #390: the module level Store does not know it IS in AWS
The top level `earthaccess` module has a default `earthaccess.store.Store` used by the `earthaccess.open` function to process the results of a granule query, and its behavior depends in very important ways on the value of its `running_in_aws` attribute.
```
import earthaccess
earthaccess.__store__.running_in_aws
```
On https://openscapes.2i2c.cloud, I'm getting `False`.

As a result, `earthaccess.open` returns file-likes opened with https rather than s3.
```
granules = earthaccess.granule_query().concept_id("C2153549818-NSIDC_CPRD").get(1)
earthaccess.open(granules)
```
```
[<File-like object HTTPFileSystem, https://data.nsidc.earthdatacloud.nasa.gov/nsidc-cumulus-prod-protected/GLAS/GLAH12/034/2003/2/GLAH12_634_1102_001_0071_0_01_0001.H5>]
```

Though I am not familiar with earthaccess source or EC2 [instance metadata service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html), it seems fairly obvious that IMDSv2 introduced a breaking change that requires a fix. You now gotta provide a token 🙄. Or just ask for one. PR in a sec.

itcarroll:
Fixes #390 with a small change to the check for instance metadata query. For IMDSv2, the first request must be a PUT to "http://169.254.169.254/latest/api/token". A Response OK on that seems like a sufficient check.

Tests ... I would love to add a test. I have no idea how a test that depends on running on an EC2 instance could be set up. If you would like to provide guidance, happy to tackle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+* Bug fixes:
+    * Follow Amazon EC2 IMDSv2 requirements to set `running-in-aws`
+
 ## [v0.8.1] 2023-12-01
 * New Features:
     * Add `kerchunk` metadata consolidation utility.
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -140,8 +140,10 @@ class Store(object):
         session = self.auth.get_session()
         try:
             # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
-            resp = session.get(
-                "http://169.254.169.254/latest/meta-data/public-ipv4", timeout=1
+            resp = session.put(
+                "http://169.254.169.254/latest/api/token",
+                headers={"X-aws-ec2-metadata-token-ttl-seconds": "21600"},
+                timeout=1,
             )
         except Exception:
             return False


jrbourbeau (commented):
Thanks @itcarroll. This LGTM. My understanding is that IMDSv2 is broadly available and in some cases required for security reasons. There may be some super old instance types where only IMDSv1 is available, but I'm not certain about that, and it also seems like enough of an edge case that I'm okay with the changes you have here (would need to add extra logic to try IMDSv1 is IMDSv2 fails). That said, I'd appreciate an extra +1 from someone else who is familiar with this code path. 

Re: testing, I've manually tested this locally and on `us-west-2` using Coiled and confirmed it's working as expected. I don't think we have any automated AWS testing at the moment.

MattF-NSIDC:
>would need to add extra logic to try IMDSv1 is IMDSv2 fails

I was going to suggest using boto3 to check, because it must implement this, right? https://github.com/boto/boto3/issues/313 :(

itcarroll:
> Could you elaborate a bit here? I'm curious what advantages that package has over making a direct https request

Well, I thought one advantage is that it was aware of both IMDS versions, but I now see that the package docs actually say that it exclusively uses IMDSv2. In that case, the minor advantage is that the ec2-metadata may be more likely to stay current with future breaking changes. E.g. they updated to IMDSv2 in 2020. That's not really convincing though.

I will keep the requests package, but make the test be for us-west-2.

itcarroll:
@jrbourbeau I guess #395 makes this obsolete?

jrbourbeau (approved):
Thanks @itcarroll. Let's go ahead without an additional +1 for the changes here. I'm happy with them and we can always follow-up as needed if others have thoughts

Issue #393, opened 2023-12-04
`open()` is not handling multi-file granules properly


betolink:
I think when we started using the EarthAccesFile wrapper we stopped handling multi-file granules, this example with data from HLS shows the issue

```python
import earthaccess

auth = earthaccess.login()

granules = earthaccess.search_data(
    short_name="HLSL30",
    count=1
)
# HSL is a multispectral dataset and each granule has many files
print(granules[0].data_links())
# earthaccess is only opening the first link
files = earthaccess.open(granules)
files
```

There is a workaround by manually collecting the links and then opening them but we are missing that sweet auto-wriring from search to access.

jrbourbeau:
Thanks @betolink. Let me push up a quick PR real quick for this one to get some thoughts

jrbourbeau:
@betolink I'm able to reproduce the issue with `open()` (see https://github.com/nsidc/earthaccess/pull/394 for a fix) but not with `download()`

```python
In [1]: import earthaccess
   ...:
   ...: auth = earthaccess.login()
   ...:
   ...: granules = earthaccess.search_data(
   ...:     short_name="HLSL30",
   ...:     count=1
   ...: )
   ...: # HSL is a multispectral dataset and each granule has many files
   ...: print(granules[0].data_links())
   ...: # earthaccess is only opening the first link
   ...: files = earthaccess.download(granules, "foo")
   ...: files
Granules found: 11072415
[...]

PR #394, opened 2023-12-04
Handle opening multi-file granules

jrbourbeau:
Marking as WIP as this doesn't handle all the open + download code paths yet and I think `mypy` isn't happy right now either. Just wanted to get a quick +1 for this general direction before finishing up. 

Closes https://github.com/nsidc/earthaccess/issues/393

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -44,23 +44,15 @@ class EarthAccessFile(fsspec.spec.AbstractBufferedFile):
 
 
 def _open_files(
-    data_links: List[str],
-    granules: Union[List[str], List[DataGranule]],
+    url_mapping: dict[str, Union[DataGranule, None]],
     fs: fsspec.AbstractFileSystem,
     threads: Optional[int] = 8,
 ) -> List[fsspec.AbstractFileSystem]:
     def multi_thread_open(data: tuple) -> EarthAccessFile:
         urls, granule = data
-        if not isinstance(granule, str):
-            if len(granule.data_links()) > 1:
-                print(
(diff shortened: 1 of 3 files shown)


earthaccess/store.py
```diff
@@ -342,20 +344,14 @@ def _open_granules(
...
+                access = "on_prem"
                 s3_fs = None
 
-            data_links = list(
-                chain.from_iterable(
-                    granule.data_links(access=access_method) for granule in granules
-                )
-            )
-
+            url_mapping = _get_url_granule_mapping(granules, access)
```
jrbourbeau, line 350: The basic idea here is we now construct this mapping from filename to granule that contains the file. Currently we just have a list of files and list of granules, which isn't quite what we need

earthaccess/store.py
```diff
@@ -44,23 +44,15 @@ def __repr__(self) -> str:
...
-            if len(granule.data_links()) > 1:
-                print(
-                    "Warning: This collection contains more than one file per granule. "
-                    "earthaccess will only open the first data link, "
-                    "try filtering the links before opening them."
-                )
         return EarthAccessFile(fs.open(urls), granule)
 
-    fileset = pqdm(zip(data_links, granules), multi_thread_open, n_jobs=threads)
+    fileset = pqdm(url_mapping.items(), multi_thread_open, n_jobs=threads)
```
betolink, line 55: awesome, this is it!

betolink (approved):
I just tested #394 and worked as expected, thanks @jrbourbeau !!

PR #395, opened 2023-12-04
Add region to running in AWS check

jrbourbeau:
I noticed that right now we're just checking whether or not one is running in AWS in any region. This can lead to errors when attempting to open granules on AWS, but in a region outside of `us-west-2`. This PR updated our `_am_i_in_aws()` check to check that the region is `us-west-2` to avoid these types of errors.

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -141,11 +141,13 @@ class Store(object):
         try:
             # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
             resp = session.get(
-                "http://169.254.169.254/latest/meta-data/public-ipv4", timeout=1
+                "http://169.254.169.254/latest/meta-data/placement/region", timeout=1
             )
         except Exception:
             return False
-        if resp.status_code == 200:
+
+        if resp.status_code == 200 and b"us-west-2" == resp.content:
+            # On AWS in region us-west-2
             return True
         return False
 


earthaccess/store.py
```diff
@@ -112,7 +112,7 @@ def __init__(self, auth: Any, pre_authorize: bool = False) -> None:
         else:
             print("Warning: the current session is not authenticated with NASA")
             self.auth = None
-        self.running_in_aws = self._am_i_in_aws()
+        self.in_region = self._running_in_us_west_2()
```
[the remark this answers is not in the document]
betolink: this makes total sense and is semantically more accurate.

earthaccess/store.py
```diff
@@ -139,18 +139,25 @@ def _own_s3_credentials(self, links: List[Dict[str, Any]]) -> Union[str, None]:
...
         session = self.auth.get_session()
         try:
             # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
-            resp = session.put(
+            token_ = session.put(
                 "http://169.254.169.254/latest/api/token",
                 headers={"X-aws-ec2-metadata-token-ttl-seconds": "21600"},
                 timeout=1,
             )
+            resp = session.get(
```
betolink, line 151: I'll test both versions in 2i2c now...

betolink: it worked as expected!

betolink (approved):

earthaccess/store.py
jrbourbeau: Great, thanks for doing that @betolink!

Issue #392, opened 2023-12-02
Make dependency on python-dateutil explicit


mfisher87:
Our source code depends on dateutil. Conda Forge's autotick bot recognizes this and complains that the dependency isn't explicitly specified:

> ### Packages found by source code inspection but not in the meta.yaml:
>
> * python-dateutil

https://github.com/conda-forge/earthdata-feedstock/pull/35

This dependency is present implicitly through another dependency, but since our source depends on it directly, I feel it should be explicitly named in our specification (both pyproject.toml and feedstock meta.yaml)

PR #397, opened 2023-12-05
Add `python-dateutil` as a direct dependency

jrbourbeau:
Closes https://github.com/nsidc/earthaccess/issues/392

cc @mfisher87


MattF-NSIDC (approved):
Awesome, thank you!

When we do the release, we need to also add this dependency to `meta.yaml` of the `conda-forge/earthdata-feedstock`. I'll record an issue over there.

Issue #399, opened 2023-12-05
Remove binder? 


jrbourbeau:
Each PR gets a comment like this one https://github.com/nsidc/earthaccess/pull/397#issuecomment-1841113333 with a binder link. Though it looks like the binder is build is broken (see [this build](https://mybinder.org/v2/gh/jrbourbeau/earthaccess/f8cafefb8e1c9950fe2837e0e57e819cb9f2b32e) for an example). I'll propose we remove the PR comments until the builds are fixed.

MattF-NSIDC:
Good catch. I got so used to those comments that I stopped seeing them :rofl:

jhkennedy:
As of #821, Binder is working again, which begs the question do we want to re-enable the binder links? 

I personally don't care either way, but I lean *not* re-enabling them and closing this issue.

mfisher87:
I'd be pro re-enabling if they worked like the ReadTheDocs links work :)

PR #400, opened 2023-12-05
Remove binder PR comments

jrbourbeau:
xref https://github.com/nsidc/earthaccess/issues/399

cc @MattF-NSIDC


jrbourbeau:
So the binder comment still happened here. My guess is it may require the changes here to be on `main`. Happy to push up an empty PR to test after this PR has been merged

MattF-NSIDC:
>So the binder comment still happened here. My guess is it may require the changes here to be on main. 

I think RTD behaves the same way. @betolink do you think we should try to fix Binder before AGU for any reason?

Issue #386, opened 2023-12-01
AttributeError: 'DataGranules' object has no attribute 'headers' with `python-cmr==0.7.0`


aldotapia:
With version 0.8.0, I'm getting `AttributeError` after querying.
```
import earthaccess

earthaccess.login()

results = earthaccess.granule_query().short_name('MOD13Q1').bounding_box(-73.73,-55.01,-67.05,-17.63).version('061').temporal('2023-01-01','2023-01-31').get_all()
```

The error is `AttributeError: 'DataGranules' object has no attribute 'headers'`:

```
AttributeError                            Traceback (most recent call last)
Input In [7], in <cell line: 1>()
----> 1 results = earthaccess.granule_query().short_name('MOD13Q1').bounding_box(-73.73,-55.01,-67.05,-17.63).version('061').temporal('2023-01-01','2023-01-31').get_all()

File ~/miniforge3/envs/hidrocl/lib/python3.10/site-packages/cmr/queries.py:104, in Query.get_all(self)
     95 def get_all(self):
     96     """
     97     Returns all of the results for the query. This will call hits() first to determine how many
     98     results their are, and then calls get() with that number. This method could take quite
   (...)
    101     :returns: query results as a list
    102     """
--> 104     return self.get(self.hits())

File ~/miniforge3/envs/hidrocl/lib/python3.10/site-packages/earthaccess/search.py:331, in DataGranules.hits(self)
    322 """
    323 Returns the number of hits the current query will return. This is done by
    324 making a lightweight query to CMR and inspecting the returned headers.
    325 
    326 :returns: number of results reported by CMR
    327 """
    329 url = self._build_url()
--> 331 response = self.session.get(url, headers=self.headers, params={"page_size": 0})
    333 try:
    334     response.raise_for_status()
```

I tested past versions and I got the same problem with 0.7.0 and 0.7.1. I rolled back to 0.6.1 and it's working now.

Any hints?

MattF-NSIDC:
We tried to reproduce and had no luck. Is this still occurring for you?

If this was an intermittent issue with an external system, we should show a better error message.

jrbourbeau:
For reference, I tried with 0.8.0 and 0.8.1, but both times wasn't able to reproduce any error

```python
In [1]: import earthaccess

[...]

PR #402, opened 2023-12-05
Add CI tests with minimum supported versions

jrbourbeau:
xref https://github.com/nsidc/earthaccess/issues/386#issuecomment-1841302469

Closes https://github.com/nsidc/earthaccess/issues/386


pyproject.toml
```diff
@@ -1,3 +1,10 @@
+[project]
+name = "earthaccess"
+version = "0.8.1"
+
+[tool.setuptools]
+packages = ["earthaccess"]
+
```
jrbourbeau, line 7: The changes in this file were needed to make `pip install -e .` work. Open to suggestions from others on a better way to do this (this was me just working through errors until things worked)

jrbourbeau:
Note we're now getting 

```
FAILED tests/unit/test_results.py::test_data_links - AttributeError: 'DataGranules' object has no attribute 'headers'
```

Like reported in https://github.com/nsidc/earthaccess/issues/386

MattF-NSIDC (approved):
I think this looks awesome. I'd really like some documentation about how this file should be maintained before merging it in. A comment block at the top, or a mention in `CONTRIBUTING.md`, or a PR template with a checklist, or ...?

pyproject.toml
jhkennedy: I don't think we need setuptools *at all* and could/should just remove `setup.py`. Notably, it states:
```
# This is a shim to hopefully allow Github to detect the package, build is done with poetry
```
https://github.com/nsidc/earthaccess/blob/main/setup.py

It was added in #181 by @betolink  -- Luis was this just so GitHub can recognise [who depends on it](https://github.com/nsidc/earthaccess/network/dependents)?

jhkennedy:
> Thoughts about keeping the mindeps file up-to-date?

It we were using setuptools, we could do something like this:
* `pyproject.toml`
  ```toml
  [tool.setuptools.dynamic]
  dependencies = { file = ["ci/requirements-min.txt"] }
  ```
* test env using conda or pip:
  ```
  conda create --name earthaccess --file  ci/requirements-min.txt
  ```
  ```
  python -m pip install ci/requirements-min.txt
  ```
I'm not sure if you can do the same with Poetry

pyproject.toml
jhkennedy: Okay, if I checkout this branch, remove these changes to `pyproject.toml` (other than the deps update), and remove `setup.py`, this is the error message from pip:
```
ERROR: Project file://.../earthdata has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook. Since it does not have a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode. Consider using a build backend that supports PEP 660.
```

If I change the build system to:
```yaml
[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```
as recommended here: https://python-poetry.org/docs/pyproject#poetry-and-pep-517

it works

jhkennedy: Looks good!

jhkennedy:
> For us I think just manually updating is fine.

I agree with this as well. I like the comments in both files, and I don't think it's a burden since we should rarely be updating those.

jrbourbeau:
Thanks all for the reviews!

PR #403, opened 2023-12-05
Update poetry lockfile

jrbourbeau:
Over in https://github.com/nsidc/earthaccess/pull/402 I'm seeing some test failures from updating the poetry lock file that I'm confused about. I'm wondering if those failures are due to the changes in that PR or not (they don't seem related).

This PR runs `poetry lock` on the current `main` branch to see if CI is happy with that. That should tell us whether the failures in https://github.com/nsidc/earthaccess/pull/402 are related to the changes there or not.


jrbourbeau:
I'm going to merge this PR to unblock https://github.com/nsidc/earthaccess/pull/402 (should be a pretty uncontroversial change). I'm happy to handle any follow-up comments folks have. 

FWIW I'd be happy to just remove all the exception handling here and let 

```python
response.raise_for_status()
```

raise directly, but wanted to keep this PR as minimal as possible.

Issue #404, opened 2023-12-05
Release 0.8.2


jrbourbeau:
There have been a few bug fixes and maintenance updates since the last release that I'd like to get out before AGU. In particular 

- [x] https://github.com/nsidc/earthaccess/pull/391 
- [x] https://github.com/nsidc/earthaccess/pull/395
- [x] https://github.com/nsidc/earthaccess/pull/402

Anything else folks would like to see included?

betolink:
I think this will also be included, right?

- [ ] https://github.com/nsidc/earthaccess/pull/394

Good timing for an end to end example with HLS data!  I'm fine with these changes. I'll be working on building a new hub image for Openscapes tomorrow. Hopefully we can have the new version out, thanks a lot for all the work @jrbourbeau @MattF-NSIDC @itcarroll et al !

betolink:
Looks like we are ready! with the new tooling changes I'm hesitant to be the "release manager", I'm not sure if you have time to release today but if not I can try! @jrbourbeau

PR #405, opened 2023-12-06
Release 0.8.2

jrbourbeau:
Closes https://github.com/nsidc/earthaccess/issues/404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 # Changelog
 
-## Unreleased
-
+## [v0.8.2] 2023-12-06
 * Bug fixes:
-    * Follow Amazon EC2 IMDSv2 requirements to set `running-in-aws`
+    * Enable AWS check with IMDSv2
+    * Add region to running in AWS check
+    * Handle opening multi-file granules
+* Maintenance:
+    * Add CI tests with minimum supported versions
+    * Update poetry lockfile
+    * Add `python-dateutil` as a direct dependency
+    * Remove binder PR comments
+    * Add YAML formatting (prettier)
 
 ## [v0.8.1] 2023-12-01
 * New Features:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,16 @@
 # Changelog
 
-## Unreleased
-
+## [v0.8.2] 2023-12-06
 * Bug fixes:
-    * Follow Amazon EC2 IMDSv2 requirements to set `running-in-aws`
+    * Enable AWS check with IMDSv2
+    * Add region to running in AWS check
```
MattF-NSIDC, line 6: There was a breaking API change here, right? We changed `.am_i_in_aws()` to `.in_region()` I think, and I think we should call that out.

CHANGELOG.md
```diff
@@ -1,9 +1,16 @@
...
+    * Handle opening multi-file granules
```
MattF-NSIDC, line 7: I didn't get a chance to look at this, but does it add any new surface to the API that we should call out by name, e.g. new method?

MattF-NSIDC (changes requested):
Would like a little more detail in the changelog before you merge if you don't mind :)

CHANGELOG.md
jrbourbeau: `_am_i_in_aws()` is a private method and my understanding is the `.running_on_aws` attribute is intended to be private (from @betolink in the offline `earthaccess` Slack channel). This is intended to just be an internal change, not a public-facing one.

MattF-NSIDC: :heavy_check_mark: thanks, just wanted to be completely sure. Without the leading underscore, I assumed :)

jrbourbeau: No worries -- same here

PR #414, opened 2023-12-18
Add workflow to update PRs with docs preview link in description

mfisher87:
This improves accessibility by removing the need to know in advance where to find the ReadTheDocs link in the GitHub PR interface. 

See demo: https://github.com/mfisher87/earthaccess/pull/11 (the link doesn't work because RTD isn't set up for my fork ;) )

### DIFF
--- a/.github/workflows/pr-rtd-link.yml
+++ b/.github/workflows/pr-rtd-link.yml
@@ -1,0 +1,19 @@
+# The ReadTheDocs preview link is "hidden" within the GitHub "Checks"
+# interface. For users who don't know this, finding the preview link may be
+# very difficult or frustrating. This workflow makes the link more
+# findable by updating PR descriptions to include it.
+name: "Add ReadTheDocs preview link to PR descriptions"
+
+on:
+  pull_request_target:
+
+permissions:
+  pull-requests: "write"
+
+jobs:
+  autolink-rtd-previews:
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: "readthedocs/actions/preview@v1"
+        with:
+          project-slug: "earthaccess"
--- a/.github/workflows/pr-rtd-link.yml
+++ b/.github/workflows/pr-rtd-link.yml
@@ -6,6 +6,8 @@ name: "Add ReadTheDocs preview link to PR descriptions"
 
 on:
   pull_request_target:
+    types:
+       - opened
 
 permissions:
   pull-requests: "write"
--- a/.github/workflows/pr-rtd-link.yml
+++ b/.github/workflows/pr-rtd-link.yml
@@ -7,7 +7,7 @@ name: "Add ReadTheDocs preview link to PR descriptions"
 on:
   pull_request_target:
     types:
-       - opened
+      - opened
 
 permissions:
   pull-requests: "write"


.github/workflows/pr-rtd-link.yml
```diff
@@ -0,0 +1,19 @@
...
+# interface. For users who don't know this, finding the preview link may be
+# very difficult or frustrating. This workflow makes the link more
+# findable by updating PR descriptions to include it.
+name: "Add ReadTheDocs preview link to PR descriptions"
+
+on:
+  pull_request_target:
+
+permissions:
+  pull-requests: "write"
```
jrbourbeau, line 11: I could be totally misunderstanding this, but does this give PRs from anyone write permissions to `nsidc/earthaccess`?

jrbourbeau (commented):
Thanks @mfisher87!

I'm curious if this a problem that has happened in a wild a few times? My gut reaction is to just point folks to the existing RTD PR build link instead of leaving a comment on every PR. Though, to be clear, if this is a common problem that would make life easier for lots of folks, then adding automation is fine (just wanting to avoid adding extra infrastructure if what we already have is reasonably sufficient)

mfisher87:
>instead of leaving a comment on every PR

To clarify, this edits the description of the PR, doesn't add a comment :)

jrbourbeau:
> To clarify, this edits the description of the PR, doesn't add a comment :)

Ah, great. I think that's better than a comment. I'm still curious about how widespread of an issue we think this is.

.github/workflows/pr-rtd-link.yml
mfisher87: This gives the action permission to do things specifically to pull requests, for example add labels, comments, or in this case, edit the description of the PR. [Full list](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests)

Also, because the action uses `pull_request_target` event, it is restricted from seeing the code in the PR branch; it sees only the code in the target branch which we own. This means that the action that runs will be the one on the target branch and contributors can't, for example, change our GitHub Actions workflow YAML to do something malicious to all our pull requests.

jrbourbeau: Okay, thanks for the extra context and clarification 

> This means that the action that runs will be the one on the target branch and contributors can't, for example, change our GitHub Actions workflow YAML to do something malicious to all our pull requests.

Great -- that's what I was wondering about

mfisher87:
>I'm still curious about how widespread of an issue we think this is.

Speaking only for myself, I see it as a widespread issue that many of us, including myself, are blind to until we encounter someone who's been contributing for a while but has no idea PR previews have been there all along. I would describe the group impacted as "all users with no (or trivial) prior experience with either GitHub or ReadTheDocs", which is a huge group!

This user persona has no way to know:

1) There is a PR preview at all
2) That the PR preview is under the "show all checks" interface item:

image

3) That "details" link will take you to the PR preview:

image

All of this language actually misdirects a user seeking a preview _away_ from clicking on it! :sweat:

I think many potential future contributors are members of science communities who are undergoing transitions towards working in open source, and have little experience with the tools that feel like turning on a faucet to me (us) after using them for so long. If we can lower the barriers to bringing them on to our ways of working, it will be easier to see the path to adoption, and the slow march towards more open and reproducible science takes another step ;)

.github/workflows/pr-rtd-link.yml
```diff
@@ -0,0 +1,19 @@
+# The ReadTheDocs preview link is "hidden" within the GitHub "Checks"
+# interface. For users who don't know this, finding the preview link may be
+# very difficult or frustrating. This workflow makes the link more
+# findable by updating PR descriptions to include it.
+name: "Add ReadTheDocs preview link to PR descriptions"
+
+on:
+  pull_request_target:
```
jhkennedy, line 8: This action recommends specifying 
```suggestion
on:
  pull_request_target:
    types:
       - opened
```
and I believe the link is stable so wouldn't need to trigger on any other event types. Is there a reason you don't include the type?

jhkennedy: Also, I lean towards only doing this on PRs with docs changes -- is there a good reason to always build the preview?

jhkennedy (commented):
I'd be in favor of this! It definitely simplifies getting to the preview, and while ReadTheDocs is frequently used, there are a lot of projects that instead use CircleCI asset previews or just publish to `github.io`, so I do not expect all/most users just to know how to get the docs preview.

mfisher87:
pre-commit.ci autofix

.github/workflows/pr-rtd-link.yml
mfisher87: Both good calls!

We will always build the preview no matter what with our current RTD configuration, but there may be a way to tweak that in readthedocs.yml. I just don't know of it :) This new workflow only surfaces the link, so yeah maybe it makes sense to add a `paths` filter too! Even on PRs without docs changes, you can access the RTD preview from the "checks" area of the PR interface.

jhkennedy: @mfisher87, if RTD generates it no matter what, let's always add the link. It sure can't hurt to have it easy to look at.

jrbourbeau:
> I'd be in favor of this!

Cool -- glad to hear there are multiple people that would find this useful. I mostly just wanted to raise the point that there's a maintenance cost associated with adding automation like this, and maintainer time is often hard to come by, so it's good to be thoughtful before taking that additional maintenance cost on. It sounds like you both have been thoughtful (not surprising), so feel free to go for it 👍

mfisher87:
Thanks James! I agree it's very important to be careful with maintenance burden, and I really appreciate having a team that helps me check my biases :) When something sounds like a good idea, I think we're all vulnerable, at least a little bit, to underestimating the costs.

PR #424, opened 2024-01-08
add us-west-2 check to API and improve auth repr

JessicaS11:
Per #231, we've added API access to `store._running_in_us_west_2()` in the form of a printed statement that the user is (or is not) running in AWS region us-west-2. In the process of implementing this, it became clear that `get_edl_token()` had not been added to `__init__.py`, so that was added as well. It also adds a repr for the `auth` object via a `auth.auth_info()` property.

I'm seeking input on:
- if this PR fully satisfies #231 - was the spirit to check specifically for us-west-2, or to enable the user to see what region they are running in? Pending that answer, it might make sense to have a more generic "earthaccess.region()` function instead.
- What should be in `auth.__repr__`? Currently it's not much better than the object ID:
```
Authentication Info
-------------------
authenticated?: True
tokens: [{'access_token': 'eyJ0eXAiOiJKV1QiLCJvcmlnaW4iOiJFYXJ0aGRhdGEgTG9naW4iLCJzaWciOiJlZGxqd3RwdWJrZXlfb3BzIiwiY[truncated]...', 'token_type': 'Bearer', 'expiration_date': '02/09/2024'}]```

### DIFF
--- a/earthaccess/__init__.py
+++ b/earthaccess/__init__.py
@@ -10,11 +10,13 @@ from .api import (
     get_requests_https_session,
     get_s3_credentials,
     get_s3fs_session,
+    get_edl_token,
     granule_query,
     login,
     open,
     search_data,
     search_datasets,
+    in_us_west_2,
 )
 from .auth import Auth
 from .kerchunk import consolidate_metadata
@@ -24,6 +26,7 @@ from .store import Store
 logger = logging.getLogger(__name__)
 
 __all__ = [
+    # api.py
     "login",
     "search_datasets",
     "search_data",
@@ -31,15 +34,21 @@ __all__ = [
     "get_fsspec_https_session",
     "get_s3fs_session",
     "get_s3_credentials",
+    "get_edl_token"
     "granule_query",
     "collection_query",
     "open",
     "download",
+    "auth_environ",
+    "in_us_west_2",
+    # search.py
     "DataGranules",
     "DataCollections",
+    # auth.py
     "Auth",
+    # store.py
     "Store",
-    "auth_environ",
+    # kerchunk
     "consolidate_metadata",
 ]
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by JessicaS11** [comment]
Thanks to @battistowx during today's earthaccess cowork session for moving this forward with me!

@andypbarrett @betolink Any thoughts on what info we want to be in the auth repr? I'm happy to move that to a separate PR if we want to move forward with the s3check portion of this one.

File: earthaccess/__init__.py [hunk 1]
```diff
@@ -24,22 +24,28 @@
...
+    # api.py
     "login",
     "search_datasets",
     "search_data",
     "get_requests_https_session",
     "get_fsspec_https_session",
     "get_s3fs_session",
     "get_s3_credentials",
-    "granule_query",
+    "get_edl_token" "granule_query",
```
mfisher87, line 35: Warning: these strings will concatenate :)

earthaccess/auth.py
```diff
@@ -76,6 +76,28 @@ def __init__(self) -> None:
         self.EDL_GENERATE_TOKENS_URL = "https://urs.earthdata.nasa.gov/api/users/token"
         self.EDL_REVOKE_TOKEN = "https://urs.earthdata.nasa.gov/api/users/revoke_token"
 
+    def __repr__(self) -> str:
+        print_str = "Authentication Info\n" + "-------------------\n"
+        for k, v in self.auth_info.items():
+            print_str += str("{}: {}\n".format(k, v))
+
+        return print_str
```
chuckwondo, line 84: I _strongly_ discourage this implementation of `__repr__` for at least 2 reasons:

1. (minor) this is arguably an "abuse" of `__repr__`, which should generally be: "If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment)." (see https://docs.python.org/3/reference/datamodel.html#object.__repr__)
2. (**very important**) this can lead to inadvertent leaking of the user's token(s) during logging since `auth_info` includes those tokens

JessicaS11:
Can someone comment if any of this is still relevant given #1011? I've cleaned up the conflicts to make it easier to see what's going on.

earthaccess/auth.py
JessicaS11: What about something like (and get rid of the auth_info function) instead:
```
    def __str__(self) -> str:
        print_str = str("Authenticated?: {}\n".format(self.authenticated))
        print_str += str("View your token with `.tokens`")

        return print_str
```

Issue #421, opened 2023-12-29
Option to use UAT environment


danielfromearth:
### Challenge

`earthaccess` does not currently support access to collections that are in NASA Earthdata's User Acceptance Testing (UAT) environment. Instead, the production (PROD) environment is assumed when logging in via `earthaccess` and there is no option to change the environment.

### Desired functionality
When executing the `earthaccess.login()` method, an option to specify which environment should be available.  And the authentication should be correctly applied to the specified environment.

### Example
I encountered this challenge when trying to test data transformation code with a data collection for the Tropospheric Emissions Monitoring of POllution (TEMPO) instrument that is in UAT.  I would like to search and download granules from this collection using `earthaccess`, but that is currently not possible.

### Benefit
If the UAT environment can be accessed with `earthaccess`, its usability would be expanded to more dataset testing.

mfisher87:
Should we support manually overriding the EDL base URL to anything? That way we can support other existing environments like System Integration Testing (SIT), and any others that may be created in the future.

danielfromearth:
Supporting SIT would be great too!

danielfromearth:
@mfisher87 , @betolink,

Currently, one of the options for `.login()`'s `strategy=` argument is "environment".  
Because of that, it seems to me like it might be confusing to add a new keyword argument to `.login()` named `environment=`, which would result in function calls looking something like this:
```python
earthdata.login(strategy="environment", environment=Env.UAT)
```

Using "environment" twice where it means two different things looks strange, right?

Would "endpoint" perhaps be a better name for the new keyword argument?  Other suggestions?

jhkennedy:
"maturity" is generally how I think about UAT/SIT/PROD, but "cmr_endpoint" is a good option as well (just endpoint makes me go "to what?")

danielfromearth:
nice, I like "cmr_endpoint" too

danielfromearth:
I've started working on this _issue_ but could use some more expert eyes to ensure an appropriate token is available when using UAT.   Specifically, my question is: 

Will the `client_id` parameter on [this line](https://github.com/nsidc/earthaccess/blob/f47668501b21163049d02dd186522e680e464bb8/earthaccess/auth.py#L61) — for the `self.EDL_GET_PROFILE` variable — need to change when the base of the URL changes from "urs.earthdata.nasa.gov" to "uat.urs.earthdata.nasa.gov"

mfisher87:
[...]

PR #426, opened 2024-01-11
Issue 421 - option to use Earthdata User Acceptance Testing (UAT) system

danielfromearth:
GitHub Issue: #421 

### Description

This change:
- enables the login and use of different Earthdata Login (EDL) and Common Metadata Repository (CMR) systems — that is, either Production (`PROD`) or User Acceptance Testing (`UAT`).
- removes the `Auth.get_user_profile()` method and use of a `client_id`.

### Local test steps

Passed new UAT unit test

### Overview of integration done

Ran the following (with an ID for a TEMPO mission collection in UAT substituted for `<UAT Collection ID>`):

```python
import earthaccess
auth = earthaccess.login(system=earthaccess.UAT)
results = earthaccess.search_data(concept_id=<UAT Collection ID>)
files = earthaccess.download(results, "./local_folder")
```

...and this generated an appropriate response, and successfully download the granule files.

## Checklist
* [x] Unit tests added/updated and passing.
* [x] Integration testing
* [x] `CHANGELOG.md` updated
* [x] Documentation updated (if needed).

----
📚 Documentation preview 📚: https://earthaccess--426.org.readthedocs.build/en/426/


mfisher87:
Daniel and I discussed integration testing strategy here. Since UAT tests could be more flaky than PROD tests, we decided we will write only unit tests for this PR (plus one sample integration test) and address the problem of flaky tests in another effort. I'll link the new issue to this one.

danielfromearth:
And with regard to unit tests, we discussed creating a new module which holds all UAT unit tests in the test suite, instead of duplicating+modifying the existing tests within the existing modules.

mfisher87:
@danielfromearth @chuckwondo sorry for the delay pushing this up. I just forgot :blush:

tests/unit/test_uat.py
```diff
@@ -0,0 +1,70 @@
...
+            status=200,
+        )
+        responses.add(
+            responses.GET,
+            f"https://uat.urs.earthdata.nasa.gov/api/users/user?client_id={CLIENT_ID}",
+            json={},
+            status=200,
+        )
+
+        # TODO: Add a mock for CMR query?  Does it need a "CMR-HITS" field in the response?
```
danielfromearth, line 41: When the following 6 lines are uncommented, then the `results = search_data(session=session)` line fails because `CMR-HITS` is expected to be in the response:

```
>       results = search_data(session=session)

test_uat.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../earthaccess/api.py:127: in search_data
    granules_found = query.hits()
../../earthaccess/search.py:435: in hits
    return int(response.headers["CMR-Hits"])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = {'Content-Type': 'application/json'}, key = 'CMR-Hits'

    def __getitem__(self, key):
>       return self._store[key.lower()][1]
E       KeyError: 'cmr-hits'
```

danielfromearth:
This is getting close, but not quite there yet, because there is still another place where the UAT environment is not being set.

`earthaccess.login(earthdata_environment=Env.UAT)` runs [this line](https://github.com/nsidc/earthaccess/blob/d09e3582fbba80de497dfbc3618a9bb51648a945/earthaccess/api.py#L150) (`earthaccess.__auth__.login(...)`).  Because that line gets the `__auth__` attribute from earthaccess, it calls the `def __getattr__(name)` function in `__init__.py` , which runs `_auth.login(strategy=strategy)` without the UAT environment specified.

Will need to find an elegant way to get the correct environment over to the `_auth` variable in `__init__.py`, without resorting to something like `earthaccess.__auth__(env='UAT').login(env='UAT')`

danielfromearth:
I believe this is ready to go, as it appears to work!  

I think it's not the _most_ beautiful result — for instance, the API could be simpler than having to do 
```
from earthaccess.auth import Env`  # in order to execute 
earthaccess.login(earthdata_environment=Env.UAT)
```
or writing it out as `earthaccess.login(earthdata_environment=earthaccess.auth.Env.UAT)`.  
In addition, the change [here](https://github.com/nsidc/earthaccess/pull/426/commits/5121511d803b3639b916226eff9ac8a9a4780ec6) is a bit hacky, but it seems to do the job.

betolink (dismissed):
Great PR @danielfromearth!

chuckwondo:
Please do not merge this PR yet. I have some suggestions, so I'd like a chance to provide comments.

chuckwondo (changes requested):
I'm marking this as "request changes", but have yet to provide comments, which are forthcoming.

betolink:
ACK Not merging yet. @chuckwondo

danielfromearth:
Sounds good @chuckwondo, standing by!

chuckwondo:
Hi @danielfromearth. Sorry for the delay in getting comments to you for this PR. I just landed the PR for #528, and next up is ironing out the kinks on #520 since Ian has been waiting longer. Then I'll circle back. In the meantime, would you mind pulling the latest changes from the main branch into your branch for this PR?

earthaccess/auth.py
```diff
@@ -21,6 +24,15 @@
...
 logger = logging.getLogger(__name__)
+CLIENT_ID = "ntD0YGC_SM3Bjs-Tnxd7bg"
+
+# Host URL options, for different Earthdata domains.
+Environment_Definitions = namedtuple("Environment_Definitions", ["PROD", "UAT", "SIT"])
+Env = Environment_Definitions(
+    {"edl": "urs.earthdata.nasa.gov", "cmr": CMR_OPS},
+    {"edl": "uat.urs.earthdata.nasa.gov", "cmr": CMR_UAT},
+    {"edl": "sit.urs.earthdata.nasa.gov", "cmr": CMR_SIT},
+)
```
chuckwondo, line 35: IIRC, part of the discussion we had a little while back was to find a good name to represent "environment" that avoids confusion with the "environment" login strategy.

I think the code above may still has potential for confusion, and has also led to a rather long parameter name elsewhere (`earthdata_environment`). To address this, and also to introduce some better typing constructs, here's what I suggest in place of the above, which uses the term "system" rather than "environment":

```python
from dataclasses import dataclass

from typing_extensions import NewType

import cmr

CMRBaseURL = NewType("CMRBaseURL", str)
EDLHostname = NewType("EDLHostname", str)


@dataclass(frozen=True)
class System:
    cmr_base_url: CMRBaseURL
    edl_hostname: EDLHostname


PROD = System(CMRBaseURL(cmr.CMR_OPS), EDLHostname("urs.earthdata.nasa.gov"))
UAT = System(CMRBaseURL(cmr.CMR_UAT), EDLHostname("uat.urs.earthdata.nasa.gov"))
SIT = System(CMRBaseURL(cmr.CMR_SIT), EDLHostname("sit.urs.earthdata.nasa.gov"))
```

I also suggest that this code be placed in its own `system.py` file, as I don't feel that putting this in `auth.py` is quite the right spot. Yes, the EDL hostname is related to auth, but making use of a "system" does not _require_ authorization.

For example, when calling the `api.search_data` function, we might want to specify a `mode` argument, like so:

```python
search_data(mode=UAT.cmr_base_url, ...)
```

Ignoring the questionable use of the name `mode` (from `python_cmr`), I can specify a different "mode" without requiring auth, so to have to import "system" from the `auth` module is perhaps not particularly intuitive.

Perhaps @betolink, @mfisher87, and @jhkennedy can provide some feedback on this suggestion before proceeding.

earthaccess/auth.py
```diff
@@ -72,12 +86,14 @@ def __init__(self) -> None:
...
-        self.EDL_GENERATE_TOKENS_URL = "https://urs.earthdata.nasa.gov/api/users/token"
-        self.EDL_REVOKE_TOKEN = "https://urs.earthdata.nasa.gov/api/users/revoke_token"
+        self._set_earthdata_environment(Env.PROD)
 
-    def login(self, strategy: str = "netrc", persist: bool = False) -> Any:
+    def login(
+        self,
+        strategy: str = "netrc",
+        persist: bool = False,
+        earthdata_environment: Optional[dict] = None,
```
chuckwondo, line 95: If we end up using my "system" suggestion, this would change as follows:

```suggestion
        system: Optional[System] = None,
```

There would be other corresponding changes to use "system", but I won't comment on all of the locations to avoid a plethora of comments.

earthaccess/auth.py
```diff
@@ -261,9 +301,9 @@ def _netrc(self) -> bool:
             raise FileNotFoundError(f"No .netrc found in {Path.home()}") from err
         except NetrcParseError as err:
             raise NetrcParseError("Unable to parse .netrc") from err
-        if my_netrc["urs.earthdata.nasa.gov"] is not None:
-            username = my_netrc["urs.earthdata.nasa.gov"]["login"]
-            password = my_netrc["urs.earthdata.nasa.gov"]["password"]
+        if my_netrc[self.earthdata_environment["edl"]] is not None:
+            username = my_netrc[self.earthdata_environment["edl"]]["login"]
+            password = my_netrc[self.earthdata_environment["edl"]]["password"]
```
chuckwondo, line 306: This might change to the following, if "system" is used:

```suggestion
        if (creds := my_netrc[self.system.edl_hostname]) is not None:
            username = creds["login"]
            password = creds["password"]
```

earthaccess/store.py
```diff
@@ -188,7 +188,7 @@ def set_requests_session(
                 resp.raise_for_status()
             else:
                 self._requests_cookies.update(new_session.cookies.get_dict())
-        elif resp.status_code >= 200 and resp.status_code <= 300:
+        elif 200 <= resp.status_code <= 300:
```
chuckwondo, line 191: I think we want this instead:

```suggestion
        elif 200 <= resp.status_code < 300:
```

We also seem to be missing handling the rest of the 3xx and 4xx range.

earthaccess/auth.py
```diff
@@ -21,6 +24,15 @@
 
 
 logger = logging.getLogger(__name__)
+CLIENT_ID = "ntD0YGC_SM3Bjs-Tnxd7bg"
```
chuckwondo, line 27: Is this the same for all environments (PROD, UAT, and SIT)? If so, then no worries, but if not, then we may need to add a `client_id` attribute to the environment/system instances.

chuckwondo (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Enhancements:
   * Corrected and enhanced static type hints for functions and methods that make
     CMR queries or handle CMR query results (#508)
+  * Enable queries to Earthdata test environments for authenticated accounts (pull #426)
```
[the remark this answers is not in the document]
danielfromearth: Changed!
I didn't change other lines in the changelog, e.g., the other two lines in the "unreleased" section, because I'm not totally sure which issues they correspond to.

earthaccess/store.py
danielfromearth: Good observation @chuckwondo! 
For the 3xx and 4xx handling, I would appreciate some suggestions, since I'm not sure myself :)

earthaccess/auth.py
danielfromearth: The code you propose and suggest putting in `system.py` looks fine to me!

Maybe we want a simpler API?  

Such as case-insensitive strings, with acceptable options of 'UAT', 'SIT', and 'PROD' — for example:
```python
login(env="UAT")
```
or 
```python
login(system="UAT")
```

chuckwondo: Using bare strings just gets us back to where we started, basically. We need a group of values, not a single string. That is, the UAT system consists of a CMR URL and an EDL hostname, so passing `env="UAT"` doesn't buy me anything. What do you propose be done with that string value?

earthaccess/store.py
chuckwondo: my suggestion would be to remove the condition on the final `else` (which currently checks for code >= 500), so that everything else raises

earthaccess/auth.py
danielfromearth: Sorry, I didn't explain enough.  I was thinking of a string only for the API. Then, earthaccess could do validation of the string, convert it to one of the `System` objects like you propose; so, internally, earthaccess would use and pass around a `System` object, not a string.

chuckwondo: We wouldn't want that either, as it would not only create extra code and a "split" (i.e., using a string in one place, and an object elsewhere), but also would not allow someone to define there own System and pass it in.

chuckwondo: @betolink, @mfisher87, @jhkennedy, I'd like to get some feedback on this change suggestion, so we can get some consensus on this PR and get it landed.

Overall, this PR looks pretty sound, so my suggestion here is primarily to add a bit more sound typing with better auto-completion, while avoiding name conflict/confusion/overloading of the term "environment."

chuckwondo: For a bit more clarity, my suggestion would mean that instead of something like the following, as @danielfromearth gave in an earlier comment:

```python
import earthaccess
auth = earthaccess.login(earthdata_environment=earthaccess.auth.Env.UAT)
```

we would do something like this:

```python
import earthaccess
auth = earthaccess.login(system=earthaccess.PROD)
```

or perhaps this, depending upon how we expose things:


```python
import earthaccess
from earthaccess.system import PROD
auth = earthaccess.login(system=PROD)
```

earthaccess/auth.py
danielfromearth: I think this is the same for now, though I'm not sure this is ideal for the future. @betolink, can you shed more light on this question?

CHANGELOG.md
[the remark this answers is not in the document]
danielfromearth: @chuckwondo, would you like me (or would you like) to create a new _issue_: for updating the changelog to have links throughout?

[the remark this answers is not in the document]
chuckwondo: Good idea. I'll create a new issue for it.

earthaccess/auth.py
mfisher87: Hey all, really time constrained right now so apologies in advance for likely missing things!

> system vs environment

I agree avoiding "environment" is useful because we've overloaded that term. No strong feeling on "system", I also have nothing better! I've heard the term "region" used to describe UAT/SIT/PROD, but that's also overloaded. Dang it!

> `earthaccess.PROD`

I like having the "system" choices at the top level import. Users will already have that. Nice idea!

> dataclass instances instead of enum

I love this. I apologize for the inferior suggestion @danielfromearth :) 

I also like the use of `NewType`!

danielfromearth: no worries @mfisher87!  
And I like having the "system" types exposed at the top-level (like `earthaccess.PROD`) too 👍

earthaccess/auth.py
betolink: This is a borrowed hard-coded cliend-id from another NSIDC app. I think we should create an earthaccess app in NASA Earthdata. It was only needed because when we interact with the token endpoints or any user endpoint the EDL API requires to know which app is requesting that info (I think this is basically enforcing that any client has to be registered).

earthaccess/auth.py
betolink: Nice! this would be cleaner!

earthaccess/auth.py
betolink: I like 

```python
import earthaccess
auth = earthaccess.login(system=earthaccess.PROD)
```
Just because is simpler.

betolink (commented):
I think the PR is almost ready to be merged just pending the `system=earthaccess.PROD` convention. I'm going to open a new PR on the client-id with EDL. I guess it's time to have our own.

danielfromearth:
@chuckwondo 
> Further, I think we may want to drop SIT, and offer only PROD and UAT. 
That's fine with me. I am only away of SIT use-cases currently.

> @danielfromearth, just to confirm my suspicion about needing a separate client ID per system/env, did you attempt to test a scenario against UAT where a client ID would be required? I would think that such a test would fail if you were to supply the PROD client ID we currently have hard-coded.

On this branch, I successfully ran `earthaccess`'s `login()`, `search_data()`, and `download()` for a collection in UAT.  It found the granules and downloaded the files. So, perhaps this client ID is associated with both "systems"?

mfisher87:
I'm not sure about this. It doesn't seem connected to anything you changed? I don't have much experience with either poetry or asdf

chuckwondo:
> @betolink, @mfisher87, @chuckwondo... any thoughts on resolving the new [CI failure for readthedocs](https://readthedocs.org/projects/earthaccess/builds/24297951/), which states there is a missing poetry version?

The problem is that the docs build system is asking `asdf` to install the latest version of `poetry`, and `asdf` thinks that 1.8.3 is the latest version, which is only half correct. The latest _tag_ of `poetry` is 1.8.3, but the latest _release_ is 1.8.2, so it seems that either `asdf` is going about things the wrong way in terms of obtaining a valid list of versions, or `poetry` had a problem releasing 1.8.3, so we're kind of stuck at the moment.

chuckwondo:
> > @betolink, @mfisher87, @chuckwondo... any thoughts on resolving the new [CI failure for readthedocs](https://readthedocs.org/projects/earthaccess/builds/24297951/), which states there is a missing poetry version?
> 
> The problem is that the docs build system is asking `asdf` to install the latest version of `poetry`, and `asdf` thinks that 1.8.3 is the latest version, which is only half correct. The latest _tag_ of `poetry` is 1.8.3, but the latest _release_ is 1.8.2, so it seems that either `asdf` is going about things the wrong way in terms of obtaining a valid list of versions, or `poetry` had a problem releasing 1.8.3, so we're kind of stuck at the moment.

One way to fix this is to change `latest` to `1.8.2` in the following lines in our `readthedocs.yml` file:

```plain
      - asdf install poetry latest
      - asdf global poetry latest
```

mfisher87:
This worked fine on this commit which I'm about to merge: https://github.com/nsidc/earthaccess/pull/498/commits/52e70c083f20c7c8ab53d9a6208cdff889bd5eb3

danielfromearth:
> > One way to fix this is to change `latest` to `1.8.2` in the following lines in our `readthedocs.yml` file:
> 
> Good call, this worked fine on this commit which I'm about to merge: [52e70c0](https://github.com/nsidc/earthaccess/commit/52e70c083f20c7c8ab53d9a6208cdff889bd5eb3)

this versioning is a configuration that won't be checked/updated by any automated CI workflow, right?  Is that okay with you all?

chuckwondo:
> > > One way to fix this is to change `latest` to `1.8.2` in the following lines in our `readthedocs.yml` file:
> > 
> > 
> > Good call, this worked fine on this commit which I'm about to merge: [52e70c0](https://github.com/nsidc/earthaccess/commit/52e70c083f20c7c8ab53d9a6208cdff889bd5eb3)
> 
> this versioning is a configuration that won't be checked/updated by any automated CI workflow, right? Is that okay with you all?

Not sure what you mean by that. This fix is precisely what should fix the currently breaking builds. Please merge main into your branch and your PR build should finally pass all checks.

chuckwondo:
> @chuckwondo
> 
> > Further, I think we may want to drop SIT, and offer only PROD and UAT.
> 
> That's fine with me. I am only away of SIT use-cases currently.
> 
> > @danielfromearth, just to confirm my suspicion about needing a separate client ID per system/env, did you attempt to test a scenario against UAT where a client ID would be required? I would think that such a test would fail if you were to supply the PROD client ID we currently have hard-coded.
> 
> On this branch, I successfully ran `earthaccess`'s `login()`, `search_data()`, and `download()` for a collection in UAT. It found the granules and downloaded the files. So, perhaps this client ID is associated with both "systems"?

I'm not sure that's a valid test of the client ID, as I believe your unit tests simply mock responses, but please correct me if I'm wrong (I haven't looked at all of your relevant unit tests to confirm this). If that's the case, then you may need to make use of `vcrpy` to make (and record) _real_ requests/responses to see if tests requiring a client ID fail when making a real request to UAT with what I suspect is a PROD-only client ID.

chuckwondo:
> > @chuckwondo
> > > Further, I think we may want to drop SIT, and offer only PROD and UAT.
> > 
> > 
> > That's fine with me. I am only away of SIT use-cases currently.
> > > @danielfromearth, just to confirm my suspicion about needing a separate client ID per system/env, did you attempt to test a scenario against UAT where a client ID would be required? I would think that such a test would fail if you were to supply the PROD client ID we currently have hard-coded.
> > 
> > 
> > On this branch, I successfully ran `earthaccess`'s `login()`, `search_data()`, and `download()` for a collection in UAT. It found the granules and downloaded the files. So, perhaps this client ID is associated with both "systems"?
> 
> I'm not sure that's a valid test of the client ID, as I believe your unit tests simply mock responses, but please correct me if I'm wrong (I haven't looked at all of your relevant unit tests to confirm this). If that's the case, then you may need to make use of `vcrpy` to make (and record) _real_ requests/responses to see if tests requiring a client ID fail when making a real request to UAT with what I suspect is a PROD-only client ID.

Upon further inspection, I'm inclined to believe that we can completely eliminate the client ID. What makes me believe this is that, on the main branch, I ran all integration tests, and I got 9 failures, which I believe are to be expected when running locally (I recall this from a previous topic of discussion during a "hack day"). I then removed _all_ uses of client ID throughout the entire codebase, reran the integration tests, and got identical results.

There are many tests that download files, and all succeeded without the client ID (other than the ones that also failed _with_ the client ID).

I believe the only case for _requiring_ a client ID is for a web app that requires redirection after login. For EDL to know where to redirect the user, you must supply a client ID and the registered app with that client ID must be configured with a redirect URL. This is not the case for `earthaccess`, so as far as I can tell, we should be able to completely drop the use of a client ID.

However, since I'm not _completely_ confident in my line of thinking here (even though test results seem to point in this direction), I have written email to support@earthdata.nasa.gov to get a definitive answer for us.

danielfromearth:
> > @chuckwondo
> > > Further, I think we may want to drop SIT, and offer only PROD and UAT.
> > 
> > 
> > That's fine with me. I am only away of SIT use-cases currently.
> > > @danielfromearth, just to confirm my suspicion about needing a separate client ID per system/env, did you attempt to test a scenario against UAT where a client ID would be required? I would think that such a test would fail if you were to supply the PROD client ID we currently have hard-coded.
> > 
> > 
> > On this branch, I successfully ran `earthaccess`'s `login()`, `search_data()`, and `download()` for a collection in UAT. It found the granules and downloaded the files. So, perhaps this client ID is associated with both "systems"?
> 
> I'm not sure that's a valid test of the client ID, as I believe your unit tests simply mock responses, but please correct me if I'm wrong (I haven't looked at all of your relevant unit tests to confirm this). If that's the case, then you may need to make use of `vcrpy` to make (and record) _real_ requests/responses to see if tests requiring a client ID fail when making a real request to UAT with what I suspect is a PROD-only client ID.

Right, the unit tests mock responses. What I ran for this test was not a unit test; instead, I manually ran earthaccess in a Python kernel to login and download files — for my case, TEMPO data granules — from UAT.  And then manually confirmed that the downloaded files were the expected files.

danielfromearth:
@mfisher87, @chuckwondo, would you be able to help address the new readthedocs failures that have cropped up after merging `main`?

chuckwondo:
> @mfisher87, @chuckwondo, would you be able to help address the new readthedocs failures that have cropped up after merging `main`?

@danielfromearth, take a look at the details of the failure: https://readthedocs.org/projects/earthaccess/builds/24322214/

You'll see this in the error output:

```plain
WARNING -  griffe: earthaccess/auth.py:251: No type or annotation for parameter 'system'
WARNING -  griffe: earthaccess/auth.py:251: Parameter 'system' does not appear in the function signature
```

Given that, here's what I see in the code in your branch that would cause that warning: https://github.com/nsidc/earthaccess/blob/aeaa2800b47b6399943537f722ccbe28032fb472/earthaccess/auth.py#L246-L251

Can you see the issue? In your docstring, you document a `system` parameter to the `get_session` method, but there is no such parameter for that method. Should there be, or should you remove the entry from the docstring?

danielfromearth:
Thanks @chuckwondo for identifying the change needed!  I was overthinking it, and didn't realize the other messages were simply "INFO" messages and not "warnings" 🤦

danielfromearth:
Are we ready to merge?!

danielfromearth:
Oh, or do we want to remove SIT?

chuckwondo:
> Do we want to remove SIT?

I suggest we do.

> Upon further inspection, I'm inclined to believe that we can completely eliminate the client ID. What makes me believe this is that, on the main branch, I ran all integration tests, and I got 9 failures, which I believe are to be expected when running locally (I recall this from a previous topic of discussion during a "hack day"). I then removed _all_ uses of client ID throughout the entire codebase, reran the integration tests, and got identical results.
> 
> There are many tests that download files, and all succeeded without the client ID (other than the ones that also failed _with_ the client ID).
> 
> I believe the only case for _requiring_ a client ID is for a web app that requires redirection after login. For EDL to know where to redirect the user, you must supply a client ID and the registered app with that client ID must be configured with a redirect URL. This is not the case for `earthaccess`, so as far as I can tell, we should be able to completely drop the use of a client ID.
> 
> However, since I'm not _completely_ confident in my line of thinking here (even though test results seem to point in this direction), I have written email to [support@earthdata.nasa.gov](mailto:support@earthdata.nasa.gov) to get a definitive answer for us.

Further, I think we should wait for me to get a reply from support@earthdata.nasa.gov regarding the client ID. Ideally, my hunch is correct, and we can simply remove all uses of the client ID. I did get an automated reply, so hopefully I'll get an answer this week. Sorry to hold this up further, but I'm not comfortable with leaving the questions surrounding the client ID unanswered.

mfisher87:
@chuckwondo could you also start a discussion on Slack about the client ID? I think Luis likely knows more but has limited time to catch up on this PR. I am fairly sure you're right that we don't need it. I vaguely remember discussions of using it for metrics purposes, but I think I'm confusing that with the user agent string.

earthaccess/search.py
```diff
@@ -452,12 +453,13 @@ def __init__(self, auth: Optional[Auth] = None, *args: Any, **kwargs: Any) -> No
         """Base class for Granule and Collection CMR queries."""
```
chuckwondo, line 453:
```suggestion
```

earthaccess/search.py
```diff
@@ -95,12 +95,13 @@ def __init__(self, auth: Optional[Auth] = None, *args: Any, **kwargs: Any) -> No
...
             # To search, we need the new bearer tokens from NASA Earthdata
-            auth.get_session(bearer_token=True)
-            if auth is not None and auth.authenticated
-            else requests.session()
-        )
+            self.session = auth.get_session(bearer_token=True)
+
+            self.mode(auth.system.cmr_base_url)
+        else:
+            self.session = requests.sessions.Session()
```
chuckwondo, line 104:
```python
        self.session = (
            # To search, we need the new bearer tokens from NASA Earthdata
            auth.get_session(bearer_token=True)
            if auth and auth.authenticated
            else requests.session()
        )

        if auth:
            self.mode(auth.system.cmr_base_url)
```

earthaccess/search.py
```diff
@@ -452,12 +453,13 @@ def __init__(self, auth: Optional[Auth] = None, *args: Any, **kwargs: Any) -> No
...
```
chuckwondo, line 462:
```python
        self.session = (
            # To search, we need the new bearer tokens from NASA Earthdata
            auth.get_session(bearer_token=True)
            if auth and auth.authenticated
            else requests.session()
        )

        if auth:
            self.mode(auth.system.cmr_base_url)
```

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@
 
 * Enhancements:
   * Corrected and enhanced static type hints for functions and methods that make
-    CMR queries or handle CMR query results (#508)
+    CMR queries or handle CMR query results ([#508](https://github.com/nsidc/earthaccess/issues/508))
+  * Enable queries to Earthdata test environments for authenticated accounts ([#421](https://github.com/nsidc/earthaccess/issues/421))
```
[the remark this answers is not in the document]
chuckwondo: Please also add:

```plain
* Changes
  * Removed the `get_user_profile` method and the `email_address` and `profile` attributes from the `Auth` class
```

chuckwondo (changes requested):
Fantastic! I have only a handful of very minor grammatical suggestions (mostly around consistent "system" wording), and a couple of very minor code tweaks.

earthaccess/auth.py
```diff
@@ -261,9 +277,9 @@ def _netrc(self) -> bool:
...
         except NetrcParseError as err:
             raise NetrcParseError("Unable to parse .netrc") from err
-        if my_netrc["urs.earthdata.nasa.gov"] is not None:
-            username = my_netrc["urs.earthdata.nasa.gov"]["login"]
-            password = my_netrc["urs.earthdata.nasa.gov"]["password"]
+        if (creds := my_netrc[self.system.edl_hostname]) is not None:
+            username = creds["login"]
+            password = creds["password"]
         else:
             return False
```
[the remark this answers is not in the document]
mfisher87: :100:

chuckwondo (approved):
Well done @danielfromearth!

danielfromearth:
Thank you @chuckwondo for your close look at all of the changes in this PR!

Issue #427, opened 2024-01-16
Broken image in sea level rise tutorial


mfisher87:
The first image on this tutorial is broken:

https://earthaccess.readthedocs.io/en/latest/tutorials/SSL/

jessnicwelch:
@mfisher87 I can't seem to do it... could you add the label `good first issue` to this one?

mfisher87:
Label added and sent an invite to this repo with triage access!

Issue #436, opened 2024-01-25
Implement EDL user agent to track usage metrics 


asteiker:
This was suggested by @betolink to ensure some way for us to track how many access requests are made through earthaccess. Something like:

`user-agent: f"earthaccess v{earthaccess.__version__}"`

@mikedorfman confirmed that those headers are identifiable in the API Gateway logs and: 

> can be either queried using CloudWatch Logs Insights or imported into CloudMetrics (I'm certain about the former, pretty sure about the latter).  TEA doesn't do anything with that header, so adding this shouldn't be detrimental.

PR #443, opened 2024-02-01
fixing searching for restricted datasets and accessing ASF on demand data from Opera

### LINKED ISSUE #438: Restricted datasets can't be used with the API notation
If we want to search for a dataset that's not publicly available 

```python
import earthaccess

auth = earthaccess.login()
results = earthaccess.search_datasets(
  short_name="RESTRICTED_DATA"
)
```
will always return zero, even though our user has the proper access permission. I think the reason is because we are not passing the authenticated session to `hits()` here:

https://github.com/nsidc/earthaccess/blob/dd61f2330866e5678ef9e1f0264a02f6922cc9fc/earthaccess/search.py#L54

A fix would be to implement the method using the auth session.

### LINKED ISSUE #439: Permissions error when downloading files from OPERA_L2_RTC-S1_V1 dataset on ASF DAAC (but not OPERA_L3_DIST-ALERT-HLS_PROVISIONAL_V0 on LP DAAC)
I am attempting to download a few granules from the OPERA_L2_RTC-S1_V1 dataset using the following:

```python
import earthaccess
earthaccess.login()

    short_name='OPERA_L2_RTC-S1_V1',
    cloud_hosted=True,
    bounding_box=(-117.33, 35.541, -117.880, 35.991),
    temporal=("2023-10-01", "2023-10-15"),
    count=50
)

files = earthaccess.download(results, "./data")
```

Even though the initial authentication step seems to be working as it should, I am encountering the following HTTP 401 error (for every granule):

```
Error while downloading the file OPERA_L2_RTC-S1_T136-290177 IW2_20231004T002041Z_20240123T055542Z_S1A_30_v1.0.h5

requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: ....
```

Interestingly, the same syntax applied to a different OPERA dataset doesn't produce the same error, and the files are downloaded as expected.

```python
import earthaccess
earthaccess.login()

    short_name='OPERA_L3_DIST-ALERT-HLS_PROVISIONAL_V0',
    cloud_hosted=True,
    bounding_box=(-117.33, 35.541, -117.880, 35.991),
    temporal=("2023-10-01", "2023-10-15"),
    count=50
)

files = earthaccess.download(results, "./data")
```
A key difference between these datasets is that the OPERA RTC product is distributed by ASF DAAC whereas the OPERA DIST-ALERT product is distributed by LP DAAC. I am wondering if the issue may stem from a difference in the way that earthaccess requests these files from the two DAACs. I am relatively new to this means of data access, so any guidance would be much appreciated!

betolink:
* Fixes #439 by implementing multiple trusted domains (ASF's) on top of NASA's URS
  * This will fix downloading data behind ASF datapool but we still can't open them because aiohttp/fsspec do not support rebuilding the authorization headers on redirects. This should work now
  * ```python
    import earthaccess
    earthaccess.login()

        short_name='OPERA_L2_RTC-S1_V1',
        cloud_hosted=True,
        bounding_box=(-117.33, 35.541, -117.880, 35.991),
        temporal=("2023-10-01", "2023-10-15"),
        count=50
    )
    files = earthaccess.download(results, "./data")
    ```
* Fixes #438 by using an authenticated session in `hits()` instead of calling parent's class `super()`
* Improves search by adding instrument and project to collection queries #427 
  * ```python
    collections = earthaccess.search_datasets(
        instrument="GEDI"
        cloud_hosted=True
    )
    ```

----
📚 Documentation preview 📚: https://earthaccess--443.org.readthedocs.build/en/443/

### DIFF
--- a/binder/environment-dev.yml
+++ b/binder/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # This environment bootstraps poetry, the actual dev environment
   # is installed and managed with poetry
-  - python=3.9
+  - python=3.10
   - jupyterlab=3
   - xarray>=0.19
   - ipyleaflet>=0.13
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -13,7 +13,7 @@ from tinynetrc import Netrc
 from .daac import DAACS
 
 try:
-    user_agent = f"earthaccess v{importlib.metadata.version('earthacess')}"
+    user_agent = f"earthaccess v{earthaccess.__version__}"
 except importlib.metadata.PackageNotFoundError:
     user_agent = "earthaccess"
 
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -18,10 +18,13 @@ class SessionWithHeaderRedirection(requests.Session):
     """
     Requests removes auth headers if the redirect happens outside the
     original req domain.
-    This is taken from https://wiki.earthdata.nasa.gov/display/EL/How+To+Access+Data+With+Python
     """
 
-    AUTH_HOST = "urs.earthdata.nasa.gov"
+    AUTH_HOSTS: List[str] = [
+        "urs.earthdata.nasa.gov",
+        "cumulus.asf.alaska.edu",
+        "datapool.asf.alaska.edu",
+    ]
 
     def __init__(
         self, username: Optional[str] = None, password: Optional[str] = None
@@ -39,11 +42,13 @@ class SessionWithHeaderRedirection(requests.Session):
         if "Authorization" in headers:
             original_parsed = urlparse(response.request.url)
             redirect_parsed = urlparse(url)
-            if (
-                (original_parsed.hostname != redirect_parsed.hostname)
-                and redirect_parsed.hostname != self.AUTH_HOST
-                and original_parsed.hostname != self.AUTH_HOST
+            if (original_parsed.hostname != redirect_parsed.hostname) and (
+                redirect_parsed.hostname not in self.AUTH_HOSTS
+                or original_parsed.hostname not in self.AUTH_HOSTS
             ):
+                logger.debug(
+                    f"Deleting Auth Headers: {original_parsed.hostname} -> {redirect_parsed.hostname}"
+                )
                 del headers["Authorization"]
         return
 
@@ -210,7 +215,7 @@ class Auth(object):
         Returns:
             class Session instance with Auth and bearer token headers
         """
-        session = requests.Session()
+        session = SessionWithHeaderRedirection()
         if bearer_token and self.authenticated:
             # This will avoid the use of the netrc after we are logged in
             session.trust_env = False
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -6,10 +6,19 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 
+
 import requests  # type: ignore
 from tinynetrc import Netrc
 
 from .daac import DAACS
+
+import importlib.metadata
+
+try:
+    user_agent = f"earthaccess v{importlib.metadata.version('earthacess')}"
+except importlib.metadata.PackageNotFoundError:
+    user_agent = "earthaccess"
+    
 
 logger = logging.getLogger(__name__)
 
@@ -31,6 +40,9 @@ class SessionWithHeaderRedirection(requests.Session):
         self, username: Optional[str] = None, password: Optional[str] = None
     ) -> None:
         super().__init__()
+        self.headers.update({"User-Agent": user_agent})
+
+
         if username and password:
             self.auth = (username, password)
 
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -31,6 +31,7 @@ class SessionWithHeaderRedirection(requests.Session):
         "urs.earthdata.nasa.gov",
         "cumulus.asf.alaska.edu",
         "sentinel1.asf.alaska.edu",
+        "nisar.asf.alaska.edu",
         "datapool.asf.alaska.edu",
     ]
 
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -1,4 +1,5 @@
 import getpass
+import importlib.metadata
 import logging
 import os
 from netrc import NetrcParseError
@@ -6,13 +7,10 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 
-
 import requests  # type: ignore
 from tinynetrc import Netrc
 
 from .daac import DAACS
-
-import importlib.metadata
 
 try:
     user_agent = f"earthaccess v{importlib.metadata.version('earthacess')}"
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -18,7 +18,7 @@ try:
     user_agent = f"earthaccess v{importlib.metadata.version('earthacess')}"
 except importlib.metadata.PackageNotFoundError:
     user_agent = "earthaccess"
-    
+
 
 logger = logging.getLogger(__name__)
 
@@ -41,7 +41,6 @@ class SessionWithHeaderRedirection(requests.Session):
     ) -> None:
         super().__init__()
         self.headers.update({"User-Agent": user_agent})
-
 
         if username and password:
             self.auth = (username, password)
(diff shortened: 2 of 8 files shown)


earthaccess/results.py
```diff
@@ -228,7 +228,9 @@ def __repr__(self) -> str:
         Temporal coverage: {self['umm']['TemporalExtent']}
         Size(MB): {self.size()}
         Data: {data_links}\n\n
-        """.strip().replace("  ", "")
+        """.strip().replace(
```
betolink, line 231: Black likes this, Ruff does not, which one is right? @mfisher87

mfisher87: I think we should pick one :) Ruff offers a formatter that re-implements Black, but I've never worked with it before. Personally, I'd turn Ruff formatting off and stick with Black due to my lack of experience, but if I had more time I'd try to learn more and switch to Ruff for everything.

earthaccess/auth.py
```diff
@@ -18,10 +18,13 @@ class SessionWithHeaderRedirection(requests.Session):
     """
     Requests removes auth headers if the redirect happens outside the
     original req domain.
-    This is taken from https://wiki.earthdata.nasa.gov/display/EL/How+To+Access+Data+With+Python
     """
 
-    AUTH_HOST = "urs.earthdata.nasa.gov"
+    AUTH_HOSTS: List[str] = [
+        "urs.earthdata.nasa.gov",
+        "cumulus.asf.alaska.edu",
```
jhkennedy, line 25:
```suggestion
        "cumulus.asf.alaska.edu",
        "sentinel1.asf.alaska.edu",
```

Sentinel-1 is behined it's own distribution app:
```
curl -L -H "Authorization: Bearer $EDL_TOKEN" https://sentinel1.asf.alaska.edu/GRD_HD/SA/S1A_IW_GRDH_1SDV_20240201T203334_20240201T203408_052370_06552E_14CD.zip -O
```
The ARIA interferrograms (netCDF) are also behind `grfn.asf.alaska.edu` but that distribution app doesn't accept bearer tokens.

earthaccess/search.py
```diff
@@ -106,6 +115,39 @@ def doi(self, doi: str) -> Type[CollectionQuery]:
...
+        self.params["instrument"] = instrument
+        return self
+
+    def project(self, project: str) -> Type[CollectionQuery]:
+        """Searh datasets by associated project
+
+        ???+ Tip
+            Not all datasets have an associated project. This works
+            only at the dataset level but not the granule (data) level.
+            Will bring datasets across DAACs matching the project.
```
jhkennedy, line 140:
```suggestion
            Will return datasets across DAACs matching the project.
```
Bring reads weird here to me

earthaccess/auth.py
mfisher87: Do we know anyone with Earthdata Login who can provide a security perspective on this? I feel a bit weird about sending SSO credentials anywhere other than the SSO host!

jhkennedy: I mean, sending credentials to `datapool.asf.alaska.edu`, which is the domain for the URLs of record in CMR is technically the incorrect part as it doesn't do auth at all (just redirects). `cumulus` and `sentinel1` are the actual auth endpoints

betolink: I think that's correct, maybe we should fix the OAuth2 behavior for future releases. However this custom domains will still be a problem.

binder/environment-dev.yml
```diff
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # This environment bootstraps poetry, the actual dev environment
   # is installed and managed with poetry
-  - python=3.9
+  - python=3.10
```
betolink, line 7: This one is only for the local dev environment and actually not necessary as we can just pip install the project locally.... I use it to bootstrap the environment and install some conda libs to test notebooks with (cartopy)

betolink:
Should we include #436 on this PR? since we are touching the EDL session?

jhkennedy:
> Should we include #436 on this PR? since we are touching the EDL session? 

Yes, that's a good idea

earthaccess/auth.py
mfisher87: >cumulus and sentinel1 are the actual auth endpoints.

What I mean is that, based on my understanding, they should be communicating with the SSO host (EDL) to determine if the user is authenticated/authorized instead of directly receiving credentials from the user. The SSO host acts as an intermediary to reduce credential exposure. Are there other apps on EDL that are taking user credentials? What is EDL's official stance on this practice?

jhkennedy: @mfisher87 your username and password should only ever be sent to EDL directly, but the bearer token is *intended* to be sent to EDL *apps* (cumulus, tea, etc), and yes, you should be conscious of which domains you're sending your token too. 

Stated here:
image

mfisher87: Thanks, makes sense! That 2-token limit is so weird.

Maybe we should open an issue to expose a way for end-users to extend the list of approved distribution apps? I assume our list isn't exhaustive or would go out of date.

jhkennedy: > That 2-token limit is so weird.

It's just like AWS Access keys -- it's really a 1 active key limit, plus space to generate a new one for rotation (you make a new one and then delete the old).

But yeah, it's annoying.

earthaccess/auth.py
```diff
@@ -11,22 +12,34 @@
...
+        "sentinel1.asf.alaska.edu",
+        "datapool.asf.alaska.edu",
+    ]
 
     def __init__(
         self, username: Optional[str] = None, password: Optional[str] = None
     ) -> None:
         super().__init__()
+        self.headers.update({"User-Agent": user_agent})
```
betolink, line 41: `earthaccess` will now use the `User-Agent` header to identify requests to EDL. 🎉

earthaccess/auth.py
```diff
@@ -11,22 +12,34 @@
...
     Requests removes auth headers if the redirect happens outside the
     original req domain.
-    This is taken from https://wiki.earthdata.nasa.gov/display/EL/How+To+Access+Data+With+Python
     """
 
-    AUTH_HOST = "urs.earthdata.nasa.gov"
+    AUTH_HOSTS: List[str] = [
+        "urs.earthdata.nasa.gov",
+        "cumulus.asf.alaska.edu",
+        "sentinel1.asf.alaska.edu",
```
jhkennedy, line 33:
```suggestion
        "nisar.asf.alaska.edu",
```
We may as well be forward-thinking here and add the NISAR distribution app:
https://nisar.asf.alaska.edu

Which does support bearer tokens and [S3 Access](https://nisar.asf.alaska.edu/s3credentialsREADME)

betolink:
Should we merge this before the other open PRs? @jhkennedy @mfisher87

mfisher87:
pre-commit.ci autofix

earthaccess/search.py
```diff
@@ -106,6 +115,39 @@ def doi(self, doi: str) -> Type[CollectionQuery]:
         self.params["doi"] = doi
         return self
 
+    def instrument(self, instrument: str) -> Type[CollectionQuery]:
+        """Searh datasets by instrument
+
+        ???+ Tip
```
mfisher87, line 121: TIL: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#collapsible-blocks

mfisher87:
>Should we merge this before the other open PRs? @jhkennedy @mfisher87

Logistically, that makes sense to me. I haven't done a proper review here and won't have time in short-term, so I'll leave the "approval" decision to others :)

earthaccess/auth.py
```diff
@@ -11,22 +12,35 @@
 
 from .daac import DAACS
 
+try:
+    user_agent = f"earthaccess v{importlib.metadata.version('earthacess')}"
```
jhkennedy, line 16:
```suggestion
    user_agent = f"earthaccess v{earthaccess.__version__}"
```
Since we already do this work when earthaccess [is imported](https://github.com/nsidc/earthaccess/blob/main/earthaccess/__init__.py#L46), it'd probably make more sense to use the version reported by the package. 

If we need more error handling, I'd be inclined to move it to the `__init__.py`

Note: imports will also need to be adjusted.

jhkennedy (approved):
I think this looks good! Should work as is, but I added a take-or-leave suggestion

betolink:
> I think this looks good! Should work as is, but I added a take-or-leave suggestion

I think that last commit broke things... 😂 @jhkennedy

jhkennedy:
> > I think this looks good! Should work as is, but I added a take-or-leave suggestion
> 
> I think that last commit broke things... 😂 @jhkennedy

Yes, hense the "Note: imports will also need to be adjusted." :wink:

betolink:
I'm using importlib so we avoid importing the library and potentially produce one of those circular import errors, I think it should be good now.

betolink:
going to merge this @jhkennedy

jhkennedy (approved):

Issue #279, opened 2023-08-15
Expected type of search_data() "bounding_box" seems too narrow


arthur-e:
The following generates a `TypeError` (Traceback at bottom of post):

```py
results = earthaccess.search_data(
    short_name = 'M2SDNXSLV',
    temporal = ("2023-01", "2023-07"), 
    bounding_box = [2, 35, 4, 37]) # NOTE: A list
```

While this example works as expected (no error):

```py
results = earthaccess.search_data(
    short_name = 'M2SDNXSLV',
    temporal = ("2023-01", "2023-07"), 
    bounding_box = (2, 35, 4, 37)) # NOTE: A tuple
```

It's not immediately obvious from [the documentation](https://nsidc.github.io/earthaccess/user-reference/api/api/#earthaccess.api.search_data) that `bounding_box` *must* be a tuple.

Maybe there is a good reason that a `list` type is not acceptable? It seems to me like `list` and `tuple` should both be accepted.

Traceback:

```py
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[68], line 1
----> 1 results = earthaccess.search_data(
      2     short_name = 'M2SDNXSLV',
      3     temporal = ("2023-01", "2023-07"), 
      4     bounding_box = [2, 35, 4, 37])
      6 time_series = []
      7 file_list = earthaccess.open(results)

File /usr/local/python-env/OpenClimate/lib/python3.8/site-packages/earthaccess/api.py:103, in search_data(count, **kwargs)
     65 def search_data(
     66     count: int = -1, **kwargs: Any
     67 ) -> List[earthaccess.results.DataGranule]:
     68     """Search dataset granules using NASA's CMR.
     69 
     70     [https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html)
   (...)
    101         ```
    102     """
--> 103     query = DataGranules().parameters(**kwargs)
    104     granules_found = query.hits()
    105     print(f"Granules found: {granules_found}")

File /usr/local/python-env/OpenClimate/lib/python3.8/site-packages/earthaccess/search.py:347, in DataGranules.parameters(self, **kwargs)
    345         methods[key](*val)
    346     else:
--> 347         methods[key](val)
    349 return self

TypeError: bounding_box() missing 3 required positional arguments: 'lower_left_lat', 'upper_right_lon', and 'upper_right_lat'
```

[...]

PR #448, opened 2024-02-06
Clarify that temporal and bbox args must be tuples in docstrings, fix docstring formatting

mfisher87:
Addresses #279, but does not resolve it. This work exposed and addresses a couple of concerns, but there are some still I think that need more work.

## Docstring formatting

On the `main` branch,  there are some formatting errors that are causing key information to be omitted from the docs. It seems _really_ easy to make these mistakes, and Mkdocs doesn't complain about them! How can we protect ourselves?

* We could really use some linting rules that can catch docstrings that are not formatted properly in the Google style. Ruff has some docstring PEP linters, but I don't think that's enough here. Can Ruff lint for Google docstring style?
* We need to be very careful with line-continuation. Line continuation must be done with an indent or everything on the next line will be either omitted or misformatted. A better option, IMO, is to use block style like many of the changes in this PR.
* We need to be very careful with whitespace. A blank line under `Parameters:` will cause the entire parameters block to be formatted like a code block instead of parsed.
* We have redundant type information in docstrings, e.g. `strategy (String):` duplicates the type (`str`) in the function signature. Mkdocs will pick up the type from the function signature so we don't need to keep these values in sync over time.

----
📚 Documentation preview 📚: https://earthaccess--448.org.readthedocs.build/en/448/

### DIFF
--- a/binder/environment-dev.yml
+++ b/binder/environment-dev.yml
@@ -10,6 +10,16 @@ dependencies:
   - ipyleaflet>=0.13
   - h5netcdf>=0.11
   - cartopy
+
+  - mkdocs>=1.2
+  - mkdocs-material>=7.1,<9.0
+  - markdown-include>=0.6
+  - mkdocstrings>=0.19.0
+  - mkdocstrings-python
+  - mkdocs-jupyter>=0.19.0
+  - pymdown-extensions>=9.2
+
   - pip
   - pip:
       - poetry
+      - markdown-callouts>=0.2.0
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -201,7 +201,7 @@ def open(
 
     Parameters:
         granules: a list of granule instances **or** list of URLs, e.g.
-            `s3://some-granule`, if a list of URLs is passed we need to specify the data
+            `s3://some-granule`, if a list of URLs is passed, we need to specify the data
             provider e.g. POCLOUD, NSIDC_CPRD etc.
 
     Returns:
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -31,38 +31,30 @@ def _normalize_location(location: Union[str, None]) -> Union[str, None]:
 def search_datasets(
     count: int = -1, **kwargs: Any
 ) -> List[earthaccess.results.DataCollection]:
-    """Search datasets using NASA's CMR
+    """Search datasets using NASA's CMR.
 
     [https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html)
 
     Parameters:
-
         count (Integer): Number of records to get, -1 = all
-        kwargs (Dict): arguments to CMR:
+        kwargs (Dict):
+            arguments to CMR:
 
             * **keyword**: case insensitive and support wild cards ? and *,
-
             * **short_name**: e.g. ATL08
-
             * **doi**: DOI for a dataset
-
             * **daac**: e.g. NSIDC or PODAAC
-
             * **provider**: particular to each DAAC, e.g. POCLOUD, LPDAAC etc.
-
             * **temporal**: a tuple representing temporal bounds in the form
-              ("yyyy-mm-dd", "yyyy-mm-dd")
-
+              `("yyyy-mm-dd", "yyyy-mm-dd")`
             * **bounding_box**: a tuple representing spatial bounds in the form
-              (lower_left_lon, lower_left_lat , upper_right_lon, upper_right_lat)
+              `(lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat)`
 
     Returns:
-
-        a list of DataCollection results that can be used to get
-        information such as concept_id, doi, etc. about a dataset.
+        a list of DataCollection results that can be used to get information about a
+            dataset, e.g. concept_id, doi, etc.
 
     Examples:
-
         ```python
         datasets = earthaccess.search_datasets(
             keyword="sea surface anomaly",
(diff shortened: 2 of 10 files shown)


binder/environment-dev.yml
```diff
@@ -10,6 +10,16 @@ dependencies:
...
+  - markdown-include>=0.6
+  - mkdocstrings>=0.19.0
+  - mkdocstrings-python
+  - mkdocs-jupyter>=0.19.0
+  - pymdown-extensions>=9.2
+
   - pip
   - pip:
       - poetry
+      - markdown-callouts>=0.2.0
```
mfisher87, line 25: This enabled me to build docs without messing with Poetry. Wasn't sure if this is intentional, as CONTRIBUTING.md doesn't really specify.

earthaccess/api.py
```diff
@@ -246,10 +241,8 @@ def get_s3_credentials(
 def collection_query() -> Type[CollectionQuery]:
     """Returns a query builder instance for NASA collections (datasets)
 
-    Parameters:
-        cloud_hosted (Boolean): initializes the query builder for cloud hosted collections.
```
mfisher87, line 250: Mkdocs was throwing warnings about these parameters which don't exist in the signature. That's awesome, but can we get it to upgrade those to errors so we can protect ourselves from info getting out of sync? There are still some unaddressed warnings after this PR.

earthaccess/results.py
```diff
@@ -111,14 +110,15 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
     def concept_id(self) -> str:
         """
         Returns:
-          Retrurns a collection's `concept_id`, this id is the most relevant search field on granule queries.
+            A collection's `concept_id`.
+            This id is the most relevant search field on granule queries.
```
mfisher87, line 114: Hey @danielfromearth, thanks for the additional cleanup :) I have some more details in the PR description, but indentation on multi-line strings is super important for the docs to render correctly! I can't comment a suggestion on my own PR :roll_eyes: , but should look like:

```
Returns:
    A collection's `concept_id`.
        This id is the most relevant search field on granule queries
```

Without the additional indentation, it will render two rows of return documentation.

earthaccess/results.py
```diff
@@ -149,7 +149,7 @@ def abstract(self) -> str:
     def landing_page(self) -> str:
         """
         Returns:
-            if available it returns the first landing page for the collection, can be many.
+            The first landing page for the collection (can be many), if available.
```
mfisher87, line 152: :100: Love making these a bit more succinct!

earthaccess/results.py
```diff
@@ -159,15 +159,16 @@ def landing_page(self) -> str:
...
         """
         links = self._filter_related_links("GET DATA")
         return links
 
     def s3_bucket(self) -> Dict[str, Any]:
         """
         Returns:
-            Returns the S3 bucket information if the collection has it (**cloud hosted collections only**)
+            The S3 bucket information if the collection has it.
+            (**cloud hosted collections only**)
```
mfisher87, line 171: Needs indentation!

earthaccess/results.py
```diff
@@ -290,17 +289,20 @@ def data_links(
         """Returns the data links form a granule
 
         Parameters:
-            access: direct or external, direct means in-region access for cloud hosted collections.
-            in_region: if we are running in us-west-2, meant for the store class, default is False
+            access: direct or external.
+              direct means in-region access for cloud-hosted collections.
+            in_region: True if we are running in us-west-2.
+              It is meant for the store class.
```
mfisher87, line 295: I believe we need two more spaces on these continuation lines.

earthaccess/results.py
```diff
@@ -310,14 +312,14 @@ def data_links(
...
-                # Even though we are in us-west-2 the user wants the HTTPS links
+                # Even though we are in us-west-2, the user wants the HTTPS links
                 # used in region they are S3 signed links from TEA
                 # https://github.com/asfadmin/thin-egress-app
                 return https_links
         else:
             # we are not in region
             if access == "direct":
-                # maybe the user wants to collect S3 links ans use them later
+                # maybe the user wants to collect S3 links and use them later
```
mfisher87, line 322: We should add `codespell` to our pre-commit config!

andypbarrett (approved):
In my opinion, this clarifies the inputs.

earthaccess/results.py
danielfromearth: That totally works for me. 
For clarity though, let me ask further... is the additional indentation a ReadTheDocs thing?  
The second example in the [Google style guide (section 3.8.3)](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) shows the `Args:` section with no additional indentation. Here is the relevant docstring example copied for convenience:

```
Args:
      table_handle:
        An open smalltable.Table instance.
      keys:
        A sequence of strings representing the key of each table row to
        fetch.  String keys will be UTF-8 encoded.
      require_all_keys:
        If True only rows with values set for all keys will be returned.
```

(Note that `earthaccess` currently uses `Parameters:` throughout its codebase instead of `Args:`.)

earthaccess/results.py
danielfromearth: I was following the example from the [Google style guide (section 3.8.3)]. But if this causes problems with the documentation generation, we should certainly add the additional space.

mfisher87: Ah. Maybe two spaces does work. Haven't actually tested it. I still prefer to go to 4 for consistency with the rest of the file, or to update the rest of the file to have 2-space indents continuation lines. Not a strong preference though, if it's working!

earthaccess/results.py
mfisher87: I'm not 100% sure. I struggled with the docstring style a lot. I was certain I had made a fix exactly like this in several other places in this file to fix rendering, but those fixes might have been subtly different than what I'm commenting on now. I've replaced a lot of that context with other stuff, I just remember checking over the output exhaustively and feeling confident about it before setting it down. I trust you're doing similarly exhaustive checking on your end, so if it's working I'm happy :)

earthaccess/results.py
danielfromearth: Confirmed that indentation is needed for it to display correctly in https://earthaccess--448.org.readthedocs.build/en/448/

earthaccess/results.py
danielfromearth: I'm adding the indentation, since it seems like it is necessary for the readthedocs to display correctly.

earthaccess/results.py
danielfromearth: Let's just stick with four spaces, so we don't have to remember "when 2?",  "when 4?", etc.

danielfromearth (approved):
I believe this is ready to go now.

mfisher87:
Wow, thanks Daniel for going all out on these docstrings!!! We needed this :)

earthaccess/results.py
mfisher87: Weird. It feels like we could/should have better validation for docstrings, but haven't taken the time to explore that yet.

Another thing I'm not clear on is why we're using `Parameters:` sections instead of `Args:`. Since it's working, I just decided to focus on other things :)

danielfromearth:
@mfisher87 , @jhkennedy , we could enable linting of docstrings via ruff [this way](https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings). What do you think?

I just ran it and got many more things to potentially fix...

snapshot

mfisher87:
Heck yes, thank you for looking in to that!!!

danielfromearth:
> Heck yes, thank you for looking in to that!!! Should we break that in to another PR so we can merge this one?

sounds good to me 👍

earthaccess/api.py
jhkennedy: `mkdocs build --strict` will exit with a return code of 1 and a message with the number of warnings raised.
https://github.com/mkdocs/mkdocs/pull/1613#issuecomment-416703384

jhkennedy (approved):
This looks good to me!

earthaccess/api.py
mfisher87: Nice! I'll open an issue referencing this comment :)

Issue #319, opened 2023-10-13
`pathlib` all the things!


MattF-NSIDC:
@betolink @jrbourbeau how do you feel about replacing all `os.path` things with `pathlib`? I'm sure there's a Ruff or other pre-commit rule we can use to ensure the whole codebase is using our preferred interface for doing pathy things. Maybe ruff or some other codemod out there can even automatically migrate us?

I don't think this is urgent of course, but it seems like a worthwhile code quality improvement to track as an issue if we all agree :)

jhkennedy:
I'll add my support to this! I've been very happy moving over to `pathlib` primarily. There is one caveat worth keeping in mind with this package though -- `pathlib` will will normalize URL paths (`https://a/b/c` --> `https:/a/b/c`), breaking them. So, we'll just want to be careful there.

MattF-NSIDC:
Ah, that is a weird little footgun. I've never encountered that myself, but now that you say it, of course... `https://a/b/c` is a valid posix path: colons can be in directory names and and duplicate path separators are ignored.

MattF-NSIDC:
I think your concern is that we may be using `os.path.join` to construct URL paths, right? I haven't looked to confirm if this is true. If there are cases like this, should we use plain f-strings (or urllib) instead of `os.path` features? Using `os.path` to make URL paths is a bit of a red herring to me as a reader.

jhkennedy:
Yeah, it's fun:
![image](https://github.com/nsidc/earthaccess/assets/7882693/19fc7ddc-0601-4796-843e-f3244ddf1875)

>  think your concern is that we may be using os.path.join to construct URL paths, right? I haven't looked to confirm if this is true. If there are cases like this, should we use plain f-strings (or urllib) instead of os.path features? Using os.path to make URL paths is a bit of a red herring to me as a reader.

It's more, the space earthaccess operates in, you might not know if you've got a local file path or an HTTP path to a file (the whole idea, is generally, users *shouldn't* care). It's just something to keep in mind when making the changes or reviewing them.

I agree, if we're constructing URLs we should be using f-strings or something like `urllib`. And for file paths, I prefer `pathlib` over `os.path`.

jhkennedy:
Doing a quick search:
https://github.com/search?q=repo%3Ansidc%2Fearthaccess%20os.path&type=code

I don't see anywhere that `os.path` is used that isn't explicitly a local path, so this change shouldn't be an issue.

[...]

PR #459, opened 2024-02-13
Improvement/use pathlib

kvenkman:
Following #319, Replaced all things `os.path` to use `pathlib` instead 

----
📚 Documentation preview 📚: https://earthaccess--459.org.readthedocs.build/en/459/

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -1,5 +1,4 @@
 import datetime
-from pathlib import Path
 import shutil
 import traceback
 from functools import lru_cache
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -526,7 +526,7 @@ class Store(object):
             # TODO: make this parallel or concurrent
             for file in data_links:
                 s3_fs.get(file, local_path)
-                file_name = Path(local_path, Path(file).name)
+                file_name = Path(local_path) / Path(file).name
                 print(f"Downloaded: {file_name}")
                 downloaded_files.append(file_name)
             return downloaded_files
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -442,7 +442,7 @@ class Store(object):
     def get(
         self,
         granules: Union[List[DataGranule], List[str]],
-        local_path: Optional[str] = None,
+        local_path: Optional[Path] = None,
         provider: Optional[str] = None,
         threads: int = 8,
     ) -> List[str]:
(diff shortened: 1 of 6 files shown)


earthaccess/store.py
```diff
@@ -526,7 +526,7 @@ def _get_urls(
             # TODO: make this parallel or concurrent
             for file in data_links:
                 s3_fs.get(file, local_path)
-                file_name = os.path.join(local_path, os.path.basename(file))
+                file_name = Path(local_path, Path(file).name)
```
jhkennedy, line 529:
```suggestion
                file_name = Path(local_path) / Path(file).name
```

earthaccess/store.py
```diff
@@ -572,7 +572,7 @@ def _get_granules(
             # TODO: make this async
             for file in data_links:
                 s3_fs.get(file, local_path)
-                file_name = os.path.join(local_path, os.path.basename(file))
+                file_name = Path(local_path, Path(file).name)
```
jhkennedy, line 575:
```suggestion
                file_name = Path(local_path) / Path(file).name
```

earthaccess/store.py
```diff
@@ -466,7 +466,7 @@ def get(
             List of downloaded files
         """
         if local_path is None:
-            local_path = os.path.join(
+            local_path = Path(
```
jhkennedy, line 469: `local_path`, when passed to any of the `_get` dispatch methods, is expected to be a string.

Either we should change the `_get` methods to take a Path or this needs to be made a string.

jhkennedy: Switching the called methods to taking a Path object looks straightforward, so I'd lean that way.

jhkennedy (changes requested):
@kvenkman looks good! I've added some suggestions and there's one change we'll want to make -- do you want to make the changes or want me to push changes?

earthaccess/store.py
mfisher87: >Switching the called methods to taking a Path object looks straightforward, so I'd lean that way.

I like this too :)

kvenkman: I'll update the methods to accept a Path objects - thanks for pointing that out!

kvenkman:
Thanks for the suggestions @jhkennedy - I've folded them in :smile:

earthaccess/store.py
mfisher87: Thanks for your work making _earthaccess_ better :)

mfisher87:
pre-commit.ci autofix

jhkennedy (approved):
I think this is g2g now! Thanks @kvenkman . 

I pushed a couple of minor changes to get the tests to all pass:
https://github.com/nsidc/earthaccess/compare/c68e634..453e7f9

PR #472, opened 2024-02-28
V0.9.0

### LINKED ISSUE #436: Implement EDL user agent to track usage metrics

### LINKED ISSUE #438: Restricted datasets can't be used with the API notation

### LINKED ISSUE #439: Permissions error when downloading files from OPERA_L2_RTC-S1_V1 dataset on ASF DAAC (but not OPERA_L3_DIST-ALERT-HLS_PROVISIONAL_V0 on LP DAAC)

betolink:
PR to release v0.9.0

* Bug fixes:
    * fixed #439 by implementing more trusted domains in the `SessionWithRedirection`
    * fixed #438 by using an authenticated session for `hits()`
* Enhancements:
    * addressing #427 by adding parameters to collection query
    * added user-agent in the request to track usage, closes #436

Note: I thought the bump version library was taking care of updating CHANGELOG.md but seems like that's not the case. 

----
📚 Documentation preview 📚: https://earthaccess--472.org.readthedocs.build/en/472/


jhkennedy (approved):
looks good! Do we have automation that writes the GitHub release notes and adds the contributors?

betolink:
> looks good! Do we have automation that writes the GitHub release notes and adds the contributors?

I though we had something, well find out soon.

Issue #480, opened 2024-03-04
Define user .netrc directory [Feature request]


Rapsodia86:
Hello! 
I would like to suggest an option to provide a user-specific directory to *.netrc file. 
Now, it automatically looks into "C:\\User\\USERNAME\\.netrc" (if in Windows).
However, I do have that file stored in another location that I use for my other scripts, and I do not see a reason to have it duplicated and spread across the machine. 
How about adding the option in login like:
1) Looks into the default directory for the file: `earthaccess.login(strategy="netrc")`
2) Looks into user-specified location: `earthaccess.login(strategy="netrc", dir="DIRECTORY\SUBDIRECTORY\")`

Thanks!

jhkennedy:
@Rapsodia86 would setting an environment variable work for you? For example:
```python
os.environ['NETRC_PATH'] = 'some/path/to/netrc'

earthaccess.login(strategy="netrc")
```
Would pick up your custom `netrc` at 'some/path/to/netrc'. 

That's only a three line change here, I think:
https://github.com/nsidc/earthaccess/blob/main/earthaccess/auth.py#L258

but plumbing an a new argument is a bit more work.

Rapsodia86:
Yes,
I think that is a good solution. No need to reinvent the wheel.
There is a following question. If I set `os.environ['NETRC_PATH']` in the script, but then for some reason I use  `earthaccess.login(strategy="interactive",persist=true)` will it pick up the .netrc file from that new (custom) directory or it will create a new .netrc file in the default location?

Thanks again!

jhkennedy:
@Rapsodia86 sounds good!

>  If I set os.environ['NETRC_PATH'] in the script, but then for some reason I use earthaccess.login(strategy="interactive",persist=true) will it pick up the .netrc file from that new (custom) directory or it will create a new .netrc file in the default location?

How would you like it to work? Either could be supported, but I'd lean toward if the `NETRC`/`NETRC_PATH` (whichever name we chose) variable is set, we would write/update the custom one when doing `earthaccess.login(strategy="interactive",persist=true)`

Rapsodia86:
Yes, I agree! The consistent usage of `NETRC/NETRC_PATH` will allow to avoid confusion! Great:)

mfisher87:
Excellent decision, folks ;)

chuckwondo:
[...]

PR #492, opened 2024-03-14
Add dodsrc and urs cookies files when creating netrc

battistowx:
This PR includes code that will also create a .urs_cookies and .dodsrc files when users persist their credentials to a .netrc file, and will copy the .dodsrc file based on the user operating system. It is based on code from this how-to, but adapted here to use pathlib: https://disc.gsfc.nasa.gov/information/howto?keywords=prerequisite&title=How%20to%20Generate%20Earthdata%20Prerequisite%20Files

I have tested this with poetry on both Linux/Windows, and it works great!

----
📚 Documentation preview 📚: https://earthaccess--492.org.readthedocs.build/en/492/

### DIFF
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -379,7 +379,7 @@ class Auth(object):
 
         # Create and write to .dodsrc file
         dodsrc_file = Path.home() / ".dodsrc"
-        dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_file}"
+        dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_path}"
         dodsrc_file.write_text(dodsrc_contents)
 
         if platform.system() == "Windows":
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -388,7 +388,7 @@ class Auth(object):
             local_dodsrc_path = Path.cwd() / dodsrc_path.name
             if not local_dodsrc_path.exists():
                 shutil.copy2(dodsrc_path, local_dodsrc_path)
-                
+
         return True
 
     def _get_cloud_auth_url(
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -383,12 +383,12 @@ class Auth(object):
                 f"HTTP.COOKIEJAR={urs_cookies_path}\nHTTP.NETRC={netrc_path}"
             )
             dodsrc_path.write_text(dodsrc_contents)
-       
+
         if platform.system() == "Windows":
             local_dodsrc_path = Path.cwd() / dodsrc_path.name
             if not local_dodsrc_path.exists():
                 shutil.copy2(dodsrc_path, local_dodsrc_path)
-            
+                
         return True
 
     def _get_cloud_auth_url(
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -372,7 +372,6 @@ class Auth(object):
         my_netrc = Netrc(str(netrc_path))
         my_netrc["urs.earthdata.nasa.gov"] = {"login": username, "password": password}
         my_netrc.save()
-        netrc_file = Path.home() / ".netrc"
 
         # Create and write .urs_cookies file
         urs_cookies_file = Path.home() / ".urs_cookies"
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -379,7 +379,7 @@ class Auth(object):
         # Create and write to .dodsrc file
         dodsrc_path = Path.home() / ".dodsrc"
         if not dodsrc_path.exists():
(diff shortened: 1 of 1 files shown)


earthaccess/auth.py
```diff
@@ -370,6 +372,20 @@ def _persist_user_credentials(self, username: str, password: str) -> bool:
         my_netrc = Netrc(str(netrc_path))
         my_netrc["urs.earthdata.nasa.gov"] = {"login": username, "password": password}
         my_netrc.save()
+        netrc_file = Path.home() / ".netrc"
```
jhkennedy, line 375:
```suggestion
```
This is already defined on line 366 as `netrc_path`

earthaccess/auth.py
```diff
@@ -370,6 +372,20 @@ def _persist_user_credentials(self, username: str, password: str) -> bool:
...
+
+        # Create and write .urs_cookies file
+        urs_cookies_file = Path.home() / ".urs_cookies"
+        urs_cookies_file.write_text("")
+
+        # Create and write to .dodsrc file
+        dodsrc_file = Path.home() / ".dodsrc"
+        dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_file}"
```
jhkennedy, line 383:
```suggestion
        dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_path}"
```

jhkennedy (commented):
:thinking: looks pretty good; I can see the desire for the `.urs_cookies` and `.dodsrc` (OpenDAP) files. 

As written, code will clobber (overwrite) any existing  `.urs_cookies` and `.dodsrc`  files, which I don't think is desirable, but I'll have to think more about how best to handle existing versions of these files... 

I also left a couple of minor suggestions.

battistowx:
Thank you for the reviewing and the suggestions @jhkennedy ! I added the suggested commits here. In terms of overwriting the existing dodsrc file, for Windows environments, you need to have additional copies of the dodsrc file in each working directory even if it exists in the root user folder. I'm not sure how long the information in the urs_cookies file is valid for, but I have not had any issues when overwriting it repeatedly, and this file is necessary for those who want to use wget/curl, so I have included it here for convenience.

earthaccess/auth.py
```diff
@@ -370,6 +372,19 @@ def _persist_user_credentials(self, username: str, password: str) -> bool:
...
+        urs_cookies_file = Path.home() / ".urs_cookies"
+        urs_cookies_file.write_text("")
+
+        # Create and write to .dodsrc file
+        dodsrc_file = Path.home() / ".dodsrc"
+        dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_path}"
+        dodsrc_file.write_text(dodsrc_contents)
+
+        if platform.system() == "Windows":
+            shutil.copy2(dodsrc_file, Path.cwd())
```
jhkennedy, line 386:
```suggestion
        urs_cookies_path = Path.home() / ".urs_cookies"
        if not usr_cookies_path.exists():
            urs_cookies_path.write_text("")

        # Create and write to .dodsrc file
        dodsrc_path = Path.home() / ".dodsrc"
        if not dodsrc_path.exists():
            dodsrc_contents = f"HTTP.COOKIEJAR={urs_cookies_file}\nHTTP.NETRC={netrc_path}"
            dodsrc_path.write_text(dodsrc_contents)
        
       if dodsrc_path.exists() and platform.system() == "Windows":
            shutil.copy2(dodsrc_path, Path.cwd())
            
```

Note: I changed `*_file` to `*_path` here to match the style of the rest of the file.

jhkennedy (commented):
@battistowx circling back to this. 

I agree overwriting the `.urs_cookies` path is not likely a concern, but I don't think there's a benefit to doing so. 

For `~/.dodsrc`, I think there's actually a decent amount of risk because:
1. there's a lot of configuration that *can* go into the `~/.dodsrc`, e.g.:
   * https://github.com/OPENDAP/libdap4/blob/version-3.20.6/README.dodsrc#L49-L60
   * https://docs.unidata.ucar.edu/nug/current/dap2.html
   * https://docs.unidata.ucar.edu/nug/current/dap4.html
2. Users may want to specify alternate cookie and netrc files, e.g., #480 

How about we only create them if they don't exist?

earthaccess/auth.py
jhkennedy: Only creating them if they don't exist is also in line with OpenDAP's NASA Tutorial's `.dodsrc` setup script:
https://github.com/OPENDAP/NASA-tutorials/blob/main/tutorials/setup_dodsrc.sh

battistowx: This all looks great! Thanks for pointing out the risks, I learn something new about these prerequisite files every day!

jhkennedy:
Woops! bad indent in my suggestion

battistowx:
> Woops! bad indent in my suggestion

No worries! Fixed them.

jhkennedy (approved):
Looks good to me! :shipit:

PR #452, opened 2024-02-07
replace deprecated pkg_resources with importlib.resources

danielfromearth:
----
📚 Documentation preview 📚: https://earthaccess--452.org.readthedocs.build/en/452/

Issue #374, opened 2023-11-30
Switch away from Poetry for build (to setuptools?)


mfisher87:
@MattF-NSIDC : Little experience with poetry, prefer setuptools

@jhkennedy : Has #Reasons, happy to elaborate if the group is amenable to switching

@jrbourbeau :  "I want to be able to `pip install -e .`" (NOTE: Modern `pip` can find dependencies that are specified somewhere other than `project.dependencies`, so `pip install -e .` does work with new enough version of `pip`. We think it's looking at `[build-system]` section to discover poetry dependencies.)

@betolink : "less is more"

jrbourbeau:
FWIW when I try `pip install -e .` with the latest `pip=23.3.1` release I get 

```
 Running setup.py develop for earthaccess
    error: subprocess-exited-with-error

    × python setup.py develop did not run successfully.
    │ exit code: 1
    ╰─> [1 lines of output]
        ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
        [end of output]

    note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× python setup.py develop did not run successfully.
│ exit code: 1
╰─> [1 lines of output]
    ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
    [end of output]

note: This error originates from a subprocess, and is likely not a problem with pip.
```

<details>
<summary>Full traceback:</summary>

```
(test2) ➜  earthaccess git:(remove-binder-comment) ✗ pip install -e .
Obtaining file:///Users/james/projects/nsidc/earthaccess
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting fsspec>=2022.1 (from earthaccess==0.8.1)
  Downloading fsspec-2023.12.1-py3-none-any.whl.metadata (6.8 kB)
Collecting multimethod>=1.8 (from earthaccess==0.8.1)
  Using cached multimethod-1.10-py3-none-any.whl.metadata (8.2 kB)
Collecting pqdm>=0.1 (from earthaccess==0.8.1)
  Using cached pqdm-0.2.0-py2.py3-none-any.whl (6.8 kB)
Collecting python-cmr>=0.7 (from earthaccess==0.8.1)
  Using cached python_cmr-0.9.0-py3-none-any.whl.metadata (9.4 kB)
[...]

PR #497, opened 2024-03-19
Replace pkg_resources with importlib_resources

chuckwondo:
Fix "DeprecationWarning: pkg_resources is deprecated as an API." seen when running tests.

In addition:

- increase test coverage for `formatters.py`
- replace deprecated `set-output` in `test.yml` workflow
- remove redundant lines in `Makefile`
- make minor grammatical and formatting fixes to `CONTRIBUTING.md`
- remove old, redundant "Code of Conduct" section from `CONTRIBUTING.md` and insert link to `CODE_OF_CONDUCT.md`
- fix bare links in `CODE_OF_CONDUCT.md`

----
📚 Documentation preview 📚: https://earthaccess--497.org.readthedocs.build/en/497/


mfisher87:
These doc fixes are so helpful, thank you for including them!

Makefile
mfisher87, line 1: :star_struck: :sparkles:

Makefile
```diff
@@ -14,55 +14,35 @@ HASH := $(shell git rev-parse HEAD)
...
 docs-live: ## make live docs
-docs-live:
 	bash scripts/docs-live.sh
 
-deploy-docs: ## deploy the docs, if on master branch
-deploy-docs:
-	# https://www.mkdocs.org/user-guide/deploying-your-docs/
-
-	# moved script to bash file for easy of reading
-	bash scripts/deploy-docs.sh
```
mfisher87, line 63: Should this task have been removed?

earthaccess/formatters.py
```diff
@@ -1,17 +1,28 @@
+import importlib.resources
+import sys
 from typing import Any, List
 from uuid import uuid4
 
-import pkg_resources
+STATIC_FILES = ["iso_bootstrap4.0.0min.css", "styles.css"]
 
-STATIC_FILES = ["css/iso_bootstrap4.0.0min.css", "css/styles.css"]
```
mfisher87, line 6: Why did the directory prefix need to be removed here?

mfisher87 (approved):

Makefile
chuckwondo: I removed it because `scripts/deploy-docs.sh` no longer exists, so `make deploy-docs` doesn't work.

earthaccess/formatters.py
mfisher87: Ah-hah, makes total sense now. Thanks so much for clarifying!

Makefile
mfisher87: Good eye. Thanks so much for the thorough cleanup you've done here!

mfisher87:
Woops, didn't notice workflows waiting for approval. :snail: :rocket:

mfisher87:
@chuckwondo we're just about to start a hack day, hope you'll be joining us again :)

danielfromearth:
Just linking this to the other PR that this supersedes: #452, which I think should be closed after this one is merged.

mfisher87:
I tried putting `Closes #452` in the OP but that didn't seem to work the way I wanted it to. I removed it. Can we close #452 now so we don't have to remember?

chuckwondo:
@mfisher87, I made a change that should fix the broken python 3.8 build. It also happens to simplify things a bit.

chuckwondo:
> Just linking this to the other PR that this supersedes: #452, which I think should be closed after this one is merged.

@danielfromearth, thanks for making note of this. It totally slipped my mind that I had previously commented on your issue there! This should close the loop on that.

mfisher87:
And thank you @danielfromearth for your work on this problem!

chuckwondo:
Ugh! Updated `ci/environment-mindeps.yaml` to include `importlib-resources`, which I had added to `pyproject.toml`. This seems like an accident waiting to happen (it just did). Any idea why the redundancy that forces us to keep these files in sync?

jhkennedy:
> Ugh! Updated ci/environment-mindeps.yaml to include importlib-resources, which I had added to pyproject.toml. This seems like an accident waiting to happen (it just did). Any idea why the redundancy that forces us to keep these files in sync?

Unfortunately, the environment-* YAMLs are conda environments and dependencies, while the deps in pyproject.toml are Poetry ones that get pulled from PyPI. There's no real way to avoid the duplication due to different package repositories and managers. 

And really, even if we went to just poetry environments, there's no way to specify "build an environment from the minimum dependencies" in Poetry, but we could with `uv` and only use that (dropping conda). Something to add to https://github.com/nsidc/earthaccess/issues/374

It's good to have conda env test coverage as well though as many of our users will be using conda.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ multimethod = ">=1.8"
 python-dateutil = ">=2.8.2"
 kerchunk = { version = ">=0.1.2", optional = true }
 dask = { version = ">=2022.1.0", optional = true }
+importlib-resources = "^6.3.2"
```
jhkennedy, line 47: Do we need to pin down to the patch version here?

ci/environment-mindeps.yaml
jhkennedy, line 1: probably should also add this to `ci/environment-dev.yml`

pyproject.toml
chuckwondo: Not necessarily, but I simply used `poetry add importlib-resources` to add it. There are other deps pinned similarly, but there is certainly no consistency.

This seems to point to a gap in the contributing guide regarding adding dependencies; not only _how_ to add them, but also _where_ to add them (other than `pyproject.toml`).

I'm happy to add a note in the contributing guide, if you can provide guidance.

Also, happy to adjust the version pin here accordingly, if there's a preference.

chuckwondo: Also, a brief explanation of the _why_ (in addition to the _how_ and _where_) might be good to add to the contributing guide.

Again, happy to add all such details to the contributing guide as part of this PR, if someone can provide a few sentences.

ci/environment-mindeps.yaml
chuckwondo: Ah, yes, there too. Let me know about the version constraint in the other comment, and I'll apply it consistently to all 3 spots.

pyproject.toml
jhkennedy: There's been a bit of lively discussion around dependencies in #468, but generally, a restrictive top-pin (down to the patch version) like this will force this package to break quickly for users. So we should remove the upper bound if we don't need it.
```suggestion
importlib-resources = ">-6.3.2"
```

Similarly, is there a reason for *this* version being the lower bound?

jhkennedy: > This seems to point to a gap in the contributing guide regarding adding dependencies; not only how to add them, but also where to add them (other than pyproject.toml).

Yes, agreed, we should add that. Especially since `poetry add` adds upper bounds to dependencies with `^`, which is bad for libraries (and will not change https://github.com/python-poetry/poetry/issues/3427)

chuckwondo: Ah, thanks for the link to that discussion. I'll definitely dig into that to understand the thinking.

In the meantime, what can I add to the contributing guide as the basic steps for a contributor to take when adding a new dependency? What, where, and why (all very briefly)?

chuckwondo: For the "what" and "where", something like this?

1. Add the dependency to `pyproject.toml` with `poetry`: `poetry add LIB` (or `poetry add --group=dev` for a development dependency)
2. On the line added to `pyproject.toml` by `poetry`, change `^` to `>=` (perhaps add a "why" bit here)
3. For a development dependency, add the same dependency and constraint to `ci/environment-dev.yaml`
4. Add the same dependency to `ci/environment-mindeps.yaml`. For a required dep, pin the exact version, but for a testing dep, leave unpinned (again, a "why" comment here might also prove helpful)

chuckwondo: > There's been a bit of lively discussion around dependencies in #468, but generally, a restrictive top-pin (down to the patch version) like this will force this package to break quickly for users. So we should remove the upper bound if we don't need it.
> 
> Similarly, is there a reason for _this_ version being the lower bound?

Nope, no particular reason. That's simply what `poetry` inserted. Do we want something different?

ci/environment-mindeps.yaml
chuckwondo: Actually, this is not a dev dep, so I don't think we want to add to `ci/environment-dev.yml`, do we? If we do, then perhaps that further reinforces the need for explanation of dep mgmt in the contributing guide.

pyproject.toml
jhkennedy: > Nope, no particular reason. That's simply what poetry inserted. Do we want something different?

I think it's fine for now. Lower bounds are less problematic overall (except when they encounter incompatible upper bounds) and it's a few versions behind the latest so should be decently inclusive

chuckwondo: > Those steps sound right to me. For (1) I'd be inclined to have the users just manually add it to the `pyproject.toml` instead of having to tell them to effectively manually do it again in (2), but (1) does update the lockfile too.
> 
> For 2 we could tell them to use [poetry-relax](https://github.com/zanieb/poetry-relax?tab=readme-ov-file#references) but that should probably be done in isolation with its own PR later.

My reasoning for letting poetry add the entry to pyproject.toml is because it automatically finds a version compatible with the rest of the deps, eliminating any need to manually go looking for a version number.

chuckwondo: > Those steps sound right to me. For (1) I'd be inclined to have the users just manually add it to the `pyproject.toml` instead of having to tell them to effectively manually do it again in (2), but (1) does update the lockfile too.
> 
> For 2 we could tell them to use [poetry-relax](https://github.com/zanieb/poetry-relax?tab=readme-ov-file#references) but that should probably be done in isolation with its own PR later.

Cool. Can you provide the "why" in the 2 spots I mentioned? Just a sentence or 2 in each spot to provide context for contributors.

jhkennedy: > What's "a few versions behind the latest"? If you're referring to importlib-resources, that version is the latest as of earlier today.

Bah, I thought I looked earlier and latest was 7.0.2; no idea what I looked at :tired_face:

jhkennedy: @chuckwondo this is getting large -- perhaps we should move the contributing guide discussion to its own issue and subsequent PR?

jhkennedy: As for the version, since we're still a poetry project, let's just go with:
```
importlib-resources = ">=6.3.2"
```
for now since that's what it decided and the rest were likely added that way.

ci/environment-mindeps.yaml
jhkennedy: @chuckwondo oh, you're right, that file doesn't set up a full dev environment, just boostraps poetry and some deps for unclear reasons; fine to ignore for now.

pyproject.toml
jhkennedy: As for the why:

> On the line added to pyproject.toml by poetry, change ^ to >= (perhaps add a "why" bit here)

because upper bounds can easily result in unsolvable user environments for Python libraries like Earthaccess. For more information, see the [poetry-relax README](https://github.com/zanieb/poetry-relax), especially the [references](https://github.com/zanieb/poetry-relax?tab=readme-ov-file#references).

> Add the same dependency to ci/environment-mindeps.yaml. For a required dep, pin the exact version, but for a testing dep, leave unpinned (again, a "why" comment here might also prove helpful)

I don't actually know the reasoning here... looks like it's because we pip install with `--no-deps` and only use the dependencies in the conda environment, so we need all dependencies reflected here. @mfisher87 @betolink got a sentence for this?

chuckwondo:
@jhkennedy and @mfisher87, I believe I addressed the dependency pinning issue. I also added a small "Managing Dependencies" section to the contributing guide. Please let me know if I missed anything.

jhkennedy (approved):
looks good to me. @mfisher87 did you want to tak another look?

chuckwondo:
@mfisher87, does this look good?

chuckwondo:
@jhkennedy can you take a look at how/why one of the checks was cancelled and rerun? I see, "Canceling since a higher priority waiting request for 'refs/pull/497/merge' exists," but can't tell exactly what caused it.

jhkennedy:
> @jhkennedy can you take a look at how/why one of the checks was cancelled and rerun? I see, "Canceling since a higher priority waiting request for 'refs/pull/497/merge' exists," but can't tell exactly what caused it.

I'm confused by that as well. I believe it's related to the concurrency setting:
https://github.com/nsidc/earthaccess/blob/main/.github/workflows/test-mindeps.yml#L8-L12

But not sure why. I don't have time to dig in right now (it works in some PRs but not others), but I'll open an issue.

For now, I think it's safe to ignore it as the test ran fine in your fork:
https://github.com/chuckwondo/earthaccess/actions/runs/8367002850

mfisher87:
Hey, so sorry all for missing your mentions! :eyes:

mfisher87 (approved):
LGTM! I have lost some context in my delay to respond to this issue and I apologize for that :crying_cat_face:

chuckwondo:
Can somebody merge this PR?

jhkennedy:
@chuckwondo done!

chuckwondo:
@mfisher87, thanks for the offer for merge permissions. I don't yet feel comfortable being granted such permissions, but perhaps after a few more PRs and "hack days" I'll take you up on the offer.

mfisher87:
I'll be sure to ask again later, then! ;)

PR #505, opened 2024-04-01
Fix erroneous workflow cancellation

### LINKED ISSUE #503: test-mindeps erroniously cancelling test-mindeps jobs
In https://github.com/nsidc/earthaccess/pull/497, the [test-mindeps](https://github.com/nsidc/earthaccess/blob/main/.github/workflows/test-mindeps.yml) GitHub Action workflow got canceled and so shows up as "not successful":
image

However, it did run successfully in @chuckwondo's fork:
https://github.com/chuckwondo/earthaccess/actions/runs/8367002850

This is likely related to the workflow's concurrency setting:
https://github.com/chuckwondo/earthaccess/actions/runs/8367002850

But it seems to work fine in other PRs, e.g.:
https://github.com/nsidc/earthaccess/pull/500

I'm not sure why this is, and I don't have time right now to dig in, but should be a good candidate for a new contributor familiar with CI/CD stuff.

chuckwondo:
The concurrency groups for both integration-tests and test-mindeps were named the same, creating a race condition such that whichever workflow was started first would be cancelled by the start of the second workflow. This fix gives them distinct group names to avoid erroneously cancelling one or the other.

Fixes #503

----
📚 Documentation preview 📚: https://earthaccess--505.org.readthedocs.build/en/505/

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -14,7 +14,7 @@ on:
 # When this workflow is queued, automatically cancel any previous running
 # or pending jobs from the same branch
 concurrency:
-  group: ${{ github.ref }}
+  group: integration-tests-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


jhkennedy (approved):

jhkennedy:
Ah, nice! Thanks @chuckwondo

mfisher87:
Great eye!

Issue #504, opened 2024-03-29
Improve polygon coordinate hadnling in search_data


jhkennedy:
Note: Reported by @iamdonovan in chat

[`search_data`](https://earthaccess.readthedocs.io/en/latest/user-reference/api/api/#earthaccess.api.search_data) allows you to specify a `polygon=` keyword argument, who's value is handled by:
https://github.com/nsidc/earthaccess/blob/v0.9.0/earthaccess/search.py#L734-L742

and passed to python-cmr's `polygon` method:
https://github.com/nasa/python_cmr/blob/develop/cmr/queries.py#L458

python-cmr expects a list of (lat, lon) tuples, in clockwise order, as documented in the CMR API docs:
https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#g-polygon

**However**, there is very little handling along the way to ensure we meet that specification. 

It would be very natural to do something like:
```python
search_area = shapely.Polygon(...)   # some shapely Geometry

earthdata.search(polygon=search_area, ...)
```

But right now, you'd need to do:
```python
earthdata.search(polygon=search_area.coords, ...)
```
which  may end up resulting in:
```
RuntimeError: {"errors":["The polygon boundary points are listed in the wrong order. Points must be provided in counter-clockwise order."]}
```

We currently have to do:
```python
exterior = search_area.exterior
if not shapely.is_ccw(exterior)
    exterior = exterior.reverse()

earthdata.search(polygon=exterior.coords, ...)
```

or
```python
polygon = shapely.geometry.polygon.orient(search_area, sign=1).exterior.coords
earthdata.search(polygon=polygon, ...)
```

**Ideally**, we'd accept any of the following:
* valid WKT string
* valid WKB string
* GDAL OGR Geometry
* Shapely Geometry
* List of coordinate tuples

and ensure we provide the correctly oriented coordinate list to python-cmr.

---

Additionally, our `search_data` documentation could be improved. Currently, we imply, by linking out to the [CMR API docs](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#g-spatial), that polygon (and point, line, and circle) is a search keyword argument, but all our examples tend to use a bounding box, and the only geometric parameter we do list is `bounding_box`:
![image](https://github.com/nsidc/earthaccess/assets/7882693/d8036e55-6bc2-4243-81e5-cec87d60ec6c)

https://earthaccess.readthedocs.io/en/latest/user-reference/api/api/#earthaccess.api.search_data

[...]

PR #506, opened 2024-04-01
improve polygon coordinate handling

iamdonovan:
As detailed/explained/requested in #504:

- Added support for `shapely` geometry, WKT strings, and WKB geometries
- checks that polygon coordinates are correctly oriented before passing to CMR
- added `shapely>=2.0.0` as a dependency to `pyproject.toml`

----
📚 Documentation preview 📚: https://earthaccess--506.org.readthedocs.build/en/506/


earthaccess/search.py
```diff
@@ -731,14 +732,31 @@ def point(self, lon: str, lat: str) -> Type[GranuleQuery]:
         super().point(lon, lat)
         return self
 
-    def polygon(self, coordinates: List[Tuple[str, str]]) -> Type[GranuleQuery]:
+    def polygon(
+        self,
+        coordinates: Union[
+            shapely.geometry.polygon.Polygon, str, bytes, List[Tuple[str, str]]
```
jhkennedy, line 738:
```suggestion
            shapely.geometry.base.BaseGeometry, str, bytes, List[Tuple[str, str]]
```
I think since the rest of the implementation works of off the BaseGeometry, we could just use that here as well. We probably want to add some tests with different geometries and especially a clockwise vs couterclockwise ones

iamdonovan: Done (https://github.com/nsidc/earthaccess/pull/506/commits/583de40d931281bd149bf92afc81cdb5dca3347c). Will think about/work on tests later this week/next week.

earthaccess/search.py
```diff
@@ -731,14 +732,31 @@ def point(self, lon: str, lat: str) -> Type[GranuleQuery]:
...
 
         Parameters:
-            coordinates: list of (lon, lat) tuples
+            coordinates: a shapely geometry, WKT or WKB, or a list of (lon, lat) tuples
         """
-        super().polygon(coordinates)
+
+        if isinstance(coordinates, shapely.geometry.base.BaseGeometry):
+            geom = coordinates
+        elif isinstance(coordinates, list):
```
chuckwondo, line 750: Using `list` is too restrictive. This should likely be `Sequence`.

asteiker:
Hi @iamdonovan I'm going through our backlog of issues and see that this draft PR was created some time ago, so I wanted to check in on it. @chuckwondo You probably have the best sense of whether we've implemented other search improvements that would affect this PR. Please let us know how we can help move this forward, if still applicable!

PR #516, opened 2024-04-11
add obdaac s3-credential endpoint

itcarroll:
OB.DAAC is joining the PARTY!

----
📚 Documentation preview 📚: https://earthaccess--516.org.readthedocs.build/en/516/

### DIFF
--- a/earthaccess/daac.py
+++ b/earthaccess/daac.py
@@ -71,7 +71,7 @@ DAACS = [
         "short-name": "OBDAAC",
         "name": "NASA's Ocean Biology Distributed Active Archive Center",
         "homepage": "https://oceancolor.gsfc.nasa.gov/",
-        "cloud-providers": ["OB_DAAC"],
+        "cloud-providers": ["OB_CLOUD"],
         "on-prem-providers": ["OB_DAAC"],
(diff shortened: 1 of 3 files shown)


jhkennedy (dismissed):
Looks good to me! I confirmed the homepage change and that the s3-redentials endpoint README is there:
https://obdaac-tea.earthdatacloud.nasa.gov/s3credentialsREADME

itcarroll:
Thanks! Waiting to unset `draft` to get a nod from our programmers.

quixrick:
This looks okay to me.  👍  
-Rick

itcarroll:
@jhkennedy Your review (which, while appreciated, came earlier than desired) was abrogated when I unset `draft`. Ready for review now, please & thank you.

betolink (approved):
Nice! just a note, S3 credentials are also picked up from the metadata if they are available. Eventually these hard-coded values should go away, in real life I don't know how good metadata will be from CMR cc @jhkennedy

itcarroll:
@betolink These hard-coded values would still be needed for `earthaccess.get_s3fs_session(daac="OBDAAC")` though, right? Maybe that would go away too?

betolink:
Yeah they are still needed for that. Only when we use the results to download or open is when we don't really need them if the metadata is right. I guess they won't go away anytime soon.

PR #520, opened 2024-04-12
update the dependency and temporal filter for python_cmr 0.10.0

### LINKED ISSUE #190: H:M:S for Date_To in the Temporal function defaults to 0:0:0
The H:M:S for the Date_To parameter in the Temporal function defaults to 0:0:0 when only the date is supplied (e.g. 2020-01-01 defaults to 2020-01-01T00:00:00). I think the H:M:S for the Date_To parameter should default to 23:59:59 unless specified by the user. This would match the behavior of NASA's Earthdata Search.

### LINKED ISSUE #330: dateutil parser breaks search for dates provided in ISO 8601 format by adding timezone offset
Searching with dates in ISO 8601 format (YYYY-MM-DDThh:mm:ssZ) fails due to the dateutil parser adding the timezone offset before sending to the CMR query. The CMR query then fails because the date string is no longer matches the ISO 8601 format that the date was originally in.

Example:

```
#!/usr/bin/env python 

import earthaccess

COLLECTION_SHORTNAME = 'SPURS2_ARGO'
START_DATE = '2019-03-10T00:00:00Z'
END_DATE = '2019-03-11T23:59:59Z'

auth = earthaccess.login()

cmr_search = earthaccess.DataGranules(auth). \
        short_name(COLLECTION_SHORTNAME).temporal(START_DATE, END_DATE)

results = cmr_search.get()
```

Fails with the following error:
```
Traceback (most recent call last):
  File "/lib/python3.11/site-packages/cmr/queries.py", line 357, in convert_to_string
    return date.strftime(iso_8601)
           ^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'strftime'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/projects/testearthaccess.py", line 12, in <module>
    short_name(COLLECTION_SHORTNAME).temporal(START_DATE, END_DATE)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/earthaccess/search.py", line 611, in temporal
    super().temporal(parsed_from, parsed_to, exclude_boundary)
  File "/lib/python3.11/site-packages/cmr/queries.py", line 368, in temporal
    date_from = convert_to_string(date_from)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/cmr/queries.py", line 361, in convert_to_string
    datetime.strptime(date, iso_8601)
  File "/usr/local/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/_strptime.py", line 349, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '2019-03-10T00:00:00+00:00Z' does not match format '%Y-%m-%dT%H:%M:%SZ'
```

itcarroll:
Version 0.10.0 of the `python_cmr` package correctly handles a broader range of date-like inputs for the temporal filters, due to our input via nasa/python_cmr#30. This PR increments our version dependency, strips our own preprocessing of date-like inputs, and modifies tests accordingly.

Fixes #190
Fixes #330 

----
📚 Documentation preview 📚: https://earthaccess--520.org.readthedocs.build/en/520/


itcarroll:
Hmm... `make format lint` is not catching whatever is breaking here.

ci/environment-mindeps.yaml
```diff
@@ -26,3 +24,6 @@ dependencies:
   - types-setuptools
   - ruff
   - vcrpy
+  - pip
+  - pip:
+      - python-cmr==0.10.0
```
mfisher87, line 29: v0.10.0 is now on conda-forge, so we can move `python-cmr` out of the `pip` subsection :)

earthaccess/search.py
```diff
@@ -310,34 +309,24 @@ def daac(self, daac_short_name: str = "") -> Type[CollectionQuery]:
...
-                print("The provided end date was not recognized")
-                date_to = ""
+        """Filter by an open or closed date range. Dates can be provided as date objects
+        or ISO 8601 strings. Multiple ranges can be provided by successive method calls.
 
+        ???+ Tip
+            Giving either `datetime.date(YYYY, MM, DD)` or `"YYYY-MM-DD"` as the `date_to`
+            parameter includes that entire day (i.e. the time is set to `23:59:59`).
+            Using `datetime.datetime(YYYY, MM, DD)` is different, because `datetime.datetime`
+            objects have `00:00:00` as their built-in default.
```
mfisher87, line 323: :100:

mfisher87 (commented):
LGTM!

earthaccess/search.py
```diff
@@ -310,34 +309,24 @@ def daac(self, daac_short_name: str = "") -> Type[CollectionQuery]:
 
     def temporal(
         self,
-        date_from: Optional[Union[str, dt.datetime]] = None,
-        date_to: Optional[Union[str, dt.datetime]] = None,
+        date_from: Optional[Union[str, dt.date]] = None,
+        date_to: Optional[Union[str, dt.date]] = None,
```
jhkennedy, line 313:
```suggestion
        date_from: Optional[Union[str, dt.date, dt.datetime]] = None,
        date_to: Optional[Union[str, dt.date,, dt.datetime]] = None,
```
I know it's not necessary or even correct since datetime is a subclass of date, but I think from a documentation standpoint this will be clearer to users.

earthaccess/search.py
```diff
@@ -616,34 +605,24 @@ def debug(self, debug: bool = True) -> Type[GranuleQuery]:
...
```
jhkennedy, line 609:
```suggestion
        date_from: Optional[Union[str, dt.date, dt.datetime]] = None,
        date_to: Optional[Union[str, dt.date,, dt.datetime]] = None,
```

tests/unit/test_collection_queries.py
```diff
@@ -5,14 +5,19 @@
 from earthaccess.search import DataCollections
 
 valid_single_dates = [
-    ("2001-12-12", "2001-12-21", "2001-12-12T00:00:00Z,2001-12-21T00:00:00Z"),
+    ("2001-12-12", "2001-12-21", "2001-12-12T00:00:00Z,2001-12-21T23:59:59Z"),
```
jhkennedy, line 8: :thinking: is it worth adding some non-utc timestamps like `'2024-04-18T22:13:21.469360-08:00'`?

jhkennedy (dismissed):
LGTM! Approving with some take-or-leave comments.

mfisher87 (dismissed):
Woops thought I approved before but looks like I accidentally selected "comment"!

chuckwondo:
@itcarroll, in addition to addressing @jhkennedy's comments, you'll need to resolve conflicts coming from the main branch.

tests/unit/test_collection_queries.py
itcarroll: Added a utc offset test.

earthaccess/search.py
itcarroll: Currently the type hint does not effect the documentation, because the docstring includes a "type". Do you want that part removed from the docstring?

If type hints are being included to help developers, then I would want them to be technically correct. If type hints are being included to help users ... well, I fell like that's a little lazy actually. Users don't benefit a whole lot from "dt.date" without knowing what "dt" is. And do they know what "Optional[Union[...]]" means? Relying on auto documentation is nice ... if it's good documentation.

You are welcome to modify however you think its best.

chuckwondo: Personally, when type annotations are provided, I drop type information from the docstrings. Otherwise, it's all too easy for the docstrings to become out of sync with the type annotations, leading to confusion, misinformation, and frustration. Modern doc generators, including mkdocs, will make use of the type annotations to put such information in the docs. Doc generators should also use canonical names for things as well, meaning that even though the code may refer to the `datetime` module as `dt`, the docs will show the name `datetime`, so there's no need for the reader to know about any aliasing within the code.

chuckwondo:
@itcarroll, let's schedule a pairing session so I can help you sort out the build failures you're encountering. Regarding the vcrpy failures, it might behoove us to wait for @doug-newman-nasa's PR (#528) to land beforehand, so you don't accidentally commit an access token like I did.

chuckwondo:
@itcarroll, I just merged @doug-newman-nasa's PR for #528, so I'm going to pick up this PR now to iron out the kinks.

chuckwondo:
@itcarroll, I see that you added `lxml-html-clean` as a dev dependency, but I don't see any reference to it in the code, nor in any docs, unless I'm missing something. What did you add it for?

itcarroll:
Darn, I meant to leave a comment on that and now I can't find the relevant issue. Roughly ....

When I updated poetry.lock for python-cmr 0.10, an [update to lxml](https://lxml.de/5.2/changes-5.2.0.html) caused document generation to fail. The nbconvert we are pinned to by mkdocs-jupyter seemed unlikely to be updated to express a new version constraint on lxml. Since it's only a dev dependency, I hoped a direct import would be an acceptable solution.

chuckwondo:
No worries. That's sufficient explanation. Thanks!

chuckwondo:
@itcarroll, please take a look at https://github.com/nsidc/earthaccess/pull/546

That's the PR I just submitted based upon your original PR. I removed the `lxml-html-clean` dep because I simply reduced the changes to `poetry.lock` to avoid the issue.

Let me know if everything looks good. If so, then I'll have you close your PR.

itcarroll:
Thank you very much @chuckwondo, looks great. Please proceed!

PR #524, opened 2024-04-16
Update contributing doc from issues surfaced in user testing

mfisher87:
----
📚 Documentation preview 📚: https://earthaccess--524.org.readthedocs.build/en/524/


CONTRIBUTING.md
```diff
@@ -37,6 +39,10 @@ and now we can run the local Jupyter Lab and run the scripts etc. using Poetry:
 poetry run jupyter lab
 ```
 
+!!! note
+
+    You may need to use `poetry run make ...` to run commands in the environment.
+
```
chuckwondo, line 45: Your original syntax is not valid. You want this:

```suggestion
> [!NOTE]
> You may need to use `poetry run make ...` to run commands in the environment.
```

CONTRIBUTING.md
```diff
@@ -62,34 +68,46 @@ Finally, for _development dependencies only_, you must add an entry to
...
+    - `make test`
+
+### ...to fix an issue or bug
 
-- Read the documentation (working on adding more)
-- create the minimally reproducible issue
-- try to edit the relevant code and see if it fixes it
-- submit the fix to the problem as a pull request
-- include an explanation of what you did and why
+- Create a GitHub issue with a minimally reproducible example (TODO: Link to SSCCE?)
```
chuckwondo, line 81: Should this be a draft PR until you address this TODO?

CONTRIBUTING.md
```diff
@@ -62,34 +68,46 @@ Finally, for _development dependencies only_, you must add an entry to
...
-- run pytest to ensure your local version of code passes all unit tests
-- try to edit the relevant code and implement your new feature in a backwards compatible manner
-- create new tests as you go, and run the test suite as you go
-- update the documentation as you go
+- Create an issue and discuss the feature's scope and its fit for this package with the team
+- Create a branch for your new feature in your fork
+- Run the unit tests successfully in your branch
+- Write the code to implement your new feature in a backwards compatible manner
+- Create at least one test that exercises your feature and run the test suite as you go
+    - Example: _TODO: Add a good example of a small PR with a feature and a test._
```
chuckwondo, line 95: Should this be a draft PR until you address this TODO?

chuckwondo (changes requested):

CONTRIBUTING.md
chuckwondo: Sorry, scratch that. The syntax is invalid for GitHub Markdown, which is what my "correction" addresses, but I see that your syntax is for the generated docs.

It's a shame that this does not also render properly on GitHub. Do we need to determine how to cleanly separate such things?

mfisher87: Yes, this is one of my pet peeves. Same with Quarto, admonitions are rendered as

```
:::{.callout-note}
My note
:::
```

And Sphinx:

```
.. note::

   My note
```

I really dislike how GitHub invented their own syntax for this on top of the fifty we already have.

mfisher87: I think we should make `/CONTRIBUTING.md` simply a link to a `docs/contributing/index.md`. Could do that as part of this PR to resolve this. What do you think?

CONTRIBUTING.md
mfisher87: For sure for this one. There's "competitors" to sscce.org, do you have strong feelings about any of them? I figured I'd google around and link to whatever seems most popular.

CONTRIBUTING.md
mfisher87: #530

CONTRIBUTING.md
mfisher87: Alright, picking one has been less than successful. I vote we use the term ["minimal reproducible example"](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and consider all these other things to be "brand names" of effectively the same thing. Choosing one seems to be folly. We can link to the wikipedia page and let people follow through to more resources?

* MCVE: [Stack overflow has an excellent page MCVE page](https://stackoverflow.com/help/minimal-reproducible-example), but I'm not certain if SO is the originator or what.
* SSCCE: Has an official <http://www.sscce.org> website. If you hit it with HTTPS by accident, the site doesn't work! Abbreviation is hard to remember.
* reprex: I think of <https://reprex.tidyverse.org/> as the "official site". The term somehow feels more jargony to me than an acronym :laughing: Just a feeling though.

mfisher87: 6fcaacd

CONTRIBUTING.md
mfisher87: 8de415d

mfisher87: I didn't think about how much harder this would make review. Apologies.

andypbarrett (approved):

PR #528, opened 2024-04-19
Remove token

doug-newman-nasa:
The new cassette for the test_data_links has some information in it pertaining to the user who executed the test. This PR redacts it.
BTW - at the hackathon I stated that I was going to add a VCR cassette for test_data_links which I attempted to PR today only to find the same work had been merged three days ago. We need to coordinate better.

----
📚 Documentation preview 📚: https://earthaccess--528.org.readthedocs.build/en/528/


chuckwondo (changes requested):
Thanks Doug! My bad on the token issue, and for failing to coordinate. It slipped my mind that you had mentioned this, but now I recall discussing with @mfisher87 the need to handle this very situation.

However, instead of simply hand-modifying the cassette, I'd like to see proper request and response filtering. Without using the vcrpy capabilities for scrubbing sensitive information, if someone were to re-record this cassette, they would run the risk of committing sensitive information again, if they didn't realize (like me) that sensitive information was being recorded.

See:

- [Filter sensitive data from the request](https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-sensitive-data-from-the-request)
- [Custom Response Filtering](https://vcrpy.readthedocs.io/en/latest/advanced.html#custom-response-filtering)

doug-newman-nasa:
Yup. I looked into using
`with my_vcr.use_cassette('test.yml', filter_post_data_parameters=['api_key']):
    requests.post('http://api.com/postdata', data={'api_key': 'secretstring'})`
but I couldn't get it working properly.

chuckwondo:
> Yup. I looked into using `with my_vcr.use_cassette('test.yml', filter_post_data_parameters=['api_key']): requests.post('http://api.com/postdata', data={'api_key': 'secretstring'})` but I couldn't get it working properly.

Ok. I'll take a look. If I can't sort it out in short order, I'll let you know and we can approve this redaction in the meantime. If that's the case, then I'll open a new issue to later address this via vcrpy functionality.

doug-newman-nasa:
Sorry, I should have been more specific. I can keep looking at it. Just not immediately.

jhkennedy:
> BTW - at the hackathon I stated that I was going to add a VCR cassette for test_data_links which I attempted to PR today only to find the same work had been merged three days ago. We need to coordinate better.

One way that might help here is to put PRs into, or back into, draft status if there's planned future work but the PR appears to be in ready-for-review or ready-to-merge state.

doug-newman-nasa:
This is a 'preliminary pull request' because I wanted to get your eyes on the `redact_key_values` code. Not being a python coder I'm sure it requires some love. I'm going to work on redacting `uri: https://urs.earthdata.nasa.gov/api/users/dschuck?client_id=`

tests/unit/test_results.py
```diff
@@ -41,6 +42,48 @@ def _get_vcr(self, **kwargs):
             "Set-Cookie",
             "User-Agent",
         ]
+        myvcr.filter_query_parameters = [
+            ("client_id", "foo"),
```
chuckwondo, line 46: I'm guessing `"foo"` will change once you address the `"uri"` redacting?

tests/unit/test_results.py
```diff
@@ -41,6 +42,48 @@ def _get_vcr(self, **kwargs):
...
+        ]
+
+        def redact_key_values(keys_to_redact):
```
chuckwondo, line 49: Nice use of a higher-order function here! I'd suggest just making this a top level function. No need to nest this within the `_get_vcr` method, nor within the `TestResults` class at all.

tests/unit/test_results.py
```diff
@@ -41,6 +42,48 @@ def _get_vcr(self, **kwargs):
...
+                string_body = response["body"]["string"].decode("utf8")
+                if "REDACTED" not in string_body:
+                    # Only do this is if the body contains one or more
+                    # of the keys to redact.
+                    if any(key in string_body for key in keys_to_redact):
+                        try:
+                            # Marshall into json object, if it is a JSON object.
+                            payload = json.loads(string_body)
+                            if isinstance(payload, list):
+                                payload = payload[0]
```
chuckwondo, line 61: Is there a particular situation where we expect the payload to be a list? If so, why would we take only the first element of the list?

tests/unit/test_results.py
```diff
@@ -41,6 +42,48 @@ def _get_vcr(self, **kwargs):
...
+                            for key in keys_to_redact:
+                                if key in payload:
+                                    # Redact the key value
+                                    payload[key] = "REDACTED"
+                            # Write out the updated json object to the response
+                            # body string.
+                            response["body"]["string"] = json.dumps(payload).encode()
+                        except ValueError:
+                            # If it is not a json object, return
+                            return response
```
chuckwondo, line 71: If you `import contextlib`, you can use the `suppress` context manager to simplify things a bit. In this case, if a `JSONDecodeError` is raised, the context manager is simply exited and execution continues at the line following the context manager block. In this case, it's the subsequent `return response` line, which is what you want to do anyway.

Note that I replaced `ValueError` with the more specific `JSONDecodeError` to avoid potentially suppressing a `ValueError` that might signify a bug.

Also, I removed the comments, as they essentially simply repeated what is fairly obvious from the code itself.

```suggestion
                        with contextlib.supress(json.JSONDecodeError):
                            payload = json.loads(string_body)
                            if isinstance(payload, list):
                                payload = payload[0]
                            for key in keys_to_redact:
                                if key in payload:
                                    payload[key] = "REDACTED"
                            response["body"]["string"] = json.dumps(payload).encode()
```

tests/unit/test_results.py
doug-newman-nasa: Some responses from EDL are JSON wrapped in a list which is not parsing via json.loads. This is a hack which I'd like some python expertise on. How do I handle this elegantly?

chuckwondo: Ah, I see that you're getting back a list from the /tokens endpoint, which makes sense. I see the problem you're facing, so let me see if I can find a convenient means for inserting my suggested code adjustments.

tests/unit/test_results.py
```diff
@@ -41,6 +45,64 @@ def _get_vcr(self, **kwargs):
...
+                            # Write out the updated json object to the response
+                            # body string.
+                            if is_list:
+                                payload = [payload]
+                            response["body"]["string"] = json.dumps(payload).encode()
+                        except ValueError:
+                            # If it is not a json object, return
+                            return response
+
+                return response
```
chuckwondo, line 80: To simplify this logic a bit, I suggest _not_ bothering to avoid redacting an already redacted payload. This redaction is idempotent, so redacting an already redacted payload yields the same result, and in the context of running unit tests, the performance impact is likely unnoticeable.

```suggestion
            def redact(payload):
                for key in keys_to_redact:
                    if key in payload:
                        payload[key] = "REDACTED"
                return payload

            def before_record_response(response):
                body = response["body"]["string"].decode("utf8")

                with contextlib.suppress(json.JSONDecodeError):
                    payload = json.loads(body)
                    redacted_payload = map(redact, payload) if isinstance(payload, list) else redact(payload)
                    response["body"]["string"] = json.dumps(redacted_payload).encode()

                return response
```

Also, please move the entirety of `redact_key_values` to the top level. It does not need to be nested with the `_get_vcr` method. The nesting just adds unnecessary clutter.

Same goes for the nested `redact_login_request` below.

doug-newman-nasa: Oddly, I move this method out of TestCase it no longer gets called. Investigating.

doug-newman-nasa:
I've made all the suggestions changes with the exception of moving `redact_key_values` out of `TestCase` which prevents it from being called. Looking into that next.

chuckwondo:
@doug-newman-nasa, I put these at the top level in `tests/unit/test_results.py` (preceding `class TestResults`), and everything works fine for me:

```python
def redact_login_request(request):
    if "/api/users/" in request.path and "/api/users/tokens" not in request.path:
        _, user_name = os.path.split(request.path)
        request.uri = request.uri.replace(user_name, REDACTED_STRING)

    return request


def redact_key_values(keys_to_redact):
    def redact(payload):
        for key in keys_to_redact:
            if key in payload:
                payload[key] = REDACTED_STRING
        return payload

    def before_record_response(response):
        body = response["body"]["string"].decode("utf8")

        with contextlib.suppress(json.JSONDecodeError):
            payload = json.loads(body)
            redacted_payload = (
                list(map(redact, payload))
                if isinstance(payload, list)
                else redact(payload)
            )
            response["body"]["string"] = json.dumps(redacted_payload).encode()

        return response

    return before_record_response
```

Give it a shot and let me know what you see.

Also, I'd like to request a minor tweak to an assertion in the function `test_get_all_more_than_2k`. Because the number of granules is changing, if someone decide to regenerate all of the cassettes, this test will fail due to a difference in the length of the granules list.

So instead of this assertion:

```python
        self.assertEqual(len(granules), 2520)
```

let's do this, which will be less strict, but still validate the results correctly:

```python
        self.assertIn(len(granules), range(2001, 3001))
```

doug-newman-nasa:
Regarding `test_get_all_more_than_2k` what I am trying to test here is that the correct number of granules were marshaled from the result. Using a range comparison will limit the effectiveness of that test. However, I'm not actually testing against the value _in the result_ rather the result I fished out manually. So I'm going to counter by asserting 
len(granules) against the value of 'hits' in the response 
`string: '{"hits":2520,"took":138,"items":[]}'`
Does that work?

chuckwondo:
> Regarding `test_get_all_more_than_2k` what I am trying to test here is that the correct number of granules were marshaled from the result. Using a range comparison will limit the effectiveness of that test. However, I'm not actually testing against the value _in the result_ rather the result I fished out manually. So I'm going to counter by asserting equality of `len(granules)` against the value of 'hits' in the response `string: '{"hits":2520,"took":138,"items":[]}'` Does that work?

Perfect! Even better!

chuckwondo (approved):
Fantastic! Thanks Doug!

chuckwondo:
I've approved this, but does anybody else want to review this?

BTW, I'd like to suggest that we squash merge PRs to have a tidier, more readable change history.

tests/unit/test_results.py
```diff
@@ -118,7 +172,9 @@ def test_get_all_more_than_2k(self):
 
         # Assert that we performed a hits query and two search results queries
         self.assertEqual(len(self.cassette), 3)
-        self.assertEqual(len(granules), 2520)
+        self.assertEqual(
+            len(granules), int(self.cassette.responses[0]["headers"]["CMR-Hits"][0])
+        )
```
chuckwondo, line 177: @doug-newman-nasa, I just realized that we actually need this to be a combination of this plus what I suggested earlier. The issue with this assertion is that if this cassette is ever recreated at a point where the response would return a value for CMR-Hits > 3000, the test would then fail because it specifies count=3000. What we want is for the number of granules to match the min of 3000 and CMR-Hits:

```suggestion
        self.assertEqual(
            len(granules), min(3000, int(self.cassette.responses[0]["headers"]["CMR-Hits"][0]))
        )
```

chuckwondo (changes requested):
Sorry Doug. One last tweak, please.

chuckwondo:
Last call for someone else to review. If I don't get any takers today, I'll proceed with a squash merge near COB US Eastern.

PR #546, opened 2024-04-30
Directly use `python_cmr`'s `temporal` methods.

### LINKED ISSUE #190: H:M:S for Date_To in the Temporal function defaults to 0:0:0

chuckwondo:
Bump `python_cmr` version to make use of additional logic in the `temporal` methods that were pushed from `earthaccess` up to `python_cmr` itself.

Fixes #190 #330

----
📚 Documentation preview 📚: https://earthaccess--546.org.readthedocs.build/en/546/


chuckwondo:
@betolink or @mfisher87, would one of you be able to give this a quick review?

This was originally from @itcarroll, who handed things off to me to fix the build issues he was encountering.

The gist of it is an upgrade to python_cmr 0.10.0 because Ian had pushed his changes to the `temporal` methods upstream, which then allowed him to remove the logic from the `temporal` methods in `earthaccess` (simply calling the superclass). As discussed elsewhere, we did not remove the `temporal` methods completely from `earthaccess`, as we still don't have a solution for ensuring a complete set of docs in such a case. (I believe the solution might be that we need to get `python_cmr` to publish docs, which should create an index that we could then pull into our own without the need to duplicate docstrings.)

stubs/cmr/queries.pyi
```diff
@@ -48,8 +48,8 @@ class GranuleCollectionBaseQuery(Query):
     def online_only(self, online_only: bool = True) -> Self: ...
     def temporal(
         self,
-        date_from: Optional[Union[str, datetime]],
-        date_to: Optional[Union[str, datetime]],
+        date_from: Optional[Union[str, date, datetime]],
+        date_to: Optional[Union[str, date, datetime]],
```
mfisher87, line 52: Do we get to delete our stubs when python-cmr v0.11.0 is released?

mfisher87 (approved):
LGTM, great work @itcarroll and @chuckwondo. Awesome to delete code :grin:

stubs/cmr/queries.pyi
chuckwondo: Yes, which reminds me, I forgot to ask when that will happen.

Issue #511, opened 2024-04-08
Add `logging` module usage in place of `print` calls


chuckwondo:
There are a number of calls to `print` throughout the code base that should probably be replaced with appropriate calls to the standard `logging` library.

botanical:
Hello 👋 I opened a PR here https://github.com/nsidc/earthaccess/pull/566/. I'm unable to assign this to myself but I did check the open PRs and it doesn't look like anyone has addressed this issue yet.

mfisher87:
Resolved by #566 , thank you @botanical !

Issue #348, opened 2023-11-10
Squash merge? 


jrbourbeau:
I don't feel super strongly, but have a mild preference for using squash merging on most projects. This makes it so contributors don't have to be careful with individual commits (e.g. `git commit -m "typo"`, `git commit -m "Ah, forgot to lint"`) in their PRs and the `git log` on `main` is relatively clean (one commit per PR). 

Thoughts from others on moving to sqash merging?

MattF-NSIDC:
I prefer to use squash merging on a case-by-case basis. In PRs with well-curated commits, I like to keep them as-is. When we get PRs without well-curated commits, I'm all for a squash merge!

MattF-NSIDC:
As I'm thinking about this I keep changing my mind. I don't think there's any "correct" decision here... 

I think "is a clean history important?" is the root question, and I think "yes", but not above all, especially not above community health or inclusivity. I like a clean history, and I like to curate my commits to the best of my ability (after a particularly bad day, that ability can leave my body :laughing:), but I don't think we should leverage that expectation on contributors.

I think the best solution may just to be loosey-goosey about it, make it a "best effort" kind of thing and it'll work out fine. Encourage merge commits for well-curated PRs and squashes for messier ones, but be OK with people selecting the default and sometimes having messy PRs not get squashed. You can still get a clean one-commit-per-PR log view of `main` with merge commits.

The strongest opinion I have on this, I think, is that we should disable rebase merging, because even with well-curated commits, I'd rather only see the merge commit on `main` than all of them. Rebasing is also not nearly as broadly understood as merging, and I remember being intimidated by rebasing myself at one time :)

jhkennedy:
>  I don't think there's any "correct" decision here...

:100: 

> The strongest opinion I have on this, I think, is that we should disable rebase merging, because even with well-curated commits, I'd rather only see the merge commit on main than all of them. Rebasing is also not nearly as broadly understood as merging, and I remember being intimidated by rebasing myself at one time :)

I can agree to that and am happy to uncheck that one at least :smile:

mfisher87:
Yuvi makes an excellent point here: https://github.com/jupyterhub/team-compass/issues/686#issuecomment-1767629271

And thinking back on how I merged #566, I apologize @botanical for not considering how the choice might affect you. Personally, Yuvi's argument has swayed me and now I have stronger feelings :)

mfisher87:
I feel like a good outcome from this ticket may be either:

[...]

PR #566, opened 2024-05-11
Replace print statements with logger

botanical:
Addresses https://github.com/nsidc/earthaccess/issues/511

----
📚 Documentation preview 📚: https://earthaccess--566.org.readthedocs.build/en/566/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,14 @@
 # Changelog
 
 ## [Unreleased]
+* Bug fixes:
+    * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
+    * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
+    * Updated to use logging library in place of print statements throught the repo.
 
-* Changes
-
-  * [#421](https://github.com/nsidc/earthaccess/issues/421): Removed the
-    `get_user_profile` method and the `email_address` and `profile` attributes
-    from the `Auth` class.  Calling the EDL API to get user profile information
-    is not intended for library access and is not necessary for this library's
-    intended use cases.
-  * [#555](https://github.com/nsidc/earthaccess/issues/555): YAML formatting is
-    now performed with `yamlfmt` instead of `prettier`.
-
-* Enhancements
-
-  * [#483](https://github.com/nsidc/earthaccess/issues/483): Now using
-    [Search After](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after)
-    for collection and granule searches to support deep-paging through large
-    result sets.
-  * [#508](https://github.com/nsidc/earthaccess/issues/508): Corrected and
-    enhanced static type hints for functions and methods that make CMR queries
-    or handle CMR query results.
-  * [#421](https://github.com/nsidc/earthaccess/issues/421): Enabled queries to
-    Earthdata User Acceptance Testing (UAT) system for authenticated accounts.
-  * [#562](https://github.com/nsidc/earthaccess/issues/562): The destination
-     path is now created prior to direct S3 downloads, if it doesn't already
-     exist.
+* Enhancements:
+  * Corrected and enhanced static type hints for functions and methods that make
+    CMR queries or handle CMR query results (#508)
 
 ## [v0.9.0] 2024-02-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Bug fixes:
     * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
     * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
-    * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print` calls with `logging` calls, where appropriate.
+    * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print` calls with `logging` calls, where appropriate and added T20 Ruff rule.
 
 * Enhancements:
   * Corrected and enhanced static type hints for functions and methods that make
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
   * [#555](https://github.com/nsidc/earthaccess/issues/555): YAML formatting is
     now performed with `yamlfmt` instead of `prettier`.
   * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print`
-    calls with `logging` calls, where appropriate and added T20 Ruff rule.
+    calls with `logging` calls where appropriate and added T20 Ruff rule.
 
 * Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Bug fixes:
     * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
     * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
-    * Updated to use logging library in place of print statements throught the repo.
+    * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print` calls with `logging` calls, where appropriate.
 
 * Enhancements:
   * Corrected and enhanced static type hints for functions and methods that make
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,33 @@
 # Changelog
 
 ## [Unreleased]
-* Bug fixes:
-    * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
-    * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
-    * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print` calls with `logging` calls, where appropriate and added T20 Ruff rule.
 
-* Enhancements:
-  * Corrected and enhanced static type hints for functions and methods that make
-    CMR queries or handle CMR query results (#508)
+* Changes
+
+  * [#421](https://github.com/nsidc/earthaccess/issues/421): Removed the
+    `get_user_profile` method and the `email_address` and `profile` attributes
+    from the `Auth` class.  Calling the EDL API to get user profile information
+    is not intended for library access and is not necessary for this library's
+    intended use cases.
+  * [#555](https://github.com/nsidc/earthaccess/issues/555): YAML formatting is
+    now performed with `yamlfmt` instead of `prettier`.
+  * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print`
+    calls with `logging` calls, where appropriate and added T20 Ruff rule.
+
+* Enhancements
+
+  * [#483](https://github.com/nsidc/earthaccess/issues/483): Now using
+    [Search After](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after)
+    for collection and granule searches to support deep-paging through large
+    result sets.
+  * [#508](https://github.com/nsidc/earthaccess/issues/508): Corrected and
+    enhanced static type hints for functions and methods that make CMR queries
+    or handle CMR query results.
+  * [#421](https://github.com/nsidc/earthaccess/issues/421): Enabled queries to
+    Earthdata User Acceptance Testing (UAT) system for authenticated accounts.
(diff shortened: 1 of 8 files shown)


botanical:
I didn't update the notebooks which also contain `print()` statements. Should I update those as well? 🤔

mfisher87:
Good question! I think the notebooks should probably stay the same.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Bug fixes:
     * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
     * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
+    * Updated to use logging library in place of print statements throught the repo.
```
mfisher87, line 9:
```suggestion
    * Updated to use logging library in place of print statements throughout the repo.
```

earthaccess/api.py
```diff
@@ -64,7 +68,7 @@ def search_datasets(count: int = -1, **kwargs: Any) -> List[DataCollection]:
         ```
     """
     if not validate.valid_dataset_parameters(**kwargs):
-        print(
+        logger.warn(
             "Warning: a valid set of parameters is needed to search for datasets on CMR"
```
mfisher87, line 72:
```suggestion
            "A valid set of parameters is needed to search for datasets on CMR"
```

earthaccess/store.py
```diff
@@ -110,7 +113,7 @@ def __init__(self, auth: Any, pre_authorize: bool = False) -> None:
                     self.set_requests_session(url)
 
         else:
-            print("Warning: the current session is not authenticated with NASA")
+            logger.warn("Warning: the current session is not authenticated with NASA")
```
mfisher87, line 116:
```suggestion
            logger.warn("The current session is not authenticated with NASA")
```

mfisher87 (commented):
Looks wonderful, thank you for taking this on! :star_struck: 

A couple nitpicks :) What do you think?

mfisher87:
Oo, also, what do you think of enabling ruff ruleset [T20](https://docs.astral.sh/ruff/rules/#flake8-print-t20) to prevent us from re-introducing print statements?

chuckwondo (changes requested):
Hi @botanical! Glad to see your PR! Just a couple of tweaks, please, but otherwise, great start to your getting involved here!

chuckwondo:
> Oo, also, what do you think of enabling ruff ruleset [T20](https://docs.astral.sh/ruff/rules/#flake8-print-t20) to prevent us from re-introducing print statements?

Good call. We would just want to disable this check for the lines in `print_help`, where we actually do want `print`, although I'd like to simply remove that method. It seems superfluous, and is not called internally, but perhaps that's a separate issue.

earthaccess/search.py
```diff
@@ -306,8 +309,8 @@ def parameters(self, **kwargs: Any) -> Self:
 
     def print_help(self, method: str = "fields") -> None:
         """Prints the help information for a given method."""
-        print("Class components: \n")
-        print([method for method in dir(self) if method.startswith("_") is False])
+        logger.info("Class components: \n")
+        logger.info([method for method in dir(self) if method.startswith("_") is False])
```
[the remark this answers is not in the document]
mfisher87: Good catch!

chuckwondo:
> Oo, also, what do you think of enabling ruff ruleset [T20](https://docs.astral.sh/ruff/rules/#flake8-print-t20) to prevent us from re-introducing print statements?

@botanical, if you can add this to the PR, that would be great. Note that at a minimum you'll have to pull in the latest changes from `main` to resolve conflicts in `CHANGELOG.md`. I recommend pulling the latest from upstream main and rebasing your branch on that, but that's not a requirement.

earthaccess/search.py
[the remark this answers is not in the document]
botanical: https://github.com/nsidc/earthaccess/pull/566/commits/3fadd28f4fa0ad1facd4484985fe0a76e8854e36 fixed!

earthaccess/api.py
```diff
@@ -199,8 +203,8 @@ def download(
     try:
         results = earthaccess.__store__.get(granules, local_path, provider, threads)
     except AttributeError as err:
-        print(err)
-        print("You must call earthaccess.login() before you can download data")
+        logger.error(err)
+        logger.info("You must call earthaccess.login() before you can download data")
```
chuckwondo, line 207:
```suggestion
        logger.error(f"{err}: You must call earthaccess.login() before you can download data")
```

earthaccess/auth.py
```diff
@@ -222,10 +222,10 @@ def get_s3_credentials(
                         cumulus_resp.url, allow_redirects=True, timeout=15
                     )
                     if not (auth_resp.ok):
-                        print(
+                        logger.error(
                             f"Authentication with Earthdata Login failed with:\n{auth_resp.text[0:1000]}"
                         )
-                        print(
+                        logger.info(
```
chuckwondo, line 228:
```suggestion
                        logger.error(
```

earthaccess/store.py
```diff
@@ -612,11 +615,11 @@ def _download_file(self, url: str, directory: Path) -> str:
                         # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
                         shutil.copyfileobj(r.raw, f, length=1024 * 1024)
             except Exception:
-                print(f"Error while downloading the file {local_filename}")
-                print(traceback.format_exc())
+                logger.error(f"Error while downloading the file {local_filename}")
+                logger.error(traceback.format_exc())
```
chuckwondo, line 619: Using `logger.exception` will include the traceback, so there's no need for an additional call.

```suggestion
                logger.exception(f"Error while downloading the file {local_filename}")
```

earthaccess/store.py
```diff
@@ -660,7 +663,7 @@ def _open_urls_https(
             try:
                 fileset = _open_files(url_mapping, https_fs, threads)
             except Exception:
-                print(
+                logger.error(
                     "An exception occurred while trying to access remote files via HTTPS: "
                     f"Exception: {traceback.format_exc()}"
                 )
```
chuckwondo, line 669:
```suggestion
                logger.exception(
                    "An exception occurred while trying to access remote files via HTTPS"
                )
```

tests/integration/test_auth.py
```diff
@@ -62,7 +62,7 @@ def test_auth_throws_exception_if_netrc_is_not_present():
     with pytest.raises(Exception) as e_info:
         earthaccess.login(strategy="netrc")
         assertions.assertRaises(FileNotFoundError)
-        print(e_info)
+        logger.info(e_info)
```
chuckwondo, line 65: This is a test function, so we don't need to log a message.

```suggestion
```

tests/integration/test_auth.py
```diff
@@ -87,12 +87,12 @@ def test_auth_can_fetch_s3_credentials():
...
-                print(f"Testing S3 credentials for {daac['short-name']}")
+                logger.info(f"Testing S3 credentials for {daac['short-name']}")
                 credentials = earthaccess.get_s3_credentials(daac["short-name"])
                 assertions.assertIsInstance(credentials, dict)
                 assertions.assertTrue("accessKeyId" in credentials)
             except Exception as e:
-                print(
+                logger.error(
                     f"An error occured while trying to fetch S3 credentials for {daac['short-name']}: {e}"
                 )
```
chuckwondo, line 97: We shouldn't be logging messages from test functions. In fact, I suspect the `try`/`except` should be removed. If an exception occurs during the test, the test should fail. However, I don't know if this is an exception to that rule (no pun intended).

tests/unit/test_auth.py
```diff
@@ -107,7 +110,7 @@ def test_auth_fails_for_wrong_credentials(self, user_input, user_password):
         auth = Auth()
         auth.login(strategy="interactive")
         with pytest.raises(Exception) as e_info:
-            print(e_info)
+            logger.info(e_info)
```
chuckwondo, line 113:
```suggestion
```

chuckwondo (changes requested):
@botanical, sorry, I made a number of individual comments rather than unified review comments.

tests/integration/test_auth.py
botanical: https://github.com/botanical/earthaccess/commit/64ec1355f707f1c457415d723f4b43ddaf4ba456 it seems intentional to me according to this commit 😯

pyproject.toml
```diff
@@ -127,6 +127,9 @@ src = ["earthaccess", "stubs", "tests"]
 [tool.ruff.lint]
 extend-select = ["I", "T20"]
 
+[tool.ruff.lint.per-file-ignores]
+"earthaccess/search.py" = ["T20"]
```
botanical, line 131: I do not know if this is necessarily what we want but I didn't know how to ignore that we have intentional print statements 

> earthaccess/search.py:312:9: T201 `print` found
> earthaccess/search.py:313:9: T201 `print` found

tests/integration/test_auth.py
mfisher87: >In fact, I suspect the try/except should be removed.

>it seems intentional to me according to this commit 😯

It _does_ still seem weird. But let's be intentional about the scope of this PR :) Can we take this to an issue / discussion?

pyproject.toml
mfisher87: On those lines, you can append a comment like `# noqa: T201` :) https://docs.astral.sh/ruff/linter/#error-suppression

chuckwondo (changes requested):
@botanical, thanks for the cleanup of the `print` calls and adding the `ruff` rule. All looks good, except that I see numerous changes that have already landed on `main`, so it seems perhaps that your attempt to merge things from `main` went awry somewhere. For example, `.codespellignore` already exists on `main`, but shows up in this PR as an addition (among several other existing changes on `main`). Can you perhaps rebase your changes onto `main`? Happy to jump on a call to help sort it out if wrangling the commits gets gnarly.

tests/integration/test_auth.py
chuckwondo: > > In fact, I suspect the try/except should be removed.
> 
> > it seems intentional to me according to this commit 😯
> 
> It _does_ still seem weird. But let's be intentional about the scope of this PR :) Can we take this to an issue / discussion?

Sounds good to me. @botanical, would you mind opening a separate issue for this? Also, would you open yet another issue about refactoring loops in tests to instead be parametrized tests (`pytest.parametrize`, and yes, that is how it's spelled, which is apparently a valid variant of _parameterize_).

mfisher87: >yes, that is how it's spelled, which is apparently a valid variant of parameterize

Gets me every time

botanical: Issues added here:

- https://github.com/nsidc/earthaccess/issues/577
- https://github.com/nsidc/earthaccess/issues/578

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
     intended use cases.
   * [#555](https://github.com/nsidc/earthaccess/issues/555): YAML formatting is
     now performed with `yamlfmt` instead of `prettier`.
+  * [#511](https://github.com/nsidc/earthaccess/issues/511): Replaced `print`
+    calls with `logging` calls, where appropriate and added T20 Ruff rule.
```
mfisher87, line 15:
```suggestion
    calls with `logging` calls where appropriate and added T20 Ruff rule.
```

mfisher87 (approved):
Thanks so much @botanical :star_struck: 

@chuckwondo (or anyone else) if you don't have any last thoughts I'll merge tomorrow!

chuckwondo:
> Thanks so much @botanical 🤩
> 
> @chuckwondo (or anyone else) if you don't have any last thoughts I'll merge tomorrow!

@mfisher87, I'd like for us to use "Squash and merge" as a matter of habit. That's what I've been doing for PRs, as it keep the commit history much cleaner, unless there's an objection to that approach.

jhkennedy:
> That's what I've been doing for PRs, as it keep the commit history much cleaner, unless there's an objection to that approach.

This could be a whole :fire:-war discussion :rofl: and it's come up on slack before, but for Earthaccess I think squash-merging matches the current development/release workflow the best.

I think we should disable all but 1 way to merge personally, so there's no question:
image

mfisher87:
I think we can call this mergeable, and squash it is! Thanks again @botanical :bow: 

I don't have strong feelings about merge strategy, but I do feel strongly let's take it over here https://github.com/nsidc/earthaccess/issues/348 :)

PR #568, opened 2024-05-12
Use yamlfmt instead of prettier to format YAML

### LINKED ISSUE #555: prettier's pre-commit mirror is archived and keeps trying to upgrade us to alpha version
---

_Originally posted by @chuckwondo in https://github.com/nsidc/earthaccess/issues/446#issuecomment-2084844024_

> What do we think is the way forward?
> 
> * Turn off auto-updates
> * Edit the auto-update PRs before merging to remove alpha bumps
> * Use alpha prettier
> * ???

In this case, I'd say we remove https://github.com/pre-commit/mirrors-prettier because it has been archived.

There aren't many alternatives for yaml formatting. A couple of others are also archived, such as this one: https://github.com/mmlb/yamlfmt

Perhaps we go with this one: https://github.com/google/yamlfmt

chuckwondo:
Fixes #555

----
📚 Documentation preview 📚: https://earthaccess--568.org.readthedocs.build/en/568/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,31 @@
 # Changelog
 
 ## [Unreleased]
+
 * Changes
-  * Removed the `get_user_profile` method and the `email_address` and `profile` attributes from the `Auth` class ([#421](https://github.com/nsidc/earthaccess/issues/421))
-* Bug fixes:
-    * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
-    * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
+
+  * [#421](https://github.com/nsidc/earthaccess/issues/421): Removed the
+    `get_user_profile` method and the `email_address` and `profile` attributes
+    from the `Auth` class.  Calling the EDL API to get user profile information
+    is not intended for library access and is not necessary for this library's
+    intended use cases.
+  * [#555](https://github.com/nsidc/earthaccess/issues/555): YAML formatting is
+    now performed with `yamlfmt` instead of `prettier`.
+
+* Enhancements
+
+  * [#483](https://github.com/nsidc/earthaccess/issues/483): Now using
+    [Search After](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after)
+    for collection and granule searches to support deep-paging through large
+    result sets.
+  * [#508](https://github.com/nsidc/earthaccess/issues/508): Corrected and
+    enhanced static type hints for functions and methods that make CMR queries
+    or handle CMR query results.
+  * [#421](https://github.com/nsidc/earthaccess/issues/421): Enabled queries to
+    Earthdata User Acceptance Testing (UAT) system for authenticated accounts.
   * [#562](https://github.com/nsidc/earthaccess/issues/562): The destination
      path is now created prior to direct S3 downloads, if it doesn't already
      exist.
-
-* Enhancements:
-  * Corrected and enhanced static type hints for functions and methods that make
-    CMR queries or handle CMR query results ([#508](https://github.com/nsidc/earthaccess/issues/508))
-  * Enable queries to Earthdata User Acceptance Testing (UAT) system for authenticated accounts ([#421](https://github.com/nsidc/earthaccess/issues/421))
 
 ## [v0.9.0] 2024-02-28
 
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -76,11 +76,10 @@ nav:
       - "Download data from on-prem location": "howto/onprem.md"
       - "Direct S3 access - Open/stream files in the cloud": "howto/cloud.md"
   - TUTORIALS:
-      - "Accesing remote NASA data with fsspec": "tutorials/file-access.ipynb"
+      - "Accessing remote NASA data with fsspec": "tutorials/file-access.ipynb"
       - "Search and access of restricted datasets": "tutorials/restricted-datasets.ipynb"
       - "Reproducing NASA sea level rise infographic": "tutorials/SSL.ipynb"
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,28 @@
...
+
+  * [#483](https://github.com/nsidc/earthaccess/issues/483): Now using
+    [Search After](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after)
+    for collection and granule searches to support deep-paging through large
+    result sets.
+  * [#508](https://github.com/nsidc/earthaccess/issues/508): Corrected and
+    enhanced static type hints for functions and methods that make CMR queries
+    or handle CMR query results.
+  * [#421](https://github.com/nsidc/earthaccess/issues/421): Enabled queries to
+    Earthdata User Acceptance Testing (UAT) system for authenticated accounts.
```
mfisher87, line 25: :star_struck: thank you :sparkles: 

I wonder if there's a changelog linting tool that can help us enforce a standard like the one you're setting here.

mkdocs.yml
```diff
@@ -80,7 +80,7 @@ nav:
       - "Search and access of restricted datasets": "tutorials/restricted-datasets.ipynb"
       - "Reproducing NASA sea level rise infographic": "tutorials/SSL.ipynb"
       - "Accessing and visualizing EMIT data with a few lines of code": "tutorials/emit-earthaccess.ipynb"
-      # - 'Distributing workloads with Dask clusters': 'tutorials/dask.ipynb'
+        # - 'Distributing workloads with Dask clusters': 'tutorials/dask.ipynb'
```
mfisher87, line 83: This is so nitpicky, but I don't like this. When I comment a YAML line, the indentation of the comment is sometimes significant, as it was in this case. Can we turn this off?

mfisher87 (approved):

mfisher87:
Integration test failures look ignorable. They're happening everywhere, and this PR doesn't touch anything that could affect runtime.

mkdocs.yml
chuckwondo: Nitpicky is fine by me :-)

Let me see if we can configure this.

However, if not, I suggest we actually remove the line. In this case, it's not a comment, per se, but rather something "commented out". To me, this is akin to committing commented out code, which I don't like. My take is that if you want to comment out something, just delete it. If you want to know what it was, look at the commit history, right?

mfisher87: Thanks for looking in to that!

Agree totally on commented-out code. But I also don't know who added that comment and what they intended (sorry if it feels like we're dogpiling on whoever wrote it: IF we're going to comment code, I like a reason to be also included, like `# we're waiting on issue #xyz to be resolved to re-add this tutorial:`), so I chose to set that aspect aside :)

chuckwondo: Doesn't look like I can control the indentation of that commented-out line. However, the indentation is changed _only_ when it is the last item in the list. If I move the item to any other position within that list, it is left as-is.

I might open an issue against yamlfmt about this behavior.

I the meantime, which would you prefer: (a) remove the line in question, or (b) move the line to a different position in the list (first, next-to-last, other)?

My preference would be to remove it, but I'm not vehemently opposed to simply changing its position.

chuckwondo:
> Integration test failures look ignorable. They're happening everywhere, and this PR doesn't touch anything that could affect runtime.

Weird.

mkdocs.yml
mfisher87: >I might open an issue against yamlfmt about this behavior.

Seems like the right next step, can't imagine why that behavior would be intended!

>I the meantime, which would you prefer: (a) remove the line in question, or (b) move the line to a different position in the list (first, next-to-last, other)?
>
>My preference would be to remove it, but I'm not vehemently opposed to simply changing its position.

Up to you, but I think if we remove the comment we should create a ticket to "figure out what to do with the dask tutorial"; re-add it to the nav (what needs to be done before we can do that?), or remove the tutorial from the repo?

mfisher87: Actually, we should have a ticket for that either way. The comment is not doing the same job as a ticket :laughing: I'll open one.

mfisher87: https://github.com/nsidc/earthaccess/issues/569

My vote is now for "delete", we'll figure it out through the issue :)

chuckwondo: Ok, I'm going to remove the comment. By the way, this indentation behavior is already known: https://github.com/google/yamlfmt/issues/112

I see that @betolink added the commented out line in this commit: https://github.com/nsidc/earthaccess/commit/dab9da4b290424dc4de2bd2212a8d1212019b947#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R66

It looks like it never existed _uncommented_ and it was originally added as a commented-out line, perhaps as a reminder. Since this comment doesn't seem to have done its job as a reminder (as you mentioned), your opening a new ticket sounds like a better reminder :-)

mfisher87: :laughing: We did the exact same sleuthing in different threads: https://github.com/nsidc/earthaccess/issues/569#issuecomment-2106313821

chuckwondo:
> Integration test failures look ignorable. They're happening everywhere, and this PR doesn't touch anything that could affect runtime.

Looks like this was a transient problem. All are passing now.

mfisher87:
There's something bigger going on here. For the last 22 minutes, I was unable to connect to any Google services from my home ISP, but my mobile ISP was fine. Just came back now.

chuckwondo:
@mfisher87, this is ready for your eyes again. I pulled in the additional changelog entry from the other PR that I just merged into main, so the indentation issue you identified is fixed.

mfisher87 (approved):
LGTM :rocket:

PR #570, opened 2024-05-13
Create dest dir for direct S3 access before download

### LINKED ISSUE #562: download while on AWS creates one file
The `earthaccess.download` function appears to be (over)writing all data received to a single file, rather than to separate files in the provided directory, when working on AWS. This example uses earthaccess 0.9.0 on cryointhecloud.

```python
import earthaccess

earthaccess.login()
results = earthaccess.search_data(short_name="PACE_OCI_L2_AOP_NRT", count=2)
earthaccess.download(results, "test")
!ls -l test
```
```
Granules found: 4148
 Getting 2 granules, approx download size: 0.07 GB
Accessing cloud dataset using dataset endpoint credentials: https://obdaac-tea.earthdatacloud.nasa.gov/s3credentials
Downloaded: test/PACE_OCI.20240409T203648.L2.OC_AOP.V1_0_0.NRT.nc
Downloaded: test/PACE_OCI.20240409T204148.L2.OC_AOP.V1_0_0.NRT.nc
-rw-r--r-- 1 jovyan jovyan 9516948 May  8 21:04 test
```
Instead of a "test" folder, I get one "test" file.

Expected behavior would be either to `raise "STAHP! Wut R u DoINg?"` or just download the files.

chuckwondo:
I didn't add a unit or integration test for this, as I wasn't sure about how best to test this, so I'm open to suggestion.

Fixes #562

----
📚 Documentation preview 📚: https://earthaccess--570.org.readthedocs.build/en/570/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * Bug fixes:
     * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
     * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
+  * [#562](https://github.com/nsidc/earthaccess/issues/562): The destination
+     path is now created prior to direct S3 downloads, if it doesn't already
+     exist.
 
 * Enhancements:
   * Corrected and enhanced static type hints for functions and methods that make
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -571,6 +571,9 @@ class Store(object):
             else:
                 print(f"Accessing cloud dataset using provider: {provider}")
                 s3_fs = self.get_s3fs_session(provider=provider)
+
+            local_path.mkdir(parents=True, exist_ok=True)
+
             # TODO: make this async
             for file in data_links:
                 s3_fs.get(file, str(local_path))


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 * Bug fixes:
     * fixed 483 by extracting a common CMR query method for collections and granules using SearchAfter header
     * Added VCR support for verifying the API call to CMR and the parsing of returned results without relying on CMR availability post development
+  * [#562](https://github.com/nsidc/earthaccess/issues/562): The destination
```
mfisher87, line 9: Indentation is off by 2 spaces with the items above

mfisher87 (approved):
Looks good so far! For testing, I'm thinking we could extract a `download_s3_granules(path: Path, data_links: list[str]) -> list[Path]` function and give it the input required to reproduce the bug in a test?

Really we should be testing that any time we write granules we will correctly create an output directory, but if nobody has time to do that refactoring right now, the above approach would at least add coverage to this change :shrug: No strong feelings.

CHANGELOG.md
chuckwondo: Oops! I'll sort that out with my other PR, where I've reformatted the section. I didn't want to create much of a merge conflict for myself here.

chuckwondo:
> Looks good so far! For testing, I'm thinking we could extract a `download_s3_granules(path: Path, data_links: list[str]) -> list[Path]` function and give it the input required to reproduce the bug in a test?
> 
> Really we should be testing that any time we write granules we will correctly create an output directory, but if nobody has time to do that refactoring right now, the above approach would at least add coverage to this change 🤷 No strong feelings.

I think the problem with testing this is that we cannot guarantee that tests will run in an AWS environment that's "in region." If that's not possible, then manually setting `earthaccess.__store__.in_region = True` doesn't help us because if the test is not actually running "in region", attempting a direct S3 download simply results in `PermissionError: Access Denied`.

mfisher87:
Hm. I don't have any great ideas here. @betolink ?

I'd be fine with merging without adding a test and recording a bug to add a test later. Better to get the fix in the hands of users sooner IMO. We can flail around figuring out unit tests on our time :rofl:

PR #571, opened 2024-05-13
Add Python 3.12 to automated test matrix

### LINKED ISSUE #457: Test against Python 3.12
Currently we test 3.8 to 3.11, but not 3.12.

chuckwondo:
Fixes #457

----
📚 Documentation preview 📚: https://earthaccess--571.org.readthedocs.build/en/571/

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -55,7 +55,12 @@ jobs:
         if: steps.cache.outputs.cache-hit == 'true'
         run: poetry run pip --version >/dev/null 2>&1 || rm -rf .venv
       - name: Install Dependencies
-        # Retry the installation command in case of transient network issues
+        if: ${{ !env.ACT }}
+        run: poetry install
+      - name: Install Dependencies
+        if: ${{ env.ACT }}
+        # When using `act` to run the workflow locally, the `poetry install` command
+        # may fail due to network issues when running multiple Docker containers.
         run: poetry install || poetry install || poetry install
       - name: Test
         env:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
       fail-fast: false
 
     steps:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -44,6 +44,7 @@ jobs:
         run: |
           poetry config virtualenvs.create true --local
           poetry config virtualenvs.in-project true --local
+          poetry self add setuptools
       - name: Set up cache
         uses: actions/cache@v4
         id: cache
@@ -54,7 +55,8 @@ jobs:
         if: steps.cache.outputs.cache-hit == 'true'
         run: poetry run pip --version >/dev/null 2>&1 || rm -rf .venv
       - name: Install Dependencies
-        run: poetry install
+        # Retry the installation command in case of transient network issues
+        run: poetry install || poetry install || poetry install
       - name: Test
         env:
           EARTHDATA_USERNAME: ${{ secrets.EDL_USERNAME }}
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -44,8 +44,6 @@ jobs:
         run: |
           poetry config virtualenvs.create true --local
           poetry config virtualenvs.in-project true --local
-          # Avoid race condition that can lead to ChefBuildError
-          poetry config installer.max-workers 1 --local
           poetry self add setuptools
       - name: Set up cache
         uses: actions/cache@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/integration-test.yml
```diff
@@ -35,17 +39,17 @@ jobs:
         id: full-python-version
         run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
       - name: Install poetry
-        run: |
-          curl -sSL https://install.python-poetry.org | python3 -
-          echo "$HOME/.poetry/bin" >> $GITHUB_PATH
+        uses: abatilo/actions-poetry@v3
```
mfisher87, line 42: :100:

.github/workflows/integration-test.yml
```diff
@@ -59,4 +63,6 @@ jobs:
           EARTHACCESS_TEST_PASSWORD: ${{ secrets.EDL_PASSWORD }}
         run: poetry run bash scripts/integration-test.sh
       - name: Upload coverage
+        # Don't upload coverage when using the `act` tool to run the workflow locally
+        if: ${{ !env.ACT }}
```
mfisher87, line 67: TIL!

pyproject.toml
```diff
@@ -8,7 +8,8 @@ maintainers = [
     "Luis Lopez <betolin@gmail.com>",
     "Joseph H. Kennedy <jhkennedy@alaska.edu>",
     "James Bourbeau <james@coiled.io>",
-    "Matt Fisher <mfisher87@gmail.com>"
+    "Matt Fisher <mfisher87@gmail.com>",
+    "Chuck Daniels <chuck@developmentseed.org>"
```
mfisher87, line 12: Good catch :) Forgot there's more to do than just give you permissions!

mfisher87 (dismissed):
Looks great as usual ;)

mfisher87:
Not sure about the build failure.

mfisher87 (commented):
Too bad we needed to set up those messy pins :(

I'm especially curious about pyproj, as the included proj database can be materially different in different versions. Would we be able to avoid these "dynamic" pins if we drop support for Python 3.8?

chuckwondo:
> Too bad we needed to set up those messy pins :(
> 
> I'm especially curious about pyproj, as the included proj database can be materially different in different versions. Would we be able to avoid these "dynamic" pins if we drop support for Python 3.8?

Yes, I believe dropping support for Python 3.8 would do the trick, but do we want to drop support prior to its EOL, which is Oct 2024 (see https://devguide.python.org/versions/)?

mfisher87:
Since members of our ecosystem (Numpy + Pyproj) are doing it already, I think it makes sense for us to follow suit. Numpy dropped support for 3.8 [over a year ago](https://numpy.org/neps/nep-0029-deprecation_policy.html). Here's their policy:

>All minor versions of Python released 42 months prior, and at minimum the two latest minor versions are supported. Python support is only dropped in a major/minor version, and never on a patch release.

chuckwondo:
Should I set min versions of numpy and pyproj to be the min versions that support Python 3.12? That is, do we want users using Python 3.9-3.11 to also get the min versions of numpy and pyproj that only support >=3.12, or is that too restrictive for dependency resolution?

More specifically, do we want these (regardless of Python version, user gets only what works for Python >=3.12):

```toml
numpy = { version = ">=1.26.0", optional = true }
pyproj = ">=3.6.1"
```

or these (looser for Python <3.12)?

```toml
numpy = { version = ">=1.26.0", optional = true, python = ">=3.12" }
pyproj = [
  { version = ">=3.5.0", python = "<3.12" },
  { version = ">=3.6.1", python = ">=3.12" }
]
```

cc: @betolink, @jhkennedy

chuckwondo:
I've pinned numpy (optional) and pyproj to their versions that are compatible with Python 3.12. This means that those using Python <3.12 _also_ get those 3.12-compatible versions. Do we want to loosen the pin to allow for earlier versions of numpy and pyproj for those using Python <3.12?

mfisher87:
> Do we want to loosen the pin to allow for earlier versions of numpy and pyproj for those using Python <3.12?

I think so. I think we should make our pins as loose as we can given our dependency policy (for now, safe to assume SPEC0, I think?) to maximize compatibility with users' environments.

We should probably include a pinning policy in our dependency policy :)

chuckwondo:
> > Do we want to loosen the pin to allow for earlier versions of numpy and pyproj for those using Python <3.12?
> 
> I think so. I think we should make our pins as loose as we can given our dependency policy (for now, safe to assume SPEC0, I think?) to maximize compatibility with users' environments.

Thanks for confirming. I suspected as much. I'll adjust accordingly.

chuckwondo:
@mfisher87, I loosened the numpy and pyproj versions for Python <3.12. I believe this should be good to go now.

I set pyproj to its previous pin, but added a new min version for Python >=3.12. For numpy, there was no existing pin, so I added a min version for Python >=3.12 (the version that numpy added support for 3.12), and a min version 2 versions earlier for Python <3.12.

chuckwondo:
@mfisher87, as you requested, so this appears at the top of your queue :-)

mfisher87 (approved):
LGTM, sorry for the delay. I appreciated the repeated reminder!

PR #580, opened 2024-05-21
Added linting of docstrings

### LINKED ISSUE #454: add linting of docstrings, using ruff
This came out of this separate docstring PR: https://github.com/nsidc/earthaccess/pull/448#issuecomment-1934742889#448 

Purpose: Enable linting of docstrings via ruff [this way](https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings).

@mfisher87

Sherwin-14:
I have added the linting for docstrings. Please have a look at it ;) Resolves #454 

----
📚 Documentation preview 📚: https://earthaccess--580.org.readthedocs.build/en/580/

### DIFF
--- a/earthaccess/results.py
+++ b/earthaccess/results.py
@@ -94,9 +94,10 @@ class DataCollection(CustomDict):
         return summary_dict
 
     def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
-        """
+        """Placeholder.
+
         Parameters:
-            umm_field: Valid UMM item, i.e. `TemporalExtent`
+            umm_field: Valid UMM item, i.e. `TemporalExtent`.
 
         Returns:
             The value of a given field inside the UMM (Unified Metadata Model).
@@ -106,17 +107,16 @@ class DataCollection(CustomDict):
         return ""
 
     def concept_id(self) -> str:
-        """
-        Returns:
-            A collection's `concept_id`.
-                This id is the most relevant search field on granule queries.
+        """Placeholder.
+
+        Returns: A collection's `concept_id`.This id is the most relevant search field on granule queries.
         """
         return self["meta"]["concept-id"]
 
     def data_type(self) -> str:
-        """
-        Returns:
-            The collection data type, i.e. HDF5, CSV etc., if available.
+        """Placeholder.
+
+        Returns: The collection data type, i.e. HDF5, CSV etc., if available.
         """
         if "ArchiveAndDistributionInformation" in self["umm"]:
             return str(
@@ -127,27 +127,27 @@ class DataCollection(CustomDict):
         return ""
 
     def version(self) -> str:
-        """
-        Returns:
-            The collection's version.
+        """Placeholder.
+
+        Returns: The collection's version.
         """
         if "Version" in self["umm"]:
             return self["umm"]["Version"]
         return ""
 
     def abstract(self) -> str:
-        """
-        Returns:
-            The abstract of a collection
+        """Placeholder.
+
+        Returns:The abstract of a collection.
         """
         if "Abstract" in self["umm"]:
             return self["umm"]["Abstract"]
         return ""
 
     def landing_page(self) -> str:
-        """
-        Returns:
-            The first landing page for the collection (can be many), if available.
+        """Placeholder.
+
+        Returns: The first landing page for the collection (can be many), if available.
         """
         links = self._filter_related_links("LANDING PAGE")
         if len(links) > 0:
@@ -155,18 +155,17 @@ class DataCollection(CustomDict):
         return ""
 
     def get_data(self) -> List[str]:
-        """
-        Returns:
-            The GET DATA links (usually a landing page link, a DAAC portal, or an FTP location).
+        """Placeholder.
+
+        Returns: The GET DATA links (usually a landing page link, a DAAC portal, or an FTP location).
         """
         links = self._filter_related_links("GET DATA")
         return links
 
     def s3_bucket(self) -> Dict[str, Any]:
-        """
-        Returns:
-            The S3 bucket information if the collection has it.
-                (**cloud hosted collections only**)
+        """Placeholder.
+
+        Returns: The S3 bucket information if the collection has it.(**cloud hosted collections only**).
         """
         if "DirectDistributionInformation" in self["umm"]:
             return self["umm"]["DirectDistributionInformation"]
@@ -214,9 +213,9 @@ class DataGranule(CustomDict):
             self.render_dict = self._filter_fields_(fields)
 
     def __repr__(self) -> str:
-        """
-        Returns:
-            A basic representation of a data granule.
+        """Placeholder.
+
+        Returns: A basic representation of a data granule.
         """
         data_links = [link for link in self.data_links()]
         rep_str = f"""
@@ -229,9 +228,9 @@ class DataGranule(CustomDict):
         return rep_str
 
     def _repr_html_(self) -> str:
-        """
-        Returns:
-            A rich representation for a data granule if we are in a Jupyter notebook.
+        """Placeholder.
+
+        Returns: A rich representation for a data granule if we are in a Jupyter notebook.
         """
         granule_html_repr = _repr_granule_html(self)
         return granule_html_repr
@@ -243,9 +242,9 @@ class DataGranule(CustomDict):
         return None
 
     def size(self) -> float:
-        """
-        Returns:
-            The total size for the granule in MB.
+        """Placeholder.
+
+        Returns: The total size for the granule in MB.
         """
         try:
             data_granule = self["umm"]["DataGranule"]
@@ -284,7 +283,9 @@ class DataGranule(CustomDict):
     def data_links(
         self, access: Optional[str] = None, in_region: bool = False
     ) -> List[str]:
-        """Returns the data links from a granule.
+        """Placeholder.
+
+        Returns the data links from a granule.
 
         Parameters:
             access: direct or external.
@@ -325,9 +326,9 @@ class DataGranule(CustomDict):
                 return https_links
 
     def dataviz_links(self) -> List[str]:
-        """
-        Returns:
-            The data visualization links, usually the browse images.
+        """Placeholder.
+
+        Returns: The data visualization links, usually the browse images.
         """
         links = self._filter_related_links("GET RELATED VISUALIZATION")
         return links
--- a/earthaccess/results.py
+++ b/earthaccess/results.py
@@ -109,14 +109,16 @@ class DataCollection(CustomDict):
     def concept_id(self) -> str:
         """Placeholder.
 
-        Returns: A collection's `concept_id`.This id is the most relevant search field on granule queries.
+        Returns:
+            A collection's `concept_id`.This id is the most relevant search field on granule queries.
         """
         return self["meta"]["concept-id"]
 
     def data_type(self) -> str:
         """Placeholder.
 
-        Returns: The collection data type, i.e. HDF5, CSV etc., if available.
+        Returns:
+            The collection data type, i.e. HDF5, CSV etc., if available.
         """
         if "ArchiveAndDistributionInformation" in self["umm"]:
             return str(
@@ -129,7 +131,8 @@ class DataCollection(CustomDict):
     def version(self) -> str:
         """Placeholder.
 
-        Returns: The collection's version.
+        Returns:
+            The collection's version.
         """
         if "Version" in self["umm"]:
             return self["umm"]["Version"]
@@ -138,7 +141,8 @@ class DataCollection(CustomDict):
     def abstract(self) -> str:
         """Placeholder.
 
-        Returns:The abstract of a collection.
+        Returns:
+            The abstract of a collection.
         """
         if "Abstract" in self["umm"]:
             return self["umm"]["Abstract"]
@@ -147,7 +151,8 @@ class DataCollection(CustomDict):
     def landing_page(self) -> str:
         """Placeholder.
 
-        Returns: The first landing page for the collection (can be many), if available.
+        Returns:
+            The first landing page for the collection (can be many), if available.
         """
         links = self._filter_related_links("LANDING PAGE")
         if len(links) > 0:
@@ -157,7 +162,8 @@ class DataCollection(CustomDict):
     def get_data(self) -> List[str]:
         """Placeholder.
 
-        Returns: The GET DATA links (usually a landing page link, a DAAC portal, or an FTP location).
+        Returns:
+            The GET DATA links (usually a landing page link, a DAAC portal, or an FTP location).
         """
         links = self._filter_related_links("GET DATA")
         return links
@@ -165,7 +171,8 @@ class DataCollection(CustomDict):
     def s3_bucket(self) -> Dict[str, Any]:
         """Placeholder.
 
-        Returns: The S3 bucket information if the collection has it.(**cloud hosted collections only**).
+        Returns:
+            The S3 bucket information if the collection has it (**cloud hosted collections only**).
         """
         if "DirectDistributionInformation" in self["umm"]:
             return self["umm"]["DirectDistributionInformation"]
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -126,6 +126,8 @@ src = ["earthaccess", "stubs", "tests"]
 
 [tool.ruff.lint]
 extend-select = ["I", "T20"]
+select = ["D"]
```
mfisher87, line 129:
```suggestion
extend-select = ["I", "T20", "D"]
```

`select` will override the default enabled rules :)

mfisher87 (changes requested):
Wow, that produced a _lot_ of [new linting errors](https://results.pre-commit.ci/run/github/399867529/1716293781.NOpGQxffTSu0h4QjiKrhWw) :laughing: :sweat_smile: 

I think we can, for now, ignore the `D1*` errors because it's going to be a lot of work to add docstrings to every module, class, and function.

Ruff seems to have [autofixers for most of the `D2*` and `D4*` errors](https://docs.astral.sh/ruff/rules/#pydocstyle-d). I'd try running ruff with `ruff check --fix`, and if that doesn't fix enough things, try `ruff check --fix --unsafe-fixes`. If you need to use unsafe fixes, please keep an extra eye out for anything weird :)

mfisher87:
@Sherwin-14 I'd highly recommend using Ruff with [pre-commit](https://pre-commit.com/) instead of installing it standalone. Once you have installed pre-commit, you can use it to install hooks for this project with `pre-commit install`. Then, every time you run a commit, checks and even automatic fixes will occur, preventing you from pushing commits that might fail checks later! It's my favorite tool :) You can also trigger checks manually with `pre-commit run -a`.

We also use it on the [Benefit Tool project](https://github.com/nsidc/usaon-benefit-tool/blob/main/.pre-commit-config.yaml).

Sherwin-14:
@mfisher87 Hey Matt, Even with your suggestion of unsafe fixes we still have 258 unsolved errors.

mfisher87:
Sounds like it auto-fixed about 500 errors, nice! Can you commit and push those? Then I can see what kind of errors are left. Maybe we want to set those to ignored.

What about ignoring the `D1*` rules? How many errors did that eliminate?

Sherwin-14:
@ Hey , we found 159 errors barring D1*

mfisher87:
Surprised `D205` wasn't autofixed... weird. `D417` and any others you're still getting you can add to the ignore list along with `D1`. All these autofixes are already a huge improvement, and we can un-ignore and fix more errors as time goes on. If you are willing to manually fix any of these errors, you're also welcome to do that :)

mfisher87:
Woops, didn't mean to make this draft.

Sherwin-14:
I think I should push the changes, adding D417 errors to the ignore list as well. What do you think about this? Also fixing some of those issues is something I too considered ;)

pyproject.toml
```diff
@@ -125,7 +125,8 @@ line-length = 88
 src = ["earthaccess", "stubs", "tests"]
 
 [tool.ruff.lint]
-extend-select = ["I", "T20"]
+extend-select = ["I", "T20", "D"]
+ignore = ["D100", "D101", "D102", "D103", "D104", "D105", "D107", "D205", "D401", "D417", "I001"]
```
mfisher87, line 129:
```suggestion
ignore = ["D1", "D205", "D401", "D417"]
```

I think we can ignore all `D1*` errors like this :point_up: Can you check if this works?

We need to keep `I001` on and fix any unsorted import errors. Did the autofixer not take care of it? If this one needs to be manually fixed, we need to do it :)

mfisher87 (commented):
I think one last tweak and we're ready to go. Nice work!

mfisher87 (approved):
LGTM, thanks @Sherwin-14 !

Will plan to merge tomorrow just in case anyone wants to take a last look. CC participants of the discussion that originated this need: @danielfromearth @jhkennedy @andypbarrett

danielfromearth:
Yay, I'm glad to see these changes!

One thing that I don't mind one way or the other, but I'm curious... docstrings like this:

```python
def get_s3_credentials(
        self,
        daac: Optional[str] = None,
        endpoint: Optional[str] = None,
    ) -> Dict[str, str]:
        """Gets AWS S3 credentials for a given NASA cloud provider.
        The easier way is to use the DAAC short name; provider is optional if we know it.

        Parameters
        ----------
            daac: The name of a NASA DAAC, e.g. NSIDC or PODAAC.
            provider: A valid cloud provider. Each DAAC has a provider code for their cloud distributions.
            endpoint: Getting the credentials directly from the S3Credentials URL.

        Returns
        -------
            A Python dictionary with the temporary AWS S3 credentials.

        """
```

Looks to me like a combination of Google style — with indents, and colons without spaces — and numpy style, using `-------` lines below the section names.  Is this combination a `ruff` style?  Does anyone here know?

mfisher87:
Good catch. It actually looks like the docs aren't rendering correctly. Should have checked that. I think we need a tweak :)

https://earthaccess--580.org.readthedocs.build/en/580/user-reference/api/api/#earthaccess.api.get_s3_credentials

https://earthaccess.readthedocs.io/en/latest/user-reference/api/api/#earthaccess.api.get_s3_credentials

mfisher87:
@Sherwin-14 I forgot a critical piece!

```
[tool.ruff.lint.pydocstyle]
convention = "google"
```

To re-apply the rules, we need to run the `--fix` (maybe also unsafe) operation. We _may_ need/want to un-do the previous fix operation (keeping only changes to `pyproject.toml`) prior to re-running the fix.

mfisher87 (dismissed):
@Sherwin-14 I forgot a critical piece! If you check the links in the post above this one, the docs aren't rendering correctly anymore because I forgot we need to set the correct style.

```
[tool.ruff.lint.pydocstyle]
convention = "google"
```

To re-apply the rules, we need to run the `--fix` (maybe also unsafe) operation. We _may_ need/want to un-do the previous fix operation (keeping only changes to `pyproject.toml`) prior to re-running the fix.

mfisher87:
Docs are looking great now! :partying_face: 

https://earthaccess--580.org.readthedocs.build/en/580/user-reference/api/api/#earthaccess.api.get_s3_credentials

earthaccess/search.py
```diff
@@ -79,11 +77,10 @@ def get_results(
...
 class DataCollections(CollectionQuery):
-    """
-    ???+ Info
-        The DataCollection class queries against
-        https://cmr.earthdata.nasa.gov/search/collections.umm_json,
-        the response has to be in umm_json to use the result classes.
+    """???+ Info
+    The DataCollection class queries against
+    https://cmr.earthdata.nasa.gov/search/collections.umm_json,
+    the response has to be in umm_json to use the result classes.
```
mfisher87, line 83: Looks like the linter messed up the indentation on this info box , so now it renders as empty (here: https://earthaccess--580.org.readthedocs.build/en/580/user-reference/collections/collections-query/)! :(

I think this happened because there's no subject line in the docstring like there should be. I think we need to add that. A placeholder would be OK, since there's nothing there right now. Are you OK with adding `Placeholder.` as the first line for docstrings missing one?

earthaccess/results.py
```diff
@@ -325,9 +314,8 @@ def data_links(
                 return https_links
 
     def dataviz_links(self) -> List[str]:
-        """
-        Returns:
-            The data visualization links, usually the browse images.
+        """Returns:
+        The data visualization links, usually the browse images.
```
mfisher87, line 318: The docs for this function also didn't render correctly because of a missing subject line :(

https://earthaccess--580.org.readthedocs.build/en/580/user-reference/granules/granules/#earthaccess.results.DataGranule.dataviz_links

mfisher87 (commented):
Hey Sherwin, sorry for another wrench in the works! Looks like we're _so close_ to being there but Ruff seems to have caused some doc pages to render incorrectly because it was too aggressive with trying to populate a subject line. We'll need to do a bit of manual editing to fix it as far as I can tell. If you feel comfortable writing a docstring subject line, that would be so appreciated! But otherwise we can just leave it as `TODO.` or `Placeholder.` so the docs render correctly and come back to it later.

If those manual edits pan out to be too time consuming, feel free to ask for help :)

Sherwin-14:
Hey Matt, wouldn't it be better to configure ruff to ignore this!

mfisher87:
>Hey Matt, wouldn't it be better to configure ruff to ignore this!

What are you suggesting we configure ruff to ignore? If we can turn off a rule and fix that behavior I'm for it :)

Sherwin-14:
@mfisher87 Hey Matt, apparently numpy is an implicit dependency brought by one of your packages. This version however, requires distutils which has been removed from [Python 3.12](https://docs.python.org/3/whatsnew/3.12.html)

earthaccess/search.py
```diff
@@ -79,11 +77,10 @@ def get_results(
...
 class DataCollections(CollectionQuery):
-    """
-    ???+ Info
-        The DataCollection class queries against
-        https://cmr.earthdata.nasa.gov/search/collections.umm_json,
-        the response has to be in umm_json to use the result classes.
+    """???+ Info "Placeholder"
+    The DataCollection class queries against
+    https://cmr.earthdata.nasa.gov/search/collections.umm_json,
+    the response has to be in umm_json to use the result classes.
```
mfisher87, line 83:
```suggestion
    """Placeholder.
    
    ???+ Info
        The DataCollection class queries against
        https://cmr.earthdata.nasa.gov/search/collections.umm_json,
        the response has to be in umm_json to use the result classes.
```

mfisher87 (approved):
Nice! @chuckwondo @danielfromearth any last minute thoughts?

earthaccess/results.py
```diff
@@ -106,17 +107,16 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
         """
         return self["meta"]["concept-id"]
 
     def data_type(self) -> str:
-        """
-        Returns:
-            The collection data type, i.e. HDF5, CSV etc., if available.
+        """Placeholder.
+
+        Returns: The collection data type, i.e. HDF5, CSV etc., if available.
```
mfisher87, line 119: This change is actually breaking some documentation:

https://earthaccess.readthedocs.io/en/latest/user-reference/granules/granules/#earthaccess.results.DataGranule.dataviz_links

https://earthaccess--580.org.readthedocs.build/en/580/user-reference/granules/granules/#earthaccess.results.DataGranule.dataviz_links

The return description is no longer being parsed correctly when this change is applied.

mfisher87 (changes requested):
Ugh, I was too hasty. There's one problem left that I can see.

Sherwin-14:
@mfisher87 Hey sorry about that I have fixed the issue now.you can take a look at it ;)

earthaccess/results.py
```diff
@@ -127,25 +129,28 @@ def data_type(self) -> str:
...
             return self["umm"]["Version"]
         return ""
 
     def abstract(self) -> str:
-        """
+        """Placeholder.
+
         Returns:
-            The abstract of a collection
+           The abstract of a collection.
```
mfisher87, line 145:
```suggestion
            The abstract of a collection.
```

earthaccess/results.py
```diff
@@ -155,18 +160,19 @@ def landing_page(self) -> str:
...
         return links
 
     def s3_bucket(self) -> Dict[str, Any]:
-        """
+        """Placeholder.
+
         Returns:
-            The S3 bucket information if the collection has it.
-                (**cloud hosted collections only**)
+           The S3 bucket information if the collection has it.(**cloud hosted collections only**).
```
mfisher87, line 175:
```suggestion
           The S3 bucket information if the collection has it (**cloud hosted collections only**).
```

earthaccess/results.py
```diff
@@ -214,9 +220,9 @@ def __init__(
             self.render_dict = self._filter_fields_(fields)
 
     def __repr__(self) -> str:
-        """
-        Returns:
-            A basic representation of a data granule.
+        """Placeholder.
+
+        Returns: A basic representation of a data granule.
```
mfisher87, line 225: "Returns" should be on its own line here

earthaccess/results.py
```diff
@@ -229,9 +235,9 @@ def __repr__(self) -> str:
         return rep_str
 
     def _repr_html_(self) -> str:
-        """
-        Returns:
-            A rich representation for a data granule if we are in a Jupyter notebook.
+        """Placeholder.
+
+        Returns: A rich representation for a data granule if we are in a Jupyter notebook.
```
mfisher87, line 240: Here too

earthaccess/results.py
```diff
@@ -127,25 +129,28 @@ def data_type(self) -> str:
         return ""
 
     def version(self) -> str:
-        """
+        """Placeholder.
+
         Returns:
-            The collection's version.
+           The collection's version.
```
mfisher87, line 135:
```suggestion
            The collection's version.
```

mfisher87 (changes requested):
A couple more spacing changes. Sorry this is so finnicky :)

danielfromearth (commented):
This looks good!  A couple of additional docstring tweaks:
- I think this [comment for `rebuild_auth()`](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L50) can be put into the method docstring instead
- there's an out-dated and unnecessary [type annotation here](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L96) that can be removed
- I think Google docstring style prefers a spacing between summary line and expanded description (see [examples here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)). Here are the places I found that look like a line spacing could be added:
  - in [refresh_tokens()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L135)
  - in [get_s3_credentials()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L191)
  - in the two hits methods: [hits()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L117) and [hits()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L471)
  - in [provider()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L563)
  - in [set_requests_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L167)
  - in [get_fsspec_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L267)
  - in [get_requests_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L284)
- since we're adding "Placeholder" for the subject line in docstrings in other places, perhaps we should also add them for:
  - the two public functions in [`daac.py`](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/daac.py) 
  - the one public function in [`kerchunk.py`](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/kerchunk.py)
  - [make_instance()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L61)

mfisher87:
Hey @danielfromearth these are all great ideas! I'd like to keep the scope on this PR as small as possible, though. We only added "Placeholder" where we needed it to enable the other parts of the docstring to work without the new linter rules messing stuff up, for example where there was no subject line but there were parameter definitions. Would you mind creating an issue for those items?

mfisher87 (approved):
Things look great, docs look great... last call anyone? @danielfromearth @chuckwondo ?

mfisher87:
:rocket: Thanks so much for your help with locking down our docstrings!

PR #587, opened 2024-06-04
Add GitHub issue metrics automation

mfisher87:
We discussed this on a call today, since this may be useful we can run it manually with the "workflow_dispatch" option.

cc @asteiker @JessicaS11 @jhkennedy @andypbarrett @betolink 

----
📚 Documentation preview 📚: https://earthaccess--587.org.readthedocs.build/en/587/

### DIFF
--- a/.github/workflows/issue-metrics.yml
+++ b/.github/workflows/issue-metrics.yml
@@ -1,0 +1,42 @@
+name: Monthly issue metrics
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '3 2 1 * *'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: issue metrics
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: read
+    steps:
+      - name: Get dates for last month
+        shell: bash
+        run: |
+          # Calculate the first day of the previous month
+          first_day=$(date -d "last month" +%Y-%m-01)
+
+          # Calculate the last day of the previous month
+          last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+
+          #Set an environment variable with the date range
+          echo "$first_day..$last_day"
+          echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
+
+      - name: Run issue-metrics tool
+        uses: github/issue-metrics@v3
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,42 @@
+name: Monthly issue metrics
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '3 2 1 * *'
```
jhkennedy, line 5: "At 02:03 on day-of-month 1" <https://crontab.guru/#3_2_1_*_*>

.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          # Calculate the first day of the previous month
+          first_day=$(date -d "last month" +%Y-%m-01)
+
+          # Calculate the last day of the previous month
+          last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+
+          #Set an environment variable with the date range
+          echo "$first_day..$last_day"
+          echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
```
jhkennedy, line 29: We could make these inputs so we can calculate custom ranges for the workflow dispatch

.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,42 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SEARCH_QUERY: 'repo:${{ github.repository.full_name }} is:issue created:${{ env.last_month }} -reason:"not planned"'
+
+      - name: Create issue
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Monthly issue metrics report
+          token: ${{ secrets.GITHUB_TOKEN }}
+          content-filepath: ./issue_metrics.md
```
jhkennedy, line 42: I wonder if it would be better to open a discussion instead?

.github/workflows/issue-metrics.yml
mfisher87: Not a bad idea, but I don't have time to think about that before the due date :)

.github/workflows/issue-metrics.yml
mfisher87: I would prefer a discussion. This was copy-pasta to get it out quickly so we can use the metrics in our proposal. I won't have time to address this before the due date. Should we go ahead and merge and run the action and open an issue to fix it?

jhkennedy: Yeah, let's merge and run the action, and then you can assign the issue to me and I'll poke at it next week.

PR #596, opened 2024-06-11
Did the necessary docstring cleanup as discussed in #590

### LINKED ISSUE #590: Some docstring cleanup
A couple of additional docstring suggestions pulled from [this comment](https://github.com/nsidc/earthaccess/pull/580#issuecomment-2150346317) in PR #580:
- I think this [comment for `rebuild_auth()`](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L50) can be put into the method docstring instead
- there's an out-dated and unnecessary [type annotation here](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L96) that can be removed
- I think Google docstring style prefers a spacing between summary line and expanded description (see [examples here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)). Here are the places I found that look like a line spacing could be added:
  - in [refresh_tokens()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L135)
  - in [get_s3_credentials()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/auth.py#L191)
  - in the two hits methods: [hits()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L117) and [hits()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L471)
  - in [provider()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/search.py#L563)
  - in [set_requests_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L167)
  - in [get_fsspec_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L267)
  - in [get_requests_session()](https://github.com/Sherwin-14/earthaccess/blob/f8901ade489c8c211658ec13a5757ce1b6259ff9/earthaccess/store.py#L284)

Sherwin-14:
I've done the changes as discussed. Please have a look at it ;). Closes #590

----
📚 Documentation preview 📚: https://earthaccess--596.org.readthedocs.build/en/596/

### DIFF
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -47,9 +47,8 @@ class SessionWithHeaderRedirection(requests.Session):
         if username and password:
             self.auth = (username, password)
 
-    # Overrides from the library to keep headers when redirected to or from
-    # the NASA auth host.
     def rebuild_auth(self, prepared_request: Any, response: Any) -> None:
+        """Overrides from the library to keep headers when redirected to or from the NASA auth host."""
         headers = prepared_request.headers
         url = prepared_request.url
 
@@ -93,7 +92,7 @@ class Auth(object):
                 * **"environment"**:
                     Retrieve a username and password from $EARTHDATA_USERNAME and $EARTHDATA_PASSWORD.
             persist: Will persist credentials in a `.netrc` file.
-            system (Env): the EDL endpoint to log in to Earthdata, defaults to PROD
+            system: the EDL endpoint to log in to Earthdata, defaults to PROD
 
         Returns:
             An instance of Auth.
@@ -132,6 +131,7 @@ class Auth(object):
 
     def refresh_tokens(self) -> bool:
         """Refresh CMR tokens.
+
         Tokens are used to do authenticated queries on CMR for restricted and early access datasets.
         This method renews the tokens to make sure we can query the collections allowed to our EDL user.
         """
@@ -188,6 +188,7 @@ class Auth(object):
         endpoint: Optional[str] = None,
     ) -> Dict[str, str]:
         """Gets AWS S3 credentials for a given NASA cloud provider.
+
         The easier way is to use the DAAC short name; provider is optional if we know it.
 
         Parameters:
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -165,6 +165,7 @@ class Store(object):
         self, url: str, method: str = "get", bearer_token: bool = False
     ) -> None:
         """Sets up a `requests` session with bearer tokens that are used by CMR.
+
         Mainly used to get the authentication cookies from different DAACs and URS.
         This HTTPS session can be used to download granules if we want to use a direct,
         lower level API.
@@ -265,6 +266,7 @@ class Store(object):
     @lru_cache
     def get_fsspec_session(self) -> fsspec.AbstractFileSystem:
         """Returns a fsspec HTTPS session with bearer tokens that are used by CMR.
+
         This HTTPS session can be used to download granules if we want to use a direct,
         lower level API.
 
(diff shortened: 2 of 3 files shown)


earthaccess/auth.py
```diff
@@ -93,14 +92,16 @@ def login(
...
 
         Returns:
             An instance of Auth.
         """
         if system is not None:
             self._set_earthdata_system(system)
 
+        if system is not None:
+            self._set_earthdata_system(system)
+
```
chuckwondo, line 104: Looks like this is a duplicate code block.
```suggestion
```

chuckwondo (changes requested):

earthaccess/auth.py
```diff
@@ -93,7 +92,6 @@ def login(
                 * **"environment"**:
                     Retrieve a username and password from $EARTHDATA_USERNAME and $EARTHDATA_PASSWORD.
             persist: Will persist credentials in a `.netrc` file.
-            system (Env): the EDL endpoint to log in to Earthdata, defaults to PROD
```
mfisher87, line 96: We only want to remove the incorrectly-documented type here. Should be `system: the EDL endpoint to log in to Earthdata, defaults to PROD`

See the type signature above, the type is actually `Optional[System]`:

```python
 system: Optional[System] = None,
```

The documentation generator is capable of reading this, so we don't need to also mention the type in the docstring!

earthaccess/daac.py
```diff
@@ -148,6 +149,7 @@ def find_provider(
 
 
 def find_provider_by_shortname(short_name: str, cloud_hosted: bool) -> Union[str, None]:
+    """Placeholder."""
```
mfisher87, line 152: Please remove the docstrings that contain _only_ "Placeholder". We were using that to fix up docstrings that contained other things (like parameter descriptions) but no subject line. For these undocumented functions, the correct fix is to write a real docstring, but we can consider that out of scope for this PR.

Additionally, if we put placeholder text, we can't use the Ruff D1 rules to catch functions, etc. without docstrings!

earthaccess/auth.py
```diff
@@ -47,9 +47,8 @@ def __init__(
         if username and password:
             self.auth = (username, password)
 
-    # Overrides from the library to keep headers when redirected to or from
-    # the NASA auth host.
```
mfisher87, line 51: Great catch!

mfisher87 (changes requested):
See above for the suggestions, I forgot to start a review and just added comments :)

Created an issue to track the need to deal with the placeholder text: https://github.com/nsidc/earthaccess/issues/597

earthaccess/daac.py
Sherwin-14: Sorry about that! forgot about your last comment on ignoring placeholder for functions with no docstrings.

mfisher87: No sweat :)

earthaccess/store.py
```diff
@@ -62,6 +62,7 @@ def make_instance(
     cls: Any, granule: DataGranule, auth: Auth, data: Any
 ) -> EarthAccessFile:
     # Attempt to re-authenticate
+    """Placeholder."""
```
mfisher87, line 65: Missed one ;)

mfisher87 (approved):
:sparkles: Looks great!

PR #603, opened 2024-06-13
remove the unreferenced `CLOUD_PROVIDERS` list

### LINKED ISSUE #564: Additional hard-coded providers needed for OB.DAAC
#516 didn't get all the changes needed for adding a new DAAC.

Additional changes needed to (at least):
- https://github.com/nsidc/earthaccess/blob/f79a005b0a75e0e1261111a72589a4d678f3c819/earthaccess/daac.py#L109-L119

Related to #563.

itcarroll:
Closes #564.

The bug does not appear to be that "OBDAAC" is missing from `earthaccess.daac.CLOUD_PROVIDERS`. Rather, the issue seems to be that this list exists without having any function. The `CLOUD_PROVIDERS` list is not referenced in the project (according to VS Code and text search) and deleting it breaks no tests. This PR removes the cruft and does nothing else.

Please close without merging if you're aware of any purpose for this variable.

As far as adding OBDAAC goes, I think it was added correctly in #516.

----
📚 Documentation preview 📚: https://earthaccess--603.org.readthedocs.build/en/603/

### DIFF
--- a/earthaccess/daac.py
+++ b/earthaccess/daac.py
@@ -106,18 +106,6 @@ DAACS = [
 ]
 
 
-CLOUD_PROVIDERS = [
-    "GES_DISC",
-    "LPCLOUD",
-    "NSIDC_CPRD",
-    "POCLOUD",
-    "ASF",
-    "GHRC_DAAC",
-    "ORNL_CLOUD",
-    "LAADS",
-    "LARC_CLOUD",
-]
-
 # Some testing urls behind EDL
 DAAC_TEST_URLS = [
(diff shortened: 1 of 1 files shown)


mfisher87 (approved):
Great catch! If we need a structure like this we can derive this from `DAACS`.

@betolink do you think we should delete it, or is there a need to keep the variable? In which case we can make it a list comprehension instead of a copy of the data.

betolink:
I thought the code was using this list to flag cloud providers from the results but maybe that logic changed? I don't see where else this is being used so I think it would be OK to remove it.

PR #606, opened 2024-06-18
Opening virtual datasets (dmr-adapter)

### LINKED ISSUE #605: Opening virtual datasets with NASA dmrpp files
The idea is to speed up opening up netcdf4/hdf5 datasets with a NASA specific optimization. Load data like `xr.open_mfdataset` with kerchunk/zarr speeds by translating **existing dmr++ metadata files to zarr metadata on the fly**. [Much more context and discussion here](https://github.com/zarr-developers/VirtualiZarr/issues/85). 

`virtualizarr` [PR for the parser here](https://github.com/zarr-developers/VirtualiZarr/pull/113)

`earthaccess` [PR here](https://github.com/nsidc/earthaccess/pull/606)

`earthaccess` additions:
1. Open a virtual dataset (like a view of the data that contains dimensions, variables, attrs, chunks, but no actual data)

Screenshot 2024-06-18 at 4 13 48 PM

2. Concatenate virtual xr.Datasets
    1. Use `xarray`s concatenation logic to create virtual views of netcdf’s ([more details in `virtualizarr` documentation](https://virtualizarr.readthedocs.io/en/latest/))
    2. Save as json/parquet/in-memory dict
    
Screenshot 2024-06-18 at 4 14 13 PM
    
3. Read netcdf/hdf5 data
    1. Use the `zarr` engine in `xarray` to load a dataset (with indexes)
    
    
Screenshot 2024-06-18 at 4 14 31 PM

Questions/Suggestions:

Changes to the API?
- **xr.combine_nested kwargs might be confusing and could be reworked if needed

NASA datasets you want me to test?
- Many with unique considerations and the goal is to handle all NASA netcdf4/hdf5 datasets
- So far I’ve tested and had success `MUR-JPL-L4-GLOB-v4.1` (netcdf), `SWOT_SSH_2.0` (netcdf), `ATL-03` ICE-SAT (hdf5)
- Feel free to test it out yourself or add a comment with a dataset and I can take a look

Take a look at the [virtualizarr parser PR](https://github.com/zarr-developers/VirtualiZarr/pull/113) and leave suggestions
- Speed improvements?
- The current bottleneck is creating the xml object (dmrpp are xml files) with `ET.fromstring(dmr_str)` since the `xml.ElementTree` library needs to read the text, validate the xml, and create a parsable object. I am looking into a non-validating parser like `xml.parsers.expat`

ayushnag:
- [x] Closes #605
- [x] Add docs
- [x] Unit tests. Update current test to test specific portions of the virtual dataset
- [x] Check indirect dmrpp reading support
- [x] Use updated `virtualizarr` version (with numpy 2.0 manifest)
- [x] Update CHANGELOG.md
- [x] ~~Add tutorial notebook [TODO in later PR]~~

----
📚 Documentation preview 📚: https://earthaccess--606.org.readthedocs.build/en/606/


earthaccess/virtualizarr.py
```diff
@@ -0,0 +1,112 @@
...
+    ----------
+    xr.Dataset
+    The virtual dataset (with virtualizarr ManifestArrays)
+
+    Raises
+    ----------
+    Exception
+    If the DMR++ file is not found or if there is an error parsing the DMR++
+    """
+    from virtualizarr.readers.dmrpp import DMRParser
```
TomNicholas, line 34: It's interesting that you're not actually using the `filetype='dmr++'` option to `virtualizarr.open_virtual_dataset` here. It seems to me that one alternative option would be for everything in https://github.com/zarr-developers/VirtualiZarr/pull/113 to also live in this library, as it already pretty much entirely uses public virtualizarr API... But I guess that depends whether you think the dmr++ option to `virtualizarr.open_virtual_dataset` is likely to be useful outside of the context of the `earthaccess` library.

ayushnag: The only reason is because the parser has an additional kwarg `data_filepath` which is required in cases where the dmr path cannot be simply derived by just adding `.dmrpp`. If there is a way for `engine` specific args in `virtualizarr.open_dataset` then I can switch to that

TomNicholas: > The only reason is because the parser has an additional kwarg data_filepath which is required in cases where the dmr path cannot be simply derived by just adding .dmrpp.

I'm not sure I understand - why is the main `filepath` you pass to `virtualizarr.open_virtual_dataset` not sufficient?

ayushnag: For this case: `virtualizarrr.open_dataset(filepath=“s3://air.dmrpp”, data_filepath=“s3://datafiles/air.nc”, engine="dmr++")` when the dmr path is independent from data path. The chunk manifest needs to store the `data_filepath` instead of the dmr filepath

TomNicholas: Huh - does the dmr++ data not contain the path to the original data??

ayushnag: No the dmr file only contains the file name and not the full path. This is an example of the name that a dmr file contains `name="20210715090000-JPL-L4_GHRSST-SSTfnd-MUR-GLOB-v02.0-fv04.1.nc"`

betolink:
This PR looks good to me (we need to fix some minor formatting issues with Ruff). Maybe the only missing thing would be a notebook demonstrating how to use this feature? @ayushnag

ayushnag:
https://gist.github.com/ayushnag/bcf946a71122f5e7a54bc72b581bd31b

earthaccess/virtualizarr.py
ayushnag: I noticed that [virtualizarr renaming paths](https://virtualizarr.readthedocs.io/en/latest/generated/virtualizarr.xarray.VirtualiZarrDatasetAccessor.rename_paths.html#virtualizarr.xarray.VirtualiZarrDatasetAccessor.rename_paths) was added and I think this can solve my issue. I will just call `vz.open_dataset` and then rename the data paths using `earthaccess` results. Then I can switch to the public `virtualizarr` API now and remove the `_parse_dmr` function

TomNicholas: So what does that imply for my original question:

> It seems to me that one alternative option would be for everything in https://github.com/zarr-developers/VirtualiZarr/pull/113 to also live in this library, as it already pretty much entirely uses public virtualizarr API...

ayushnag: @betolink How about if I move the parser code into `earthaccess`? Now that I'm thinking about it makes more sense to be in the NASA related repository. It would also make unit tests easier since `earthaccess` can easily access NASA dmrpp files. Then this PR will add `dmrpp.py` and `virtualizarr.py`

ayushnag:
Yes I can clarify. At some point the dmrpp parser was going to be a part of [earthaccess instead of virtualizarr](https://github.com/zarr-developers/VirtualiZarr/pull/113#pullrequestreview-2248868063). However that is not the case anymore and this PR will be updated to just call virtualizarr directly

ayushnag:
This PR is ready to review now. I have updated the `pyproject.toml` with an added dependency but let me know if the `uv.lock` or other files need to modified.

betolink:
I can take a look this evening, great work @ayushnag !!

betolink (approved):
Great work Ayush! thank you for all the work you put on this PR and what you did on `virtualizarr`. I hope we can talk with @jgallagher59701 and Miguel Jimenez today/tomorrow about what's next. Maybe we can have people hacking on this access pattern at the Pangeo event!

ayushnag:
@betolink One quick comment before we merge. This function doesn't appear to show up in the API reference with the docs build. I think if the new functions are imported with [.api that should fix it](https://github.com/nsidc/earthaccess/blob/ffbfddd9e0aeacd7285f460c14117bb141ea4560/earthaccess/__init__.py#L6)?

ayushnag:
Although then the function has to be part of `api.py`. Maybe not in that case

betolink:
> @betolink One quick comment before we merge. This function doesn't appear to show up in the API reference with the docs build. I think if the new functions are imported with [.api that should fix it](https://github.com/nsidc/earthaccess/blob/ffbfddd9e0aeacd7285f460c14117bb141ea4560/earthaccess/__init__.py#L6)?

i think you're correct, I don't think the function needs to be part of API but it definitely needs to be imported, if we want the dosc to show there. I think since we are including it also in the `__init__.py` it's already part of `earthaccess.`, good catch! do you want to push one commit and see how the docs render?

danielfromearth:
> I think this PR is ready to be merged, there are many considerations to the future of virtual datasets with NASA data but this PR get us closer to a workflow where we won't have to generate metadata when there is a format already available. We need to produce a few examples and put it in the documentation, maybe using SWOT, ICESat-2 and TEMPO datasets cc @danielfromearth @DeanHenze
> 
> Is there anything we are missing @TomNicholas @ayushnag?

@betolink, do you think we should create a new issue for examples with open virtual dataset?

pyproject.toml
```diff
@@ -64,6 +64,9 @@ kerchunk = [
   "h5netcdf",
   "xarray",
 ]
+virtualizarr = [
+  "virtualizarr @ git+https://github.com/zarr-developers/VirtualiZarr.git"
```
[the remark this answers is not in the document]
TomNicholas: I released like 2 days ago, so you probably want to just pin to `>=1.2.0`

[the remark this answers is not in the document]
ayushnag: Great, will update that!

earthaccess/__init__.py
```diff
@@ -58,6 +59,9 @@
     "Store",
     # kerchunk
     "consolidate_metadata",
+    # virtualizarr
+    "open_virtual_dataset",
+    "open_virtual_mfdataset",
```
TomNicholas, line 64: We should probably add a `virtualizarr.open_virtual_mfdataset` upstream in virtualizarr. Though I would like to be more confident about the best way to parallelize reference generation first

https://github.com/zarr-developers/VirtualiZarr/issues/123

https://github.com/zarr-developers/VirtualiZarr/issues/7

TomNicholas: tracked in https://github.com/zarr-developers/VirtualiZarr/issues/345

earthaccess/dmrpp_zarr.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def open_virtual_mfdataset(
+    granules: list[earthaccess.DataGranule],
+    group: str | None = None,
+    access: str = "indirect",
+    load: bool = False,
+    preprocess: callable | None = None,  # type: ignore
+    parallel: bool = True,
+    **xr_combine_nested_kwargs: Any,
```
TomNicholas, line 18: Technically if you are able to load coordinates into memory (and therefore create pandas indexes) you could also support `xr.combine_by_coords` here too.

earthaccess/dmrpp_zarr.py
```diff
@@ -0,0 +1,199 @@
...
+        vdatasets = [preprocess(ds) for ds in vdatasets]
+    if parallel:
+        vdatasets = dask.compute(vdatasets)[0]  # type: ignore
+    if len(vdatasets) == 1:
+        vds = vdatasets[0]
+    else:
+        vds = xr.combine_nested(vdatasets, **xr_combine_nested_kwargs)
+    if load:
+        refs = vds.virtualize.to_kerchunk(filepath=None, format="dict")
+        return xr.open_dataset(
```
TomNicholas, line 130: So if we implemented @ayushnag 's idea in https://github.com/zarr-developers/VirtualiZarr/issues/124 then we could just call that instead of using `xarray.open_dataset` here.

tests/integration/test_virtualizarr.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_dmrpp(granule):
+    from virtualizarr import open_virtual_dataset  # type: ignore
+
+    fs = earthaccess.get_fsspec_https_session()
+    data_path = granule.data_links(access="indirect")[0]
+    dmrpp_path = data_path + ".dmrpp"
+
+    result = open_virtual_dataset(
```
TomNicholas, line 33: So the test of correctness here is assuming that virtualizarr has correct behaviour? You might want to add more tests because virtualizarr definitely still has bugs in it (e.g. around CF coordinate decoding).

TomNicholas (commented):
Amazing work @ayushnag.

tests/integration/test_virtualizarr.py
ayushnag: Yes, we are assuming that virtualizarr has the correct behavior. However this is mostly an integration test checking that the function operates as expected. The actual tests for correctness are in virtualizarr (for the dmrpp backend)

ayushnag:
@TomNicholas do you think after changing the virtualizarr version in `pyproject.toml` we can merge? The other changes seem more like long term updates to me. Or which changes do you think are required now?

betolink (approved):
Docs look good, we're merging now!

betolink (approved):
We'll address @TomNicholas comments about pushing some functionality to VirtualiZarr, other than that we need the example notebooks!

Mikejmnez:
Just catching up here. Awesome work @ayushnag  🥇 !

PR #620, opened 2024-06-26
the earthaccess.EarthAccessFile wrapper need not subclass anything

### LINKED ISSUE #610: method resolution surprise on `EarthAccessFile`
There's something funny going on with the wrapping of `fsspec.spec.AbstractBufferedFile`, where the method resolution isn't coming out as expected.

The following gives me an `AssertionError`
```python
import earthaccess

auth = earthaccess.login()
results = earthaccess.search_data(short_name="VIIRSJ1_L2_OC", count=1)
files = earthaccess.open(results)
with files[0] as g:
    assert g.read is getattr(g.f, "read")
```

I don't know about the internals, but `getattr(g, "read")` and `g.__getattribute__("read")` both follow the MRO to the `read` method of `fsspec.spec.AbstractBufferedFile`. We don't want that! We want the `read` method, given by `g.__getattr__("read")` as `fsspec.implementations.http.HTTPFile.read` as resulting from `getattr(g.f)` in the class definition.

https://github.com/nsidc/earthaccess/blob/b7443fe217d78313c7a0c9a6708132d0aa31a11f/earthaccess/store.py#L28-L34

Is it possible that `__getattr__` should be `__getattribute__`?

*Sorry for the multiple edits; confused myself with two `f`s*

itcarroll:
Fixes #610.

This PR removes a base class from the definition of `earthaccess.EarthAccessFile`. With inheritance, methods defined on `fsspec.spec.AbstractBufferedFilesystem` are discovered before any overrides defined on `f`.

The `earthaccess.EarthAccessFile` wrapper is handed the results of a `fsspec.filesystem().open()` call. That result is a `fsspec.spec.AbstractBufferedFile` (this PR fixes the type hint on that, btw) and stored in `f`. This wrapper's main job is to pass attribute calls off to `f`. The wrapper itself is not a file-like object; we can't make fsspec give us a custom subclass.

Would appreciate running workflows on this. Unit tests are good, but I still can't get all the integration tests working on openscapes.2i2c.cloud.

ToDo if integration tests look okay:
- [ ] add to changelog
- [ ] check for any changes needed to docs

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -25,8 +25,10 @@ from .search import DataCollections
 logger = logging.getLogger(__name__)
 
 
-class EarthAccessFile(fsspec.spec.AbstractBufferedFile):
-    def __init__(self, f: fsspec.AbstractFileSystem, granule: DataGranule) -> None:
+class EarthAccessFile:
+    def __init__(
+        self, f: fsspec.spec.AbstractBufferedFile, granule: DataGranule
+    ) -> None:
         self.f = f
         self.granule = granule
 
@@ -42,7 +44,7 @@ class EarthAccessFile(fsspec.spec.AbstractBufferedFile):
         )
 
     def __repr__(self) -> str:
-        return str(self.f)
+        return repr(self.f)
 
 
 def _open_files(


mfisher87:
I need to play with this a little bit to better understand what's going on, but I may not have time until the next hack day.

chuckwondo:
> I need to play with this a little bit to better understand what's going on, but I may not have time until the next hack day.

@mfisher87, have you had a chance to do this?

@itcarroll or @betolink, I suppose the larger question for me is, what's the point of this class to begin with? Why do we even need it?

itcarroll:
Thanks for checking on this one @chuckwondo! My guess on the need for this class was something to do with deserializing into a useable object in case authentication timed out. A guess only though, as I don't know when `EarthAccessFile.__reduce__` would be called.

betolink:
@chuckwondo 

> I suppose the larger question for me is, what's the point of this class to begin with? Why do we even need it?

@jrbourbeau can explain in detail but the gist of it is that this class allows a serialization trick, if we open granules from our laptop but we are offloading an xarray operation to a Dask cluster in `us-west-2`, it will re-open the files in place using `s3://url` instead of the `https://cloud-front-tea` url. I think James mentioned that the speed improvement was ~2x vs only using HTTPS.

itcarroll:
@jrbourbeau The essential question is whether removing `fsspec.spec.AbstractBufferedFile` from the MRO of `EarthAccessFile` will break the usage @betolink describes. I don't think we have a test using coiled. (We still have the file-like instance attached as the `f` attribute and used in the `__getattr__` redirection.)

chuckwondo:
> > I suppose the larger question for me is, what's the point of this class to begin with? Why do we even need it?
> 
> @jrbourbeau can explain in detail but the gist of it is that this class allows a serialization trick, if we open granules from our laptop but we are offloading an xarray operation to a Dask cluster in `us-west-2`, it will re-open the files in place using `s3://url` instead of the `https://cloud-front-tea` url. I think James mentioned that the speed improvement was ~2x vs only using HTTPS.

I didn't notice the `make_instance` function earlier, but now that I've looked at it, your description is now more clear to me.

However, I'm questioning the whole idea of pickling the `fsspec.spec.AbstractBufferedFile` to begin with. Opening files from your laptop, then distributing the open files across a cluster seems like you may be inviting more trouble than it's worth. Pickling potentially oddly complex objects to distribute them to other processes can be more problematic than passing around simpler things, such as the URLs from which the files were opened, rather than the opened files themselves.

I may very well be thinking about this poorly, or simply be too inexperienced with using dask (and the like), but without seeing an example of the types of things you think this would be helpful for, offhand I would ask you why you're not simply distributing the URLs rather than the open files?

mfisher87:
>have you had a chance to do this?

I have not yet, sorry :X

chuckwondo:
> > > I suppose the larger question for me is, what's the point of this class to begin with? Why do we even need it?
> > 
> > 
> > @jrbourbeau can explain in detail but the gist of it is that this class allows a serialization trick, if we open granules from our laptop but we are offloading an xarray operation to a Dask cluster in `us-west-2`, it will re-open the files in place using `s3://url` instead of the `https://cloud-front-tea` url. I think James mentioned that the speed improvement was ~2x vs only using HTTPS.
> 
> I didn't notice the `make_instance` function earlier, but now that I've looked at it, your description is now more clear to me.
> 
> However, I'm questioning the whole idea of pickling the `fsspec.spec.AbstractBufferedFile` to begin with. Opening files from your laptop, then distributing the open files across a cluster seems like you may be inviting more trouble than it's worth. Pickling potentially oddly complex objects to distribute them to other processes can be more problematic than passing around simpler things, such as the URLs from which the files were opened, rather than the opened files themselves.
> 
> I may very well be thinking about this poorly, or simply be too inexperienced with using dask (and the like), but without seeing an example of the types of things you think this would be helpful for, offhand I would ask you why you're not simply distributing the URLs rather than the open files?
> 
> Is it so that you don't have to also potentially distribute credentials across such clusters as well?

Regarding the specific changes in the PR, they look fine to me, but is anybody able to answer my question above?

chuckwondo (approved):
LGTM

itcarroll:
pre-commit.ci autofix

tests/unit/test_store.py
```diff
@@ -126,3 +127,12 @@ def test_store_can_create_s3_fsspec_session(self):
             store.get_s3_filesystem()
 
         return None
+
+
+def test_earthaccess_file_getattr():
```
jrbourbeau, line 132: Nice. Can confirm this fails on `main` as expected

```
____________________________________________________________________________________________________ test_earthaccess_file_getattr ____________________________________________________________________________________________________

    def test_earthaccess_file_getattr():
        fs = fsspec.filesystem("memory")
        with fs.open("/foo", "wb") as f:
            earthaccess_file = EarthAccessFile(f, granule="foo")
>           assert f.tell() == earthaccess_file.tell()

tests/unit/test_store.py:135:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../mambaforge/envs/earthaccess-dev4/lib/python3.11/site-packages/fsspec/spec.py:1711: in tell
    return self.loc
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fsspec.implementations.memory.MemoryFile object at 0x104bd7d80>, method = 'loc'

    def __getattr__(self, method: str) -> Any:
>       return getattr(self.f, method)
E       AttributeError: 'MemoryFile' object has no attribute 'loc'

earthaccess/store.py:35: AttributeError
```

but passes here

jrbourbeau (approved):
Thanks @itcarroll! Apologies for the lack of engagement here. The changes here LGTM. Though not sure exactly what's going on with the integration tests 

> the gist of it is that this class allows a serialization trick, if we open granules from our laptop but we are offloading an xarray operation to a Dask cluster in us-west-2, it will re-open the files in place using s3://url instead of the https://cloud-front-tea url

Yeah, that's exactly right :+1: Obviously there's no http <--> s3 swapping when granules are opened and computed on in the same environment (i.e. both happen outside us-west-2 or both happen inside us-west-2) 

> The essential question is whether removing fsspec.spec.AbstractBufferedFile from the MRO of EarthAccessFile will break the usage @betolink describes

Nope -- we still keep the https <--> s3 behavior but fix the attribute lookup like you mentioned

itcarroll:
@chuckwondo:
> Pickling potentially oddly complex objects to distribute them to other processes can be more problematic than passing around simpler things, such as the URLs from which the files were opened, rather than the opened files themselves.

Some complexity will be needed to handle the s3 <--> https hand-off between laptops and in-region workers spun up by coiled. Here it's a pickled file-like object, but additional in_region checks in `__store__._get_urls` could be a better approach. It looks like you are okay with making that a separate issue for an enhancement, rather than addressing it in this bug fix.

itcarroll:
No worries @jrbourbeau, thanks for the review!

> Though not sure exactly what's going on with the integration tests

Someone with write-access needs to trigger them is all.

chuckwondo:
> No worries @jrbourbeau, thanks for the review!
> 
> > Though not sure exactly what's going on with the integration tests
> 
> Someone with write-access needs to trigger them is all.

This is not yet possible. Once #808 lands, this will be possible. If you are willing to wait for #808 to be approved and merged to main, then you could merge main into your PR, and this will retrigger int. tests, which will then fail again, but then a maintainer will be able to re-run the failed int. tests so that they can pass (assuming nothing was broken).

jrbourbeau:
> Someone with write-access needs to trigger them is all
> This is not yet possible

Not sure exactly what the state of integrations tests are but FWIW I went ahead and bumped those failing CI jobs and they're passing now

chuckwondo:
> > Someone with write-access needs to trigger them is all
> > This is not yet possible
> 
> Not sure exactly what the state of integrations tests are but FWIW I went ahead and bumped those failing CI jobs and they're passing now

LOL! Apologies. I'm getting my PRs mixed up. I forgot that part already landed 🤦

Issue #618, opened 2024-06-26
Remove duplicate dependency lists in conda environment YAML files


mfisher87:
We could:

* Remove them. Personally, I don't think this pure python project has a reason to include them. We can advise conda users to use conda to create the environment, but use pip to install into it.
* Replace them with environment specification files that don't duplicate the dependency lists:
    ```yaml
    name: earthaccess
    channels:
      - conda-forge
      - nodefaults
    dependencies:
      - python>=3.9
      - pip
      - pip:
          - --editable .[dev]
    ```

itcarroll:
I knew I could find another one fixed by #733. @mfisher87 was on a tear!

mfisher87:
Love it! Thanks for digging for these Ian!!

PR #733, opened 2024-06-30
Replace `poetry` with `pip`, build with `hatchling`, dynamically calculate minimum dependencies with `uv`

### LINKED ISSUE #374: Switch away from Poetry for build (to setuptools?)

### LINKED ISSUE #619: Reconsider dependency groups
Right now we only have a dev group. In that group, many things are listed, ranging from testing to docs to linting. IMO, we should have a `dev` group which includes only dev tools like pre-commit and Nox which are capable of managing their own environments. Then we can have two more groups: `docs` for mkdocs and friends. `test` for pytest, mypy, and friends. We can then use nox to handle running tasks like building docs, typechecking, and testing. Pre-commit handles everything that doesn't require the project's dependencies, like linting and formatting and various miscellaneous safety checks.

Regardless of whether we use Nox, I think it will be valuable to break out these 3 groups. More discussion here! https://github.com/nsidc/earthaccess/issues/614

### LINKED ISSUE #734: Test workflow runs double on pull requests
It runs for the "push" event and the "pull_request" event. We should only run for push events to specific branches (only main for now) to avoid this.

mfisher87:
"Dependency groups" are a poetry-specific feature that's different from PEP621 "optional dependencies", aka "extras". To make things compatible with other ways of dependency management, switch to PEP621 optional dependencies, but specified using the Poetry configuration aliases to retain support for poetry. While the Poetry "dependency groups" functionality may be more aimed at development workflows, it's not compatible with other tools and in my experience the dominant convention in the ecosystem is to use optional dependencies. As @itcarroll points out below, [Hatch supports a similar concept called "environment dependencies" and specifies them with a different tool-specific configuration key](https://hatch.pypa.io/latest/config/environment/overview/#dependencies). PDM _also_ calls it something different, ["dev dependencies"](https://pdm-project.org/latest/usage/dependency/#add-development-only-dependencies), and uses another unique config table. :sweat: 

I really don't like the way poetry requires optional dependencies to be specified. It requires duplication.

Thoughts? I don't want to add all this duplication to our codebase. Can we go forward with replacing Poetry with good ol' setuptools or maybe hatchling for builds and switch the optional dependency groups to the standard format? Anything PEP621 compatible would be a step forward IMO. I can take that on as part of this PR. The [PEP621 example](https://peps.python.org/pep-0621/#example) shows our use case, test dependencies, implemented in the conventional way.

Closes #734 
Closes #619 
Closes #374 

----
📚 Documentation preview 📚: https://earthaccess--733.org.readthedocs.build/en/733/

TODO:

- [x] Update docs


.pre-commit-config.yaml
```diff
@@ -12,6 +12,14 @@ repos:
       - id: check-toml
       - id: check-json
 
+  - repo: https://github.com/python-poetry/poetry
+    rev: '1.8.3'
+    hooks:
+      - id: poetry-check
```
mfisher87, line 18: this will fail if the lockfile hasn't been updated, or if there's a typo in a package name in poetry config.

itcarroll:
> While the Poetry "dependency groups" functionality may be more aimed at development workflows, it's not compatible with other tools and in my experience the dominant convention in the ecosystem is to use optional dependencies.

I have been exploring hatch w/ hatchling, and it is both PEP compliant and also (like Poetry) avoids using extras (a.k.a optional-dependencies) for managing development environments. Information about extras is built into the distribution metadata and primarily provides extra packages users might want (i.e. `xarray[io]` if you want netCDF4, etc.). Hatch (and Poetry) allow separate ways of providing packages for and setting up development environments; information not included in distribution metadata. I haven't been compelled to duplicate package lists for simple cases with Hatch. Note that the current potential downside with Hatch is a lack of lock files.

mfisher87:
Just to clarify, I'm for including these optional dependency groups in the package metadata, as that's in my experience a very common method and is compatible with all environment package managers, and as far as I know there is no standard way to specify dependency groups that do not go in the package metadata.

I'd like to avoid pros and cons of environment / package managers here (i.e. Hatch, Poetry), as we can isolate that concern from our build tooling and configuration. Regardless of whether we decide down the road to document / encourage Hatch or Poetry or something else for package management, if our build configuration is PEP621 compliant, contributors can choose any tool.

mfisher87:
To make things more confusing, [PDM has "dev dependencies"](https://pdm-project.org/latest/usage/dependency/#add-development-only-dependencies), their own uniquely named version of Poetry's "dependency groups" and Hatch's "environment dependencies". Wow! The same thing three times with three names and three config keys.

I'd really like to treat this Python environment management feature as non-standard and avoid it until there's a unified interface, which I'm sure there's already an open PEP for :laughing:

mfisher87:
Found it: https://peps.python.org/pep-0735/

chuckwondo (approved):
@mfisher87, sorry this has remained unreviewed for so long! LGTM!

I'm approving, but do you need to update any "contributing" instructions regarding commands that a contributor needs to run to get their environment setup with dev/test deps?

mfisher87:
Thanks, Chuck! Good point, I do imagine I missed some stuff there.

TODO:

- [ ] Ensure docs are up-to-date with dependency management changes

mfisher87:
@chuckwondo @danielfromearth I committed the changes we were working on in the call, and clearly they are no longer poetry compatible :) I didn't have time to debug yet!

danielfromearth:
So, where do we want to go from here?  If there's no way to remove the redundant dependency specifications, then get this working with another manager, like hatch, and then drop the poetry usage?

pyproject.toml
```diff
@@ -31,76 +31,85 @@ classifiers = [
...
-tinynetrc = "^1.3.1"
-multimethod = ">=1.8"
-kerchunk = { version = ">=0.1.2", optional = true }
+dependencies = [
+  "python-cmr >=0.10.0",
+  "pqdm >=0.1",
+  "requests >=2.26",
+  "s3fs >=2022.11",
+  "fsspec >=2022.11",
+  "tinynetrc ^1.3.1",
```
itcarroll, line 45: Caret requirements are not supported in the project.dependencies table.

itcarroll: Oh, nevermind. this is still poetry dependencies.

mfisher87: Good catch! IMO we shouldn't be using caret notation since it's non-standard.

mfisher87:
Personally, dropping poetry in favor of hatch, uv, pip, and all the other standards-compliant dependency managers feels like the right way forward. Do we have consensus on that? I'm not sure :) #761

itcarroll:
> dropping poetry in favor of hatch, uv, pip

Reminder that hatch does not support lock files (waiting for a standard, I believe).

> So, where do we want to go from here?

An option is to wait until poetry supports the project table (https://github.com/python-poetry/poetry/issues/9448). On that ... I am confused by the inclusion of the `project.optional-dependencies` table in this PR, because AFAIK poetry does not support the project table at all. Is it included here to demonstrate that failure (maybe mark as "Draft" in that case)?

mfisher87:
>I am confused by the inclusion of the `project.optional-dependencies` table in this PR, because AFAIK poetry does not support the project table at all

The most recent commit was made without fully understanding that fact in an effort to test what would happen :) If we revert that commit, it goes back to the poetry-specific specification. But... is that what we want? :shrug:

danielfromearth:
hmm, hadn't seen `uv` yet, but that looks quite promising!

It does look like the particular poetry PR for the `project` section dependencies (https://github.com/python-poetry/poetry/pull/9135) might be close to being finalized (all tests are passing just marked ready-for-review 4 days ago).

docs/contributing/development.md
```diff
@@ -1,62 +1,54 @@
...
-`pyproject.toml` (after changing `^` to `>=` there), and that development
-dependencies should be left unpinned.
+## Managing Dependencies
 
-Finally, for _development dependencies only_, you must add an entry to
-`ci/environment-dev.yaml` with the same version constraint as in
-`pyproject.toml`.
+If you need to add a new dependency, edit `pyproject.toml` and insert the dependency in
+the correct location (either in the `dependencies` array or
+`[project.optional-dependencies]`.
```
mfisher87, line 54: `uv` and `pixi` provide good `add` user experiences, without the need to do the weird replacement of `^`s.

pyproject.toml
mfisher87: edd6442

.github/workflows/test-mindeps.yml
```diff
@@ -25,25 +20,26 @@ jobs:
...
-          channel-priority: strict
-          environment-file: ci/environment-mindeps.yaml
-          activate-environment: test-environment
-          auto-activate-base: false
+          python-version-file: pyproject.toml
 
-      - name: Install
-        run: python -m pip install --no-deps -e .
+      - name: Install minimum-compatible dependencies
+        run: uv sync --resolution lowest-direct
```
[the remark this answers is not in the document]
mfisher87: There are things other than `uv` that can do "lowest" resolution and lockfiles, I just reached for something I knew was fast and had single-file multi-platform lockfiles and wasn't poetry.

ci/environment-dev.yml
mfisher87, line 1: To me, it doesn't make sense for this to be in the `ci` directory.

.pre-commit-config.yaml
```diff
@@ -33,6 +33,12 @@ repos:
       - id: ruff
         args: ["--fix", "--exit-non-zero-on-fix"]
       - id: ruff-format
+
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: "0.4.7"
+    hooks:
+      - id: uv-lock
```
mfisher87, line 40: Lockfile stays up-to-date without user needing to learn to use any new tooling

.github/workflows/integration-test.yml
```diff
@@ -17,58 +17,29 @@ concurrency:
...
-  run:
-    shell: bash -l {0}
-
 jobs:
   integration-tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.12"]
```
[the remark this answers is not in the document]
betolink: I disagree here, I think it's valuable to test all the versions because there are subtle differences that only arise when we run code.

.github/workflows/test-mindeps.yml
```diff
@@ -25,25 +20,26 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v4.1.1
-      - name: Setup Conda Environment
-        uses: conda-incubator/setup-miniconda@v3.0.4
+
+      - name: Install uv
```
betolink, line 24: this is cool! that we can use fewer (and faster) tooling

binder/environment.yml
```diff
@@ -18,3 +18,7 @@ dependencies:
   - hvplot
   - holoviews
   - panel
+
+  - pip
+  - pip:
+      - "."
```
betolink, line 24: Will this install earthaccess from source in the Binder environment?

scripts/integration-test.sh
```diff
@@ -14,5 +14,3 @@ if [[ $RET == 99 ]]; then
 elif [[ $RET != 0 ]]; then
     exit $RET
 fi
```
betolink, line 16: Are the integration tests still being run in this scenario?

ci/environment-dev.yml
```diff
@@ -2,25 +2,10 @@ name: earthaccess-dev
...
-  - mkdocstrings>=0.19.0
-  - mkdocstrings-python
-  - mkdocs-jupyter>=0.19.0
-  - pymdown-extensions>=9.2
 
   - pip
   - pip:
-      - poetry
-      - markdown-callouts>=0.2.0
+      - --editable .[dev]
```
betolink, line 11: Are the docs deps also installed here?

betolink (commented):
I think this is a great PR, it greatly simplifies the tooling and makes things simpler for developers. I have a couple questions regarding integration tests and the dev environment. Thanks a lot @mfisher87!!

binder/environment.yml
mfisher87: Correct, but not editable. We can change it to `"--editable ."` to achieve that if we want. I haven't tested this -- is there a way to test with Binder before merge to `main`?

scripts/integration-test.sh
mfisher87: What scenario? The changes here, aside from decoupling linting, are from #738

ci/environment-dev.yml
mfisher87: No, but good point, they should be :)

.github/workflows/integration-test.yml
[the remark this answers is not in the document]
mfisher87: Yeah, agree, but we're doing all the versions for our unit tests already, and integration tests are more expensive. I don't know how many trees we'll be saving by cutting our integration test runs in half, but maybe not enough to matter at all :laughing:  In any case, we obviously don't have consensus on this yet, so will change it back and we can talk about it another time :)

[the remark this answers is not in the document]
mfisher87: 89b79d6

environment-dev.yml
mfisher87, line 1: Sneaky last second change: I moved this file out of the `ci/` directory, so now that directory doesn't exist at all.

.github/actions/install-pkg/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Install earthaccess Python package
+description: Install earthaccess Python package and testing dependencies'
```
chuckwondo, line 2:
```suggestion
description: Install earthaccess Python package and testing dependencies
```

.github/actions/install-pkg/action.yml
```diff
@@ -0,0 +1,23 @@
...
+  steps:
+    - uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: pip
+
+    - name: Display full python version
+      shell: bash
+      id: full-python-version
+      run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
```
chuckwondo, line 19: This syntax is deprecated, per https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

Instead, it should be:

```suggestion
      run: echo "{version}=$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")" >> $GITHUB_OUTPUT
```

README.md
```diff
@@ -82,7 +82,7 @@ Visit [our quick start guide](https://earthaccess.readthedocs.io/en/latest/quick
 
 ## Compatibility
 
-Only **Python 3.8+** is supported.
+Only **Python 3.9+** is supported.
```
chuckwondo, line 85: I think the `+` is easy to miss, so it is easy to read the original wording as though _exactly_ Python 3.9 is supported (and nothing else), so this is perhaps a bit more clear:

```suggestion
The _minimum_ supported Python version is **3.9**.
```

docs/contributing/development.md
```diff
@@ -1,62 +1,54 @@
...
+```console
+$ nox -s typecheck              # Typecheck only
+$ nox -s tests                  # Python tests
+$ nox -s build_docs -- --serve  # Build and serve the docs
+$ nox -s build_pkg              # Make an SDist and wheel
 ```
 
-After activating our environment and installing the library with Poetry we can run Jupyter lab and start testing the local distribution or we can use `make` to run the tests and lint the code.
-Now we can create a feature branch and push those changes to our fork!
+Nox handles everything for you, including setting up an temporary virtual
```
chuckwondo, line 31:
```suggestion
Nox handles everything for you, including setting up a temporary virtual
```

docs/contributing/development.md
```diff
@@ -1,62 +1,54 @@
...
+## Manual development environment setup
 
-If we want to use Poetry, first we need to [install it](https://python-poetry.org/docs/#installation). After installing Poetry we can use the same workflow we used for Conda, first we install the library locally:
+Create and activate a virtual environment with `venv`, which comes by default with
+Python, in the `.venv` directory:
 
 ```bash
-poetry install
+python -m venv .venv
+. .venv/bin/activate
```
chuckwondo, line 41: The leading `.` (in place of `source) is easy to miss:

```suggestion
source .venv/bin/activate
```

environment-dev.yml
chuckwondo: Why don't we simply eliminate it now that everything is installed with `pip`? I think leaving this file around can lead to confusion now that there is no mention of using `conda` for anything anymore (unless I missed something, perhaps not shown in the PR), thankfully.

chuckwondo (changes requested):
@mfisher87, fantastic work on cleaning up the dev process! I'm so happy to see conda an poetry gone and things greatly simplified and clarified! I have just a few minor change requests.

.github/actions/install-pkg/action.yml
mfisher87: Nice! Good catch, I just copied it without paying close attention :)

docs/contributing/development.md
mfisher87: Excellent call!

README.md
mfisher87: Agreed!

environment-dev.yml
mfisher87: I don't really feel comfortable making that call in this PR because I think there are some folks that preferred to keep this around based on discussion in #374 #741 #618 and perhaps other spaces I'm forgetting. Although I do agree it would be nice to remove conda as a variable we're responsible for in this repository. Can we treat that as out of scope?

chuckwondo: I think it makes perfect sense to remove this file as part of this PR since the whole point of this PR is to eliminate the existing confusion caused by having multiple dependency files used by multiple tools. I feel that leaving this file in place is counter to the point of this PR, as it still leaves confusion and prevents finally getting us to "one and only one way of doing things."

Regarding the references:

- #374 doesn't seem to mention anything that would support keeping this file in place -- conda isn't even mentioned, other than perhaps in favor of eliminating this file, not keeping it
- #741: I'm a hard "no" on supporting multiple package managers -- I've never run across a project that explicitly does so (of course, I can't say they don't exist, but in my entire career, I can't recall seeing one), and this is precisely why a new contributor almost invariably gets confused about how to start developing/contributing to this project
- #618: I'm all for removing duplication, but I still see no reason to provide instructions for users to create an env using conda, only to have them use nox to do everything else

Of course, we publish to both pypi and conda-forge, so user's of the library are free to use whichever tool they wish for there _own_ projects. As far as _contributors_ go, there should be only _one_ tool, in my opinion.

Again, I really like everything you've done in this PR because it really streamlines, simplifies, clarifies, expedites the contribution process. To me, leaving this file simply leaves one last stumbling block in the way, and I frankly don't see any justification in the above issues for keeping it.

If you are still uncomfortable with removing this file, then I want to see something added to the contributor docs explaining what it's for, why it's hear, and how to use it, making it painfully clear as to why it exists.

docs/contributing/development.md
```diff
@@ -1,62 +1,54 @@
...
+## Manual development environment setup
 
-If we want to use Poetry, first we need to [install it](https://python-poetry.org/docs/#installation). After installing Poetry we can use the same workflow we used for Conda, first we install the library locally:
+Create and activate a virtual environment with `venv`, which comes by default with
+Python, in the `.venv` directory:
 
 ```bash
-poetry install
+python -m venv .venv
+source .venv/bin/activate
```
chuckwondo, line 41: Any reason this isn't exposed as a `nox` command?

Seems like pushing as much into `nox` as possible would further simplify things, and even make it easier to change things in the future (if desired) without requiring a contributor to do anything differently (i.e., I see "nox" as the interface and the stuff in the noxfile as the "implementation", such that "nox" encapsulates things, allowing use to change the implementation as necessary, without affecting the nox user).

docs/contributing/development.md
```diff
@@ -1,62 +1,54 @@
...
 ```
 
-and now we can run the local Jupyter Lab and run the scripts etc. using Poetry:
+Install _earthaccess_ in editable mode with optional development dependencies:
 
 ```bash
-poetry run jupyter lab
+pip install --editable .[dev,test,docs]
```
chuckwondo, line 47: Can we also expose this as a nox command instead?

environment-dev.yml
mfisher87: I'm totally on the same page with you on this... in a vacuum :)  My main intent with this PR is just to remove poetry, not necessarily to excise conda. And while I agree removing this file is the correct decision, I simply don't feel we have consensus on that yet. I went as far as I feel we do have consensus, which is deduplicating the dependency lists per #618.

Would you be willing to add such an entry to the contributing docs? I would like to get this merged in ASAP, since there's so much value to our contributors in the simplifications, and I won't have time to write more docs for a bit. Maybe at the next hack day coming up!

docs/contributing/development.md
mfisher87: I haven't done or seen that done before -- I assumed that environment activation would not be possible to do within a `nox` command because the command doesn't execute in the current shell?

docs/contributing/development.md
mfisher87: I'm not sure! What would this look like? It doesn't really map to how I normally use and think about `nox`. In my mental model I create "sessions" as independent and stateless things, e.g. running typechecking nox creates and can cache an environment with mypy and all the runtime dependencies, and for linting, nox can depend on an environment that solely contains ruff.

Maybe this is more common than I think, but using nox to do setup just doesn't seem natural to me :)

chuckwondo: I consider `nox` analogous to `make`, both of which I consider being able to wrap "complex" commands within simpler commands. So, rather than having to type the `pip` command above, someone might instead be able to type `nox install` and be done with it, without fat-fingering arguments.

To take this further, `nox` (and `make` et al.) can eliminate duplication. Another suggestion I have is to modify the github workflows to _also_ use `nox` rather than the "bare" commands. This way, both a contributor and github are using the same commands (both using `nox`). Granted, a contributor might not run all of the things that github would, but by having both use `nox` (in our case), there's no need to update docs or workflows to match changes to the noxfile (for _existing_ commands), thus eliminating (or greatly reducing) the likelihood of docs and files getting out of sync.

However, to expedite landing this PR, we can open a separate issue to address this.

docs/contributing/development.md
chuckwondo: Sorry, yes, I was looking only at the venv creation. Activation, of course, must be done explicitly by the user, so that's one bit that "leaks" through to the user, which cannot be encapsulated/hidden by `nox` (or `make`).

I'm suggesting we put as much as possible behind `nox`, so that as much as possible, the user uses `nox` as the one way of doing things.

environment-dev.yml
chuckwondo: Fair enough. However, part of my point here is that I have no idea how to document this, because I don't even know what you are attempting to achieve with this bare bones file left behind. That's why I want _you_ to document it. Given everything else you've accomplished in the PR (again, fantastic work!), I have no clue what this file is for.

docs/contributing/development.md
mfisher87: I agree that this should move into a separate PR. How we think about tasks runners, philosophically, is complicated :) This page in the scientific python library development guide provides some suggestions: https://learn.scientific-python.org/development/guides/tasks/#task-runners

environment-dev.yml
mfisher87: Understandable. I will work on that during the upcoming hack day. I'll leave it up to you whether this can be merged now. I personally don't see this as an obstacle to merging at the moment, since nothing has regressed (i.e. this file was always undocumented), but if you would prefer to wait to merge I'm also cool with that! Thanks, Chuck :)

chuckwondo: I wouldn't go so far to say the file was completely undocumented, as it is referenced in the development setup as part of the required commands: https://earthaccess.readthedocs.io/en/latest/contributing/development/.

However, in this PR, it truly is completely undocumented because it is no longer referenced at all, except _erroneously_ in `docs/contributing/maintainers-guide.md`.

I won't belabor this further because I think our next hackday is a good forum to get input from others as well.

However, at a minimum, please correct the wording in `docs/contributing/maintainers-guide.md` and then I think this will be sufficient to approve and merge.

mfisher87: >I wouldn't go so far to say the file was completely undocumented, as it is referenced in the development setup as part of the required commands: https://earthaccess.readthedocs.io/en/latest/contributing/development/.
>
>However, in this PR, it truly is completely undocumented because it is no longer referenced at all, except erroneously in docs/contributing/maintainers-guide.md.

Wow, you're right. I was really trusting my memory there and should have checked my assumptions first! My apologies.

itcarroll (approved):
Fully approve, with a couple minor comments.

I haven't yet groked when the uv lock file is created or what knows to use it (i would probably have to dive into `nox` or `uv` or maybe `hatchling` docs to understand), but the great thing is I don't have to know because it's on autopilot now.

environment-dev.yml
itcarroll: My first question about this PR was why `environment-dev.yml` still existed, until I read this. I understand the hesitation to do too much in one PR, but I fully support removing this. It made sense as part of a CI pipeline, but doesn't seem useful as ... what, a handy shortcut for contributors? It's not much of a shortcut and it introduces confusion about the expected development tools. So I'll join the consensus to remove.

Makefile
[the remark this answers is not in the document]
mfisher87: NICE FIND! How did I miss that!?

environment-dev.yml
mfisher87: d1b619d

@itcarroll would you mind checking out the proposed consensus process here? I'd love to hear your thoughts! https://github.com/nsidc/earthaccess/issues/761#issuecomment-2336425761

docs/contributing/development.md
mfisher87: I don't know if I agree that we should position `nox` as the "one way" of doing things. I definitely think it's valuable as a quickstart to common tasks, though. E.g.: "I don't know how to install a dev environment but I would be embarassed if I pushed my simple error message change and it didn't pass tests"

pyproject.toml
```diff
@@ -31,85 +33,91 @@ classifiers = [
...
-# for notebook integrations and documentation examples
-xarray = ">=2023.01"
-matplotlib = ">=3.3"
-h5netcdf = ">=0.11"
-pyproj = [
-  { version = ">=3.5.0", python = "<3.12" },
-  { version = ">=3.6.1", python = ">=3.12" }
+dev = [
+  "bump-my-version >=0.10.0",
+  "nox",
```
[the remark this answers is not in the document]
mfisher87: What do you think? df578ed

Makefile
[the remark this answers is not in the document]
mfisher87: 0923beb ?

chuckwondo (dismissed):

chuckwondo (approved):

jhkennedy (approved):
:tada: :tada: Looks great! :tada: :tada: 

# :shipit:

chuckwondo (approved):

mfisher87:
Thanks all for your help! :bow:

Issue #36, opened 2021-11-16
Error messages when user hasn't accepted dataset EULA can be confusing


betolink:
We've identified two behaviors that make this user experience frustrating:

* `earthaccess.download()`, when it encounters a 403 due to a user not accepting EULA. @jessnicwelch showed a way to reproduce in this thread. When this error happens, we need to display a message like: `Access to this data has been denied because it requires you to accept a End-User License Agreement (EULA). Follow this link to view and accept the EULA: https://{the rest of it}`. The URL we need to display should be contained in the 403 response data. *This is what this ticket is about*.

* `earthaccess.download()` will continue after an error. *New ticket for this: #581*

Original description:
---

If we get a 302 redirect for a particular dataset, see if we can detect if the reason is the lack of an EULA (NASA requires users to approve an EULA for some datasets) and explain what happened.

Currently, users can receive a confusing message:

```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: ...
```

asteiker:
The UMM-C (CMR Collection) schema was updated to include EULA information, so this can now be determined programmatically by a client in order to provide better error handling: https://bugs.earthdata.nasa.gov/browse/SDRT-1236. I don't have an example handy but it looks like this is now included as a EULA identifiers list under UseConstraints (https://bugs.earthdata.nasa.gov/browse/ECSE-1213).

asteiker:
@mfisher87 I believe there is indeed data in the 403 response that tells a client whether or not the EULA was accepted. The EULA response was utilized in the Harmony API for this exact use case. Does this Harmony test help? Can we leverage something similar for earthaccess? https://github.com/nasa/harmony/blob/main/services/harmony/test/eula-acceptance.ts

jessnicwelch:
Howdy! I recreated the problem using a fresh EDL account without accepting EULAs

```
>>> import earthaccess
>>> earthaccess.login()

Enter your Earthdata Login username: jessicanwelch
Enter your Earthdata password: 
<earthaccess.auth.Auth object at 0x000002BE59952C40>

>>> results = earthaccess.search_data(
...     short_name='SEA_SURFACE_HEIGHT_ALT_GRIDS_L4_2SATS_5DAY_6THDEG_V_JPL2205',
...     cloud_hosted=True,
...     bounding_box=(-10, 20, 10, 50),
...     temporal=("1999-02", "2019-03"),
...     count=10
... )
Granules found: 1467

[...]

PR #749, opened 2024-07-06
Refactor integration tests to remove random collection sampling

### LINKED ISSUE #215: Test top 50 datasets
Right now the integration tests use randomly sampled datasets from different DAACs, this results in intermittent failures due EULAs/unavailability etc. A probably better approach would be to test only the most popular datasets (+IS2 data). A list of the most popular datasets from NASA is attached. 
[collection_short_names.txt](https://github.com/nsidc/earthaccess/files/10949414/collection_short_names.txt)

mfisher87:
Resolves #215 

Replaces random collection sampling with hardcoded lists of 100 top collections per provider in popularity order, with script to regenerate the lists as needed. Instead of sampling `n` random collections we select `n` most popular.

There's still a clear need for refactoring of the 4 cloud/onprem download/open test modules. They share a lot of code that can be fixturized. I don't want this PR to grow larger than it already is, so IMO that should  be a follow-up activity.


tests/integration/popular_collections/generate.py
```diff
@@ -0,0 +1,41 @@
...
+            "provider": provider,
+            "has_granules_or_cwic": True,
+            "include_facets": "v2",
+            "include_granule_counts": True,
+            "include_has_granules": True,
+            "include_tags": "edsc.*,opensearch.granule.osdd",
+            "page_num": 1,
+            "page_size": 100,
+            "sort_key[]": "-usage_score",
+        },
```
mfisher87, line 23: I just copied this out of my network tab after doing a similar query in earthdata search client. I'm sure we can run an equivalent query with earthaccess.

mfisher87:
Worked on this with @itcarroll during hack day. Notes: https://github.com/nsidc/earthaccess/discussions/755

mfisher87:
We considered the usefulness of random sampling tests. We don't think we should be doing this for integration tests, especially when they execute on every PR. We could, for example, run them on a cron job and create reports, but that seems like overkill when we have a community to help us identify datasets and connect with the right support channel if there's an issue with the provider.

We may still consider a cron job for, for examle, recalculating the most popular datasets on a monthly basis.

mfisher87:
We decided we can hardcode a small number and expand the list as we go. Other things like random tests on a cron or updating the list of popular datasets on a cron can be addressed separately.

mfisher87:
We will update the .txt files to .csv files and add boolean field for "does the collection have a EULA?" and then we'll use that field to mark those tests as `xfail`.

mfisher87:
Two major milestones:

1) @danielfromearth updated the script which generates the top collection lists to use all providers supported by earthaccess :tada: Still TODO: Make them CSVs with a boolean representing whether the collection has a EULA 
2) I just got the `test_onprem_download.py` module working without randomization! :tada: Still TODO: Refactor the other 3 integration test modules to share this behavior. Let's try and remove duplicate code while we're at it!

earthaccess/results.py
```diff
@@ -244,6 +244,9 @@ def _repr_html_(self) -> str:
         granule_html_repr = _repr_granule_html(self)
         return granule_html_repr
 
+    def __hash__(self) -> int:
+        return hash(self["meta"]["concept-id"])
```
mfisher87, line 248: @betolink @chuckwondo This seems reasonable, but please validate me :)

mfisher87:
Also still TODO: Run generate.py in GHA on a monthly/quarterly cron and auto-open a PR with the changes to top collections?

earthaccess/results.py
mfisher87: Thinking about it for like 5 minutes, this is obviously a bad idea. This class is subclassing `dict`. We'd need to implement like a `frozendict`.

mfisher87:
If we want to determine whether a collection has a EULA, this example was provided:

```
curl -i -XGET "https://cmr.earthdata.nasa.gov/search/collections.json?concept_id=C1808440897-ASF&pretty=true"
```

The metadata `"eula_identifiers" : [ "1b454cfb-c298-4072-ae3c-3c133ce810c8" ]` is present in the response. We're not 100% sure whether this can be used authoritatively. Discussion in progress: https://nsidc.slack.com/archives/C2LRKMDEV/p1724179804149239

tests/integration/test_onprem_open.py
mfisher87, line 1: TODO: Add tests for OBDAAC on-prem open. Related to #828 - we want to make sure the data streams successfully. Opening data from OBDAAC on-prem relies on both #828 _and_ a (potentially) unreleased change to fsspec! (check the September release notes)

danielfromearth:
Looks like part of this issue may be related to work on EULAs in [this issue](https://github.com/nsidc/earthaccess/issues/36).

mfisher87:
For some reason, running

```
nox -s integration-tests -- --cov=earthaccess --cov-report=term-missing --capture=no --tb=native --log-cli-level=INFO
```

is failing, when

```
nox -s integration-tests
```

is passing.

mfisher87:
Looks like this was a coincidence with intermittent uptime on an external dependency.

tests/integration/popular_collections/POCLOUD.txt
```diff
@@ -0,0 +1,100 @@
+C2799438299-POCLOUD
+C1996881146-POCLOUD
+# C2204129664-POCLOUD
```
mfisher87, line 3: This collection isn't working so good :face_with_thermometer: 

We get 0 granules:

```
>           assert len(granules) > 0, msg
E           AssertionError: AssertionError for C2204129664-POCLOUD
E           assert 0 > 0
E            +  where 0 = len([])
```

tests/integration/test_kerchunk.py
```diff
@@ -6,6 +6,7 @@
 from fsspec.core import strip_protocol
 
 logger = logging.getLogger(__name__)
+pytestmark = pytest.mark.skip(reason="Tests are broken.")
```
mfisher87, line 9: These tests are failing on the release. Maybe xfail is a better mark. I preferred not to get into fixing this in this PR.

mkdocs.yml
mfisher87, line 1: I felt this needed simplification as I added more.

- We have a few "guide" things, so I gave them a naming pattern so they can be mentally grouped
- Removed the word "Our" because it wasn't adding anything
- "Naming conventions" felt out of place, too specific. Like the new integration test doc. So I created a new "Topics" subsection (but not a subdirectory to keep the URL flatter). I don't like "Topics", but it's the best I have thought of so far.

betolink:
Wow, this is a big one! I can start today but I'm not sure if I can finish today! great work @mfisher87 !!

mfisher87:
Thanks for taking a look, @betolink ! There are some opportunities for refactoring, but I really tried to keep the scope narrow in this PR to avoid growing even bigger :)

tests/integration/test_kerchunk.py
betolink: fixed the tests by pinning zarr to 2.x, but we have to merge main into the PR.

betolink (commented):
I looked at the PR again and although we haven't set on "should we test all these" I think this is a great improvement on reproducibility and we can pass the info to other DAACs on relevant datasets that for some reason fail. I think we need to merge main, regenerate the uv lock and this should be good to go IMO! great work @mfisher87!!

tests/integration/test_cloud_download.py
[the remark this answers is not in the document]
mfisher87: What do you mean?

tests/integration/test_kerchunk.py
mfisher87: Merged and removed mark!

betolink (dismissed):
This is good to me, waiting for Chuck and/or Joe's feedback.

tests/integration/popular_collections/generate.py
```diff
@@ -0,0 +1,61 @@
...
+    return collection_ids
+
+
+def main():
+    for provider in all_providers:
+        collection_ids = top_collections(provider=provider)
+
+        output = THIS_DIR / f"{provider}.txt"
+        with output.open("w") as f:
+            f.write("\n".join(collection_ids))
```
jhkennedy, line 57:
```suggestion
        output.write_text("\n".join(collection_ids))
```
Since output is a pathlib Path, we can use this method directly.

docs/contributing/integration-tests.md
```diff
@@ -0,0 +1,16 @@
+# Integration tests
+
+## Testing most popular datasets
+
+Some integration tests operate on the most popular collections for each provider in CMR.
+Those collections are cached as static data in `tests/integration/popular_collections/`
```
jhkennedy, line 6:
```suggestion
Those collection IDs are cached as static data in `tests/integration/popular_collections/`
```

I read this like we cached all the data, not just the names of the collections we care about.

jhkennedy (approved):
This looks good to me -- I left two tiny comments that you can take or leave. Approval stands either way.

tests/integration/popular_collections/generate.py
mfisher87: Great call :)

docs/contributing/integration-tests.md
mfisher87: :100:

jhkennedy (approved):

mfisher87:
Oh no! The tests are failing :rofl: 

@betolink NSIDC doing maintenance?

```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='n5eil01u.ecs.nsidc.org', port=443): Max retries exceeded with url: /DP5/ATLAS/ATL08.006/2018.10.14/ATL08_20181014070058_02390107_006_02.h5 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fbb54137850>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

PR #750, opened 2024-07-06
Adopt nox

### LINKED ISSUE #735: Consider replacing scripts with Nox
https://nox.thea.codes/en/stable/

Nox enables us to couple the behaviors in the current test/integration-test/lint/format/etc. scripts with the environments those behaviors depend on, and also manages those extra environments for us. Nox could be a dev dependency, and then developers only ever need to install the dev environment; pre-commit and Nox would manage all the other environments that are needed.

The [Scientific Python Development guide provides some reasons to use Nox](https://learn.scientific-python.org/development/guides/tasks), or a task-runner in general.

mfisher87:
Depends on #733 

Resolves #735 

TODO: Replace all the places where the scripts are called with calls to Nox


mfisher87:
Possibly superceded by #733

PR #752, opened 2024-07-08
Test updation

Sherwin-14:
I've added the fixtures but it seems like a lot more errors are pop up.


mfisher87:
Converted to draft until we can figure out new errors

mfisher87:
```
ERROR tests/integration/test_cloud_download.py::test_earthaccess_can_download_cloud_collection_granules[daac0] - assert False is True
 +  where False = <earthaccess.auth.Auth object at 0x7f878cba0310>.authenticated
```

Looks like our session isn't authenticated as expected. Perhaps we could debug tomorrow during the hack day? I know that's not a very accessible time of day in your timezone, so if you aren't able to make it, I'll plan on spending that time investigating and share back here :)

chuckwondo:
@Sherwin-14, is this PR still relevant?

Sherwin-14:
> @Sherwin-14, is this PR still relevant?

I think it is. I think this stalled for a reason that I am not able to recall at the moment, @mfisher87 might have more context about this.

chuckwondo:
@Sherwin-14, please merge main into your branch and resolve conflicts, and then we can review it.

Sherwin-14:
> @Sherwin-14, please merge main into your branch and resolve conflicts, and then we can review it.

This PR needs to be rewritten imho, because of the amount of conflicts that have been accumulated. I don't feel like I will be able to resolve all of that for now, so I guess we can work on this PR later. Once I get a hang of the context, then I can rewrite this PR again using PR diff as a guide.

Issue #581, opened 2024-05-21
`earthaccess.download()` ignores errors


mfisher87:
We found the `pqdm` library is the thing swallowing all the error messages. `pqdm` has an `exception_behavior` option, which defaults to `"ignore"`, surprisingly. The other two settings, `"deferred"` and `"immediate"`, are both good options for _earthaccess_.

@chuckwondo and I agree `"immediate"` is the most intuitive default, and we can establish a new flag `fail_fast: bool = True` to override it to `"deferred"` by passing `download(..., fail_fast=False)`. Prior discussion in #36. Please continue discussion here :)

I feel exposing the `"ignore"` behavior isn't necessary because the user can do their own exception handling with `try`.

Thoughts?

Sherwin-14:
@mfisher87 This seems interesting, what should we proceed  `immediate`or `deferred`?

mfisher87:
@Sherwin-14 don't forget to take a look at #36 for more context! Do you want to be assigned this issue?

Sherwin-14:
```py

def get(
        self,
        granules: Union[List[DataGranule], List[str]],
        local_path: Union[Path, str, None] = None,
        provider: Optional[str] = None,
        threads: int = 8,
    ) -> List[str]:

        if local_path is None:
            today = datetime.datetime.today().strftime("%Y-%m-%d")
            uuid = uuid4().hex[:6]
            local_path = Path.cwd() / "data" / f"{today}-{uuid}"
        elif isinstance(local_path, str):
            local_path = Path(local_path)

        if len(granules):
            files = self._get(granules, local_path, provider, threads)
            return files
        else:
            raise ValueError("List of URLs or DataGranule instances expected")

 @singledispatchmethod
 def _get(
        self,
        granules: Union[List[DataGranule], List[str]],
        local_path: Path,
        provider: Optional[str] = None,
        threads: int = 8,
    ) -> List[str]:

        raise NotImplementedError(f"Cannot _get {granules}")

```
@mfisher87 @chuckwondo I had been looking into this and it seems like `pqdm` isn't used in the `get` and `_get` functions which are used in `download` function. But as chuck mentioned in #36 the 'pretty progress bars" are a result of `pqdm` which I too had when I  tried to recreate the error as described in the linked post above.

chuckwondo:
@Sherwin-14, just search the code base for `pqdm`.

Whichever functions you find it in, do the following:

[...]

Issue #788, opened 2024-08-16
Lint and format Jupyter Notebooks


mfisher87:
Ruff 0.6.0 is out and lints Jupyter notebooks by default: https://astral.sh/blog/ruff-v0.6.0#jupyter-notebooks-are-now-linted-and-formatted-by-default

To prevent errors on PRs which don't introduce any new lint/formatting issues, we should either lint & format the notebooks or disable for Jupyter notebooks following the info in the link above.

mfisher87:
#793 disables linting. Updating this issue to actually lint the notebooks.

PR #792, opened 2024-08-24
Added parameters and exception behaviour to pqdm

Sherwin-14:
I did all the steps outlined in the issue #581, we can now build upon this.

----
📚 Documentation preview 📚: https://earthaccess--792.org.readthedocs.build/en/792/

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -3,7 +3,7 @@ import logging
 import requests
 import s3fs
 from fsspec import AbstractFileSystem
-from typing_extensions import Any, Dict, List, Optional, Union, deprecated, Mapping
+from typing_extensions import Any, Dict, List, Mapping, Optional, Union, deprecated
 
 import earthaccess
 from earthaccess.services import DataServices
@@ -238,7 +238,7 @@ def download(
     try:
         results = earthaccess.__store__.get(
             granules, local_path, provider, threads, pqdm_kwargs
-        )    
+        )
     except AttributeError as err:
         logger.error(
             f"{err}: You must call earthaccess.login() before you can download data"
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -3,7 +3,7 @@ import logging
 import requests
 import s3fs
 from fsspec import AbstractFileSystem
-from typing_extensions import Any, Dict, List, Optional, Union, deprecated
+from typing_extensions import Any, Dict, List, Optional, Union, deprecated, Mapping
 
 import earthaccess
 from earthaccess.services import DataServices
@@ -205,7 +205,7 @@ def download(
     local_path: Optional[str],
     provider: Optional[str] = None,
     threads: int = 8,
-    fail_fast: bool = True,
+    pqdm_kwargs: Optional[Mapping[str, Any]] = None,
 ) -> List[str]:
     """Retrieves data granules from a remote storage system.
 
@@ -226,14 +226,19 @@ def download(
         Exception: A file download failed.
     """
     provider = _normalize_location(provider)
+    pqdm_kwargs = {
+        "exception_behavior": "immediate",
+        "n_jobs": threads,
+        **(pqdm_kwargs or {}),
+    }
     if isinstance(granules, DataGranule):
         granules = [granules]
     elif isinstance(granules, str):
         granules = [granules]
     try:
         results = earthaccess.__store__.get(
-            granules, local_path, provider, threads, fail_fast=fail_fast
-        )
+            granules, local_path, provider, threads, pqdm_kwargs
+        )    
     except AttributeError as err:
         logger.error(
             f"{err}: You must call earthaccess.login() before you can download data"
@@ -245,7 +250,7 @@ def download(
 def open(
     granules: Union[List[str], List[DataGranule]],
     provider: Optional[str] = None,
-    fail_fast: bool = True,
+    pqdm_kwargs: Optional[Mapping[str, Any]] = None,
 ) -> List[AbstractFileSystem]:
     """Returns a list of file-like objects that can be used to access files
     hosted on S3 or HTTPS by third party libraries like xarray.
@@ -259,8 +264,13 @@ def open(
         A list of "file pointers" to remote (i.e. s3 or https) files.
     """
     provider = _normalize_location(provider)
+    pqdm_kwargs = {
+        "exception_behavior": "immediate",
+        "n_jobs": threads,
+        **(pqdm_kwargs or {}),
+    }
     results = earthaccess.__store__.open(
-        granules=granules, provider=provider, fail_fast=fail_fast
+        granules=granules, provider=provider, pqdm_kwargs=pqdm_kwargs
     )
     return results
 
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -205,6 +205,7 @@ def download(
     local_path: Optional[str],
     provider: Optional[str] = None,
     threads: int = 8,
+    fail_fast: bool = True,
 ) -> List[str]:
     """Retrieves data granules from a remote storage system.
 
@@ -230,7 +231,9 @@ def download(
     elif isinstance(granules, str):
         granules = [granules]
     try:
-        results = earthaccess.__store__.get(granules, local_path, provider, threads)
+        results = earthaccess.__store__.get(
+            granules, local_path, provider, threads, fail_fast=fail_fast
+        )
     except AttributeError as err:
         logger.error(
             f"{err}: You must call earthaccess.login() before you can download data"
@@ -242,6 +245,7 @@ def download(
 def open(
     granules: Union[List[str], List[DataGranule]],
     provider: Optional[str] = None,
+    fail_fast: bool = True,
 ) -> List[AbstractFileSystem]:
     """Returns a list of file-like objects that can be used to access files
     hosted on S3 or HTTPS by third party libraries like xarray.
@@ -255,7 +259,9 @@ def open(
         A list of "file pointers" to remote (i.e. s3 or https) files.
     """
     provider = _normalize_location(provider)
-    results = earthaccess.__store__.open(granules=granules, provider=provider)
+    results = earthaccess.__store__.open(
+        granules=granules, provider=provider, fail_fast=fail_fast
+    )
     return results
 
 
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -229,7 +229,6 @@ def download(
         Exception: A file download failed.
     """
     provider = _normalize_location(provider)
-    pqdm_kwargs = dict(pqdm_kwargs) if pqdm_kwargs is not None else {}
     pqdm_kwargs = {
         "exception_behavior": "immediate",
         "n_jobs": threads,
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -218,6 +218,9 @@ def download(
         local_path: local directory to store the remote data granules
         provider: if we download a list of URLs, we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
+        pqdm_kwargs: Additional keyword arguments to pass to pqdm, a parallel processing library.
+            See pqdm documentation for available options. Default is to use immediate exception behavior
+            and the number of jobs specified by the `threads` parameter.
 
     Returns:
         List of downloaded files
@@ -226,6 +229,7 @@ def download(
         Exception: A file download failed.
     """
     provider = _normalize_location(provider)
+    pqdm_kwargs = dict(pqdm_kwargs) if pqdm_kwargs is not None else {}
     pqdm_kwargs = {
         "exception_behavior": "immediate",
         "n_jobs": threads,
@@ -259,6 +263,9 @@ def open(
         granules: a list of granule instances **or** list of URLs, e.g. `s3://some-granule`.
             If a list of URLs is passed, we need to specify the data provider.
         provider: e.g. POCLOUD, NSIDC_CPRD, etc.
+        pqdm_kwargs: Additional keyword arguments to pass to pqdm, a parallel processing library.
+            See pqdm documentation for available options. Default is to use immediate exception behavior
+            and the number of jobs specified by the `threads` parameter.
 
     Returns:
         A list of "file pointers" to remote (i.e. s3 or https) files.
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -266,7 +266,6 @@ def open(
     provider = _normalize_location(provider)
     pqdm_kwargs = {
         "exception_behavior": "immediate",
-        "n_jobs": threads,
         **(pqdm_kwargs or {}),
     }
     results = earthaccess.__store__.open(
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -69,7 +69,7 @@ def _open_files(
         url, granule = data
         return EarthAccessFile(fs.open(url), granule)  # type: ignore
 
-    fileset = pqdm(url_mapping.items(), multi_thread_open, n_jobs=threads)
+    fileset = pqdm(url_mapping.items(), multi_thread_open, n_jobs=threads,exception_behaviour=exception_behavior)
     return fileset
 
 
@@ -447,6 +447,7 @@ class Store(object):
                             url_mapping,
                             fs=s3_fs,
                             threads=threads,
+                            fail_fast = fail_fast
                         )
                     except Exception as e:
                         raise RuntimeError(
@@ -544,6 +545,7 @@ class Store(object):
         local_path: Path,
         provider: Optional[str] = None,
         threads: int = 8,
+        fail_fast: bool = True
     ) -> List[str]:
         data_links = granules
         downloaded_files: List = []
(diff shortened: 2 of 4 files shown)


chuckwondo:
@mfisher87, I see lots of Ruff error messages regarding a number of files that @Sherwin-14 has not touched in this PR. Is there something already on `main` to address these errors, such that Sherwin simply needs to merge `main` into his own branch to silence them?

mfisher87:
Not yet, but that should be an easy fix. I'll take care of it now :) https://github.com/nsidc/earthaccess/issues/788

mfisher87:
https://github.com/nsidc/earthaccess/pull/793

chuckwondo:
> Not yet, but that should be an easy fix. I'll take care of it now :) #788
> #793

Excellent! Thank you for the quick fix.

@Sherwin-14, Matt's fix is now merged into `main`, so please update your branch with the fix from `main`, which will eliminate the ruff errors in your build.

chuckwondo:
@Sherwin-14, do you have enough information to proceed with the changes I requested?

Sherwin-14:
> @Sherwin-14, do you have enough information to proceed with the changes I requested?

Yeah I have the info I needed, I had put this on hold since I was preoccupied with some other things for the last couple of days. I'll resume my work on this asap.

betolink:
Would it be possible and in scope of this PR to use `**kwargs` and pass them to `pqdm` for both `open()` and `download()` with sane defaults? Users could be in control of the failing behavior but also the number of threads to use and the verbosity (don't display the progress bar). I'm not sure about the style, in Xarray they use `**kwargs` and an equivalent `backend_kwargs={}` that in our case could be something like `pqdm_opts={}`

```python
# code in api.py and something similar in store.py
# in this example we leave threads out to ensure backwards compatibility and will use it in the store class if present
def open(granules, treads, provider, pqdm_opts):
    default_pqdm_opts = {
      n_jobs=threads,
      exception_behaviour='immediate'
      colour="green",
      disable=False # True means no output including the progress bar.
    }

    if not pqdm_opts;
        pqdm_opts = default_pqdm_opts
    earthaccess.__store__.open(granules, treads, pqdm_opts)

```

and actually we should do the same with `fsspec` so we let users tune their caching behavior when opening the files, aka a **"smart_open"**. What do you all think?

mfisher87:
> I had put this on hold since I was preoccupied with some other things for the last couple of days. I'll resume my work on this asap.

@Sherwin-14 No rush! We really appreciate any time you can spare, and don't want to ask for more than that :)

mfisher87:
@betolink I really like that idea. Basically "Allow customizing all pqdm behaviors" (and separately, "Allow customizing all fsspec behaviors")?

I'm not sure how I feel about including that as part of this PR. I lean towards separate out of respect for @Sherwin-14 's time, but I don't want to speak for you, Sherwin!

Sherwin-14:
> We decided during hack day today that we want to allow users to customize pqdm behaviors like so:
> 
> ```python
> def open(something, threads=16, *, pqdm_kwargs={...}, ...):
>     pqdm_kwargs_defaults = {"exception_behavior": "immediate", "threads": threads}
>     pqdm_kwargs = pqdm_kwargs_defaults.update(pqdm_kwargs)
>     ...
> ```
> 
> We considered the `threads` argument as fundamental to the function of earthaccess, so we think it should remain a top-level parameter. We'll use threads whether we use pqdm or not in 5 years.

How should we now proceed with the changes we discussed in the hackday?

chuckwondo:
> > We decided during hack day today that we want to allow users to customize pqdm behaviors like so:
> > ```python
> > def open(something, threads=16, *, pqdm_kwargs={...}, ...):
> >     pqdm_kwargs_defaults = {"exception_behavior": "immediate", "threads": threads}
> >     pqdm_kwargs = pqdm_kwargs_defaults.update(pqdm_kwargs)
> >     ...
> > ```
> >   
> > We considered the `threads` argument as fundamental to the function of earthaccess, so we think it should remain a top-level parameter. We'll use threads whether we use pqdm or not in 5 years.
> 
> How should we now proceed with the changes we discussed during the hackday?

Do the same thing you did in the PR you already started for this, but instead of `fail_fast` (which is what you pretty much completed) change it to `pqdm_kwargs` and adjust your code changes similar to what is shown above, although the code above won't work because `pqdm` has an `n_jobs` arg, not a `threads` arg, so I suggest something like so, instead, within the functions/methods that invoke `pqdm`:

```python
pqdm_kwargs = {
    "exception_behavior": "immediate",
    "n_jobs": threads,
    **pqdm_kwargs,
}

...

pqdm(<other args>, **pqdm_kwargs)
```

mfisher87:
Great point! Much clearer with the double splat too ;)

chuckwondo:
@Sherwin-14, do you need some help with this?

Sherwin-14:
> @Sherwin-14, do you need some help with this?

I will try to accomodate the change asap, I was caught up in something else up untill now

chuckwondo:
> > @Sherwin-14, do you need some help with this?
> 
> I will try to accomodate the change asap, I was caught up in something else up untill now

No rush. I just wanted to check in to see if you need any help or if something else was requiring you to postpone your continuation.

Sherwin-14:
@chuckwondo 

``` python
def download(
    granules: Union[DataGranule, List[DataGranule], str, List[str]],
    local_path: Optional[str],
    threads: int = 8,
    pqdm_kwargs: dict = {}
) -> List[str]:
    """Retrieves data granules from a remote storage system.

       * If we run this in the cloud, we will be using S3 to move data to `local_path`.
       * If we run it outside AWS (us-west-2 region) and the dataset is cloud hosted,
            we'll use HTTP links.

    Parameters:
        granules: a granule, list of granules, a granule link (HTTP), or a list of granule links (HTTP)
        local_path: local directory to store the remote data granules
        provider: if we download a list of URLs, we need to specify the provider.

    Returns:
        List of downloaded files

    Raises:
        Exception: A file download failed.
    """
    provider = _normalize_location(provider)
    pqdm_kwargs = {
        "exception_behavior": "immediate",
        "n_jobs": threads,
        **pqdm_kwargs,
    }
    if isinstance(granules, DataGranule):
        granules = [granules]
    elif isinstance(granules, str):
        granules = [granules]
    try:
        results = earthaccess.__store__.get(
            granules, local_path, provider, threads, pqdm_kwargs
        )    
    except AttributeError as err:
        logger.error(
            f"{err}: You must call earthaccess.login() before you can download data"
        )
        return []
    return results

```

Just for clarification, the public API function download would look something like this. The `pqdm_kwargs` would be taken in at the user interface and then passed on to the lower level functions in the entire call chain isn't it? or am I missing something?

chuckwondo:
Correct, but you _almost never_ want to use mutable values for default values, and this is one of those cases (among other references, see <https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments>).

So, do NOT do this:

```python
def download(
    granules: Union[DataGranule, List[DataGranule], str, List[str]],
    local_path: Optional[str],
    threads: int = 8,
    pqdm_kwargs: dict = {}
) -> List[str]:
    pqdm_kwargs = {
        "exception_behavior": "immediate",
        "n_jobs": threads,
        **pqdm_kwargs,
    }
```

Instead, DO THIS:

```python
def download(
    granules: Union[DataGranule, List[DataGranule], str, List[str]],
    local_path: Optional[str],
    threads: int = 8,
    pqdm_kwargs: Optional[Mapping[str, Any]] = None,
) -> List[str]:
    pqdm_kwargs = {
        "exception_behavior": "immediate",
        "n_jobs": threads,
        **(pqdm_kwargs or {}),
    }
```

earthaccess/store.py
```diff
@@ -391,14 +397,22 @@ def _open_granules(
...
+                access = "direct"
+                provider = granules[0]["meta"]["provider-id"]
+                # if the data has its own S3 credentials endpoint, we will use it
+                endpoint = self._own_s3_credentials(granules[0]["umm"]["RelatedUrls"])
+                if endpoint is not None:
+                    logger.info(f"using endpoint: {endpoint}")
+                    s3_fs = self.get_s3_filesystem(endpoint=endpoint)
+                else:
+                    logger.info(f"using provider: {provider}")
+                    s3_fs = self.get_s3_filesystem(provider=provider)
```
chuckwondo, line 409: Where did this come from?

earthaccess/api.py
```diff
@@ -205,6 +205,7 @@ def download(
     local_path: Optional[str],
     provider: Optional[str] = None,
     threads: int = 8,
+    pqdm_kwargs: Optional[Mapping[str, Any]] = None,
```
chuckwondo, line 208: Please update the doctstring appropriately.

earthaccess/api.py
```diff
@@ -242,6 +250,7 @@ def download(
 def open(
     granules: Union[List[str], List[DataGranule]],
     provider: Optional[str] = None,
+    pqdm_kwargs: Optional[Mapping[str, Any]] = None,
```
chuckwondo, line 253: Please update the docstring appropriately.

earthaccess/store.py
```diff
@@ -336,6 +339,7 @@ def open(
         self,
         granules: Union[List[str], List[DataGranule]],
         provider: Optional[str] = None,
+        pqdm_kwargs: Optional[Mapping[str, Any]] = None,
```
chuckwondo, line 342: Please update the docstring appropriately.

chuckwondo (changes requested):
Thanks @Sherwin-14. In addition to my other comments, please add at least 1 unit test. I don't think you need to test passing all possible pqdm kwargs, but perhaps just test that you can produce an error for downloading and have it raise an error (not that the default exception behavior is "immediate") instead of returning with a list containing errors.

chuckwondo (changes requested):
@Sherwin-14, I suspect you marked many suggestions as "resolved," but forgot to push the changes that resolve the suggestions.

Also, please remember to go back to the Conversations tab of your PR and, in the upper-right area of the tab, click the circular double-arrows next to a reviewer's name to request a re-review after the reviewer has requested changes. Otherwise, the reviewer doesn't know when you're finished making changes and ready for a follow-up review.

Clicking that icon will send another email indicating that you want another review.

image

chuckwondo:
@Sherwin-14, several other PRs have landed on `main`, so please also be sure to merge the changes from `main` and resolve conflicts.

Sherwin-14:
@chuckwondo Hey Chuck, I have made the appropirate changes but I haven't pushed the code yet since I was writing unittest for this one which you suggested earlier. I will re-request for a review once I feel this is ready to be reviewed

chuckwondo:
> @chuckwondo Hey Chuck, I have made the appropriate changes but I haven't pushed the code yet since I was writing unittest for this one which you suggested earlier. I will re-request for a review once I feel this is ready to be reviewed

Sounds good, but please wait to mark any suggestions as resolved until you push the corresponding code changes, otherwise it is confusing to others.

Sherwin-14:
```py
class TestEarthAccessDownload(unittest.TestCase):
    def test_download(self):
        # Call the login function
        earthaccess.login()

        # Call the search function
            short_name='ATL06',  # ATLAS/ICESat-2 L3A Land Ice Height
            bounding_box=(-10, 20, 10, 50),  # Only include files in area of interest...
            temporal=("1999-02", "2019-03"),  # ...and time period of interest.
            count=10
        )

        # Mock earthaccess.__store__
        with patch.object(earthaccess, '__store__', return_value= Mock()) as mock_store:
            with patch.object(mock_store, 'get') as mock_get:
                # Simulate a download failure
                mock_get.side_effect = Exception('Download failed')
    
                # Call the download function
                files = earthaccess.download(results, "/home/download-folder")
    
                # Assert that the download failed
                self.assertEqual(files, [])

if __name__ == '__main__':
    unittest.main()

```

This is the test function I came up with and this is giving me an error log that looks something like [this](https://0x0.st/XU3h.txt). Is this the thing we want to achieve at this point? Also if the test is correct then where should I place it should I make a completely new file for this like `test_api.py`?

chuckwondo:
Hi @Sherwin-14. Thanks for working on the unit tests for this PR.

The first thing I suggest you do is merge the latest from `main` into your branch, so you're working with the latest code.

Regarding the test you've written, it looks like your generally heading in the right direction.

I have some suggestions:

- Sure, if you want to put your new tests in a new `tests/unit/test_api.py` file, that sounds good to me.
- Do not use the unittest library. There's no need for the additional "overhead". Simply use top-level test functions and built-in `assert` statements. We have almost completely moved away from unittest in our integration tests, and we have a mix in our unit tests, but we should continue moving away from unittest (although I suspect this may be up for debate).
- Regarding your test shown above, that looks good, but since the new behavior is for the call to `download` to fail immediately upon any single download failure (i.e., pqdm `exception_behavior="immediate"`), then you will want to put your call to `download` in a context manager, as by default we expect an exception to be raised (as shown in your linked test failure output), thus we do not expect to be able to assert that we get back an empty list of results:

```
        with patch.object(earthaccess, '__store__', return_value= Mock()) as mock_store:
            with patch.object(mock_store, 'get') as mock_get:
                mock_get.side_effect = Exception('Download failed')

                with pytest.raises(Exception, match="Download failed"):
                    earthaccess.download(results, "/home/download-folder")
```

Sherwin-14:
I have taken out unittest but it seems that it is necessary to use unittest for importing the likes of Mock. Another option here would be to use the pytest mock library but that means we will need to add an additional dependency.

chuckwondo:
> I have taken out unittest but it seems that it is necessary to use unittest for importing the likes of Mock. Another option here would be to use the pytest mock library but that means we will need to add an additional dependency.

You can still use unittest Mocks, but there's no need for creating a subclass of unittest.TestCase.

tests/unit/test_api.py
```diff
@@ -0,0 +1,29 @@
+
+import pytest
+import earthaccess
+from earthaccess import api,download
+from unittest.mock import Mock
+
+def test_download(monkeypatch):
```
chuckwondo, line 7:
```suggestion
def test_download_immediate_failure(monkeypatch):
```

tests/unit/test_api.py
```diff
@@ -0,0 +1,29 @@
...
+        raise Exception('Download failed')
+
+    mock_store = Mock()
+    monkeypatch.setattr(earthaccess, '__store__', mock_store)
+    monkeypatch.setattr(mock_store, 'get', mock_get)
+
+    with pytest.raises(Exception, match="Download failed"):
+        earthaccess.download(results, "/home/download-folder")
+
+
```
chuckwondo, line 28: Please also add a 2nd test like so (I haven't tested, so you may need to adjust it).

We want to make sure that overriding the default pqdm kwargs works. We should probably try a few more things, but this should at least suffice for now:

```python
def test_download_deferred_failure(monkeypatch):
    earthaccess.login()

            short_name='ATL06',  
            bounding_box=(-10, 20, 10, 50),  
            temporal=("1999-02", "2019-03"), 
            count=10,
    )

    def mock_get(*args, **kwargs):
        raise Exception('Download failed')

    mock_store = Mock()
    monkeypatch.setattr(earthaccess, '__store__', mock_store)
    monkeypatch.setattr(mock_store, 'get', mock_get)

    results = earthaccess.download(results, "/home/download-folder",
            {"exception_behavior": "deferred"})

    assert len(results) == 10 and all(isinstance(e, Exception), results)
```

chuckwondo (changes requested):
In addition to my comments, please also make corrections to fix the pre-commit issues: https://results.pre-commit.ci/run/github/399867529/1730125212.zyU74SY_QNaQKq_PZmUrIQ

It looks like you perhaps have not installed pre-commit hooks locally, otherwise your attempt to commit your changes should have produced those same pre-commit issues locally, thus preventing your commit from completing.

Please install `pre-commit`, if you haven't already done so, an then run the following from your forked/cloned repo to install the hooks configured in the repo: `pre-commit install --install-hooks`. Then run `pre-commit run -a` to run the pre-commit checks against all the files, which should then help you fix the pre-commit errors.

chuckwondo (changes requested):
It looks like you have not correctly merged/rebased from `main`. I see changes that I should not see. For example, in CHANGELOG.md, I see a new Fixed section, which is _not_ new. It is already on `main`, so should _not_ show up as something new on your branch.

I think there are some other spots that seem out of sync too, such as in the `earthacces/store.py` method `_open_files`, where the type of the `threads` parameter seems to have been clobbered. It should be `threads: int = 8` (a recent change), instead of `threads: Optional[int] = 8`.

chuckwondo (dismissed):
Thanks @Sherwin-14.

I'm approving, but do @mfisher87, @betolink, or @jhkennedy want to take a pass over this before it gets merged?

tests/unit/test_api.py
```diff
@@ -0,0 +1,50 @@
+from unittest.mock import Mock
+
+import earthaccess
+import pytest
+
+
+def test_download(monkeypatch):
```
mfisher87, line 7:
```suggestion
def test_download_immediate_failure(monkeypatch):
```

mfisher87 (commented):
Looks amazing, @Sherwin-14 ! Thank you! A little nitpick, and let's merge this!

chuckwondo (approved):

PR #763, opened 2024-07-12
Docs for backwards compatibility

Sherwin-14:
Added the docs for backwards compatibility as per the issue descriptions. "resolves #471".

----
📚 Documentation preview 📚: https://earthaccess--763.org.readthedocs.build/en/763/

Issue #501, opened 2024-03-26
`ConnectionError` - Proxy env variables overridden?


maawoo:
I'm trying to download GEDI data on my university's HPC system. The following sample code results in a `ConnectionError`:

```python
results = earthaccess.search_data(
    short_name='GEDI02_A',
    bounding_box=(31.52,-25.08,31.64,-24.99),
    temporal=("2019-01-01", "2024-01-01"),
    count=-1
)
```
```
ConnectionError: HTTPSConnectionPool(host='cmr.earthdata.nasa.gov', port=443): Max retries exceeded with url: [/search/granules.umm_json](https://vscode-remote+ssh-002dremote-002bdraco2.vscode-resource.vscode-cdn.net/search/granules.umm_json)?short_name=GEDI02_A&bounding_box=31.52,-25.08,31.64,-24.99&temporal%5B%5D=2019-01-01T00:00:00Z,2024-01-01T00:00:00Z&page_size=0 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f69ef2d40b0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

My initial thought was that the API is not whitelisted in our HTTP/HTTPS proxies, which are set via environment variables. However, according to our sysadmin this should not be an issue. I was able to confirm this by requesting the same URL via curl:
```
>> curl "https://cmr.earthdata.nasa.gov/search/granules.umm_json?short_name=GEDI02_A&bounding_box=31.52,-25.08,31.64,-24.99&temporal%5B%5D=2019-01-01T00:00:00Z,2024-01-01T00:00:00Z&page_size=0"
{"hits":92,"took":394,"items":[]}
```

Any ideas / workarounds would be appreciated!

mfisher87:
It's really hard to say what's going on here without knowing more about your university HPC system. Based on the error, it looks like VSCode is somehow involved?

```
https://vscode-remote+ssh-002dremote-002bdraco2.vscode-resource.vscode-cdn.net/search/granules.umm_json
```

Can you provide some more detail on how VSCode is involved in your workflow? `host='cmr.earthdata.nasa.gov'` indicates that earthaccess is at least attempting to talk to the correct host, and the Requests library seems to agree!

maawoo:
[...]

PR #823, opened 2024-09-25
Set proxy from environment variables when initialize request session

Isotr0py:
**Related issue**: 
- #501 

**Motivation**
- The direct connection to Earthdata through our university's network is slow and unstable, which always causes corrupted downloads even the failed search.
- We have a proxy to download data from data providers like Earthdata. However, setting proxy for `earthaccess` is a little bit tricky refer to the example code in #501.

**Description**:
- This PR lets user initialize the request session with proxy, if they have proxy set in environment variables.

- [ ] Please review our
      before getting started.
- [ ] Ensure an issue exists representing the problem being solved in this PR.
      example `closes #1`. See
      decorators, etc.

GitHub at this point!

</details>

will help you out!

</details>

### DIFF
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -45,6 +45,12 @@ class SessionWithHeaderRedirection(requests.Session):
         super().__init__()
         self.headers.update({"User-Agent": user_agent})
 
+        proxy = {
+            "http": os.getenv("HTTP_PROXY", ""),
+            "https": os.getenv("HTTPS_PROXY", ""),
+        }
+        self.proxies.update({k: v for k, v in proxy.items() if v})
+
         if username and password:
             self.auth = (username, password)
 


chuckwondo (changes requested):
Thanks for the contribution, but we'll need to discuss whether or not this is an appropriate/safe way to address #501. As I've commented there, you might be able to address the issue by exporting the appropriate env var, as there are cases where they are automatically used, and thus no need to explicitly set them on the session. Specifically, `requests` uses `urllib`, which recognizes the proxy env vars.

For cases where the proxy env vars are _not_ used, that's because the session's `trust_env` property is set to `False`, indicating that the env should _not_ be trusted. Therefore, the code in this PR _ignores_ the `trust_env` property, which is perhaps not the most secure approach to solving this.

Isotr0py:
Thanks for informing! I didn't notice the `trust_env` is set to `False` for safety before. 

So how about adding a method like `earthaccess.set_proxy()` to expose proxy settings to user?

chuckwondo:
> Thanks for informing! I didn't notice the `trust_env` is set to `False` for safety before.

It's set to `False` in _some_ cases, which is why I suggest trying to set the proxy env vars and seeing if that works. If it does, great. If not, then you've hit a case where `trust_env` is set to `False`.
 
> So how about adding a method like `earthaccess.set_proxy()` to expose proxy settings to user?

The auth logic in this library has arguably become a bit "messy," so we'll need to make a thorough review of the code to determine a reasonable approach, which might very well be roughly as simple as your suggestion.

Isotr0py:
> It's set to False in some cases, which is why I suggest trying to set the proxy env vars and seeing if that works. If it does, great. If not, then you've hit a case where trust_env is set to False.

I have double checked the proxy env vars, and it didn't work. I guess the case is that I'm using `interactive` strategy to login in, which will create bearer tokens and blocking env vars here:
https://github.com/nsidc/earthaccess/blob/4aa122390a78280a0b1a3af67abac21e6bf280bb/earthaccess/auth.py#L200-L216

chuckwondo:
> > It's set to False in some cases, which is why I suggest trying to set the proxy env vars and seeing if that works. If it does, great. If not, then you've hit a case where trust_env is set to False.
> 
> I have double checked the proxy env vars, and it didn't work. I guess the case is that I'm using `interactive` strategy to login in, which will create bearer tokens and blocking env vars here:

Can you try creating a `~/.netrc` file with your Earthdata Login creds?

Isotr0py:
> Can you try creating a ~/.netrc file with your Earthdata Login creds?

I tried both `netrc` and `environment` strategies, and they all return `session.trust_env=False` with header contents bearer tokens. I wonder if there is an issue related to the auth logic.

**My test script**
```python3
import earthaccess

earthaccess.login(strategy="netrc")
session = earthaccess.get_requests_https_session()
print(session.headers)
print(session.trust_env)
``` 

**Outputs**
```text
{'User-Agent': 'earthaccess v0.10.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Authorization': 'Bearer <My_token>'}
False
```

chuckwondo:
> > Can you try creating a ~/.netrc file with your Earthdata Login creds?
> 
> I tried both `netrc` and `environment` strategies, and they all return `session.trust_env=False` with header contents bearer tokens. I wonder if there is an issue related to the auth logic.

Don't bother looking at that for the moment. That may be misleading. Instead, try to do the thing that currently isn't working for you. For example, if you have a script to do something with earthaccess that currently fails due to not making use of the proxy, and the script is named `foo.py`, run it like so to see if it works:

```plain
https_proxy=http://HOST:PORT python foo.py
```

where HOST and PORT are whatever is appropriate for your proxy.

I am able so run a local proxy on my laptop at http://localhost:8080 and I am able to login and download via earthaccess.

Isotr0py:
I have a simple download script:
```python3
import argparse
import earthaccess
import os


def download(date_start, date_end, shortname):
    earthaccess.login()
        short_name=shortname,
        cloud_hosted=True,
        temporal=(date_start, date_end),
    )
    session = earthaccess.get_requests_https_session()
    print("http_proxy:", os.environ["http_proxy"])
    print("https_proxy:", os.environ["https_proxy"])
    print(session.headers)
    print(session.trust_env)
    earthaccess.download(results, f"download/{shortname}")


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--shortname", type=str)
    parser.add_argument("--date-start", type=str)
    parser.add_argument("--date-end", type=str)
    args = parser.parse_args()
    download(args.date_start, args.date_end, args.shortname)
```

And I run with below command on Windows Powershell, however, the download has failed with corrupted files, while `http_proxy` and `https_proxy` did has been passed to env vars.
```powershell
$env:HTTP_PROXY="http://127.0.0.1:7890"; $env:HTTPS_PROXY="http://127.0.0.1:7890"; python .\test_scripts.py --shortname GOCI_L2_OC --date-start 2015-01-01 --date-end 2015-01-02
```
**Outputs**
```
http_proxy: http://127.0.0.1:7890
https_proxy: http://127.0.0.1:7890
{'User-Agent': 'earthaccess v0.10.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Authorization': 'Bearer <my token>'}
False

QUEUEING TASKS | : 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 16/16 [00:00<00:00, 557.52it/s]
PROCESSING TASKS | :   0%|                                                                                                                                                                         | 0/16 [00:00<?, ?it/s]Error while downloading the file G2015001001640.L2_COMS_OC.nc
Traceback (most recent call last):
  File "F:\GIS\earthaccess\.venv\Lib\site-packages\urllib3\response.py", line 748, in _error_catcher
    yield
  File "F:\GIS\earthaccess\.venv\Lib\site-packages\urllib3\response.py", line 894, in _raw_read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
urllib3.exceptions.IncompleteRead: IncompleteRead(17812224 bytes read, 185151892 more expected)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "F:\GIS\earthaccess\earthaccess\store.py", line 646, in _download_file
    shutil.copyfileobj(r.raw, f, length=1024 * 1024)
  File "D:\miniconda3\Lib\shutil.py", line 197, in copyfileobj
    buf = fsrc_read(length)
          ^^^^^^^^^^^^^^^^^
  File "F:\GIS\earthaccess\.venv\Lib\site-packages\urllib3\response.py", line 949, in read
    data = self._raw_read(amt)
           ^^^^^^^^^^^^^^^^^^^
  File "F:\GIS\earthaccess\.venv\Lib\site-packages\urllib3\response.py", line 872, in _raw_read
    with self._error_catcher():
  File "D:\miniconda3\Lib\contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "F:\GIS\earthaccess\.venv\Lib\site-packages\urllib3\response.py", line 772, in _error_catcher
    raise ProtocolError(arg, e) from e
urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(17812224 bytes read, 185151892 more expected)', IncompleteRead(17812224 bytes read, 185151892 more expected))
```

However, when I switched to the script provided in https://github.com/nsidc/earthaccess/issues/501#issuecomment-2023663133, it can download 16 complete files through proxy.

chuckwondo:
Thank you. That's all helpful information.

chuckwondo:
@Isotr0py, I believe this might suffice as a workaround for you for the moment, so please let me know if this works for you. I've tweaked your sample code from above.

The basic idea is that I'm setting the proxies on the session obtained from `Auth.get_session` (which is invoked on the call path from `earthaccess.get_requests_https_session`) _and_ caching that session, such that every subsequent call to `Auth.get_session` will use that same session. This is necessary because otherwise, every call to `Auth.get_session` creates and returns a new session object.

This might be a way around modifying earthaccess code directly, so let me know if it does the trick.

```python
import argparse
import os
from functools import cache
from typing import Callable
from typing_extensions import ParamSpec

import earthaccess
import requests


P = ParamSpec("P")


def set_proxies(f: Callable[P, requests.Session]) -> Callable[P, requests.Session]:
    def go(*args: P.args, **kwargs: P.kwargs) -> requests.Session:
        s = f(*args, **kwargs)
        s.proxies.update(
            {
                k: v
                for k in ("HTTP_PROXY", "HTTPS_PROXY")
                if (v := os.environ.get(k, os.environ.get(k.casefold())))
            }
        )

        return s

    return go


def download(date_start, date_end, shortname):
    earthaccess.login()
        short_name=shortname,
        cloud_hosted=True,
        temporal=(date_start, date_end),
    )
    auth: earthaccess.Auth = earthaccess.__store__.auth
    auth.get_session = cache(set_proxies(auth.get_session))
    session = earthaccess.get_requests_https_session()
    print("http_proxy:", os.environ.get("http_proxy"))
    print("https_proxy:", os.environ.get("https_proxy"))
    print(f"{session.proxies=}")
    print(session.headers)
    print(session.trust_env)
    earthaccess.download(results, f"download/{shortname}")


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--shortname", type=str)
    parser.add_argument("--date-start", type=str)
    parser.add_argument("--date-end", type=str)
    args = parser.parse_args()
    download(args.date_start, args.date_end, args.shortname)
```

Isotr0py:
@chuckwondo The above script didn't work, because the `set_proxy` is using incorrect keys (`HTTP_PROXY` and `HTTPS_PROXY`) in proxy dict. 

In fact, this script worked well after I modified the `set_proxies` with the correct keys (`http` and `https`):
```python3
def set_proxies(f: Callable[P, requests.Session]) -> Callable[P, requests.Session]:
    def go(*args: P.args, **kwargs: P.kwargs) -> requests.Session:
        s = f(*args, **kwargs)
        # session use 'http' and `https` as proxies dict keys
        s.proxies.update(
            {
                k.split("_")[0].lower(): v
                for k in ("HTTP_PROXY", "HTTPS_PROXY")
                if (v := os.environ.get(k, os.environ.get(k.casefold())))
            }
        )

        return s

    return go
```

chuckwondo (changes requested):
@Isotr0py, the example I provided was for you to use _instead_ of this code change. I'm glad it works for you (with the small tweak you made to it).

We still do _not_ want to make this particular PR change internally because, as I mentioned earlier, this completely ignores cases where the session's `trust_env` property is set to `False`. We want to respect that setting and instead find a better way to give the caller more control over the session settings themselves.

Please close this PR, as this is a larger issue that the earthaccess team needs to discuss and find a solution for.

chuckwondo:
> @chuckwondo The above script didn't work, because the `set_proxy` is using incorrect keys (`HTTP_PROXY` and `HTTPS_PROXY`) in proxy dict.
> 
> In fact, this script worked well after I modified the `set_proxies` with the correct keys (`http` and `https`):
> 
> ```python
> def set_proxies(f: Callable[P, requests.Session]) -> Callable[P, requests.Session]:
>     def go(*args: P.args, **kwargs: P.kwargs) -> requests.Session:
>         s = f(*args, **kwargs)
>         # session use 'http' and `https` as proxies dict keys
>         s.proxies.update(
>             {
>                 k.split("_")[0].lower(): v
>                 for k in ("HTTP_PROXY", "HTTPS_PROXY")
>                 if (v := os.environ.get(k, os.environ.get(k.casefold())))
>             }
>         )
> 
>         return s
> 
>     return go
> ```
> 
> **Outputs**
> 
> ```
> http_proxy: http://127.0.0.1:7890
> https_proxy: http://127.0.0.1:7890
> session.proxies={'http': 'http://127.0.0.1:7890', 'https': 'http://127.0.0.1:7890'}
> {'User-Agent': 'earthaccess v0.10.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Authorization': 'Bearer <my token>'}
> False
> QUEUEING TASKS | : 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:00<00:00, 776.63it/s]
> PROCESSING TASKS | : 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [01:44<00:00, 13.09s/it]
> ```

Awesome! I'm glad my suggestion (with your minor tweak) worked for you!

Personally, I'd probably tweak it like so, but just a personal preference:

```python
def set_proxies(f: Callable[P, requests.Session]) -> Callable[P, requests.Session]:
    def go(*args: P.args, **kwargs: P.kwargs) -> requests.Session:
        s = f(*args, **kwargs)
        s.proxies.update(
            {
                scheme: v
                for scheme in ("http", "https")
                if (
                    v := os.environ.get(
                        k := f"{scheme}_proxy", os.environ.get(k.upper())
                    )
                )
            }
        )

        return s

    return go
```

Issue #833, opened 2024-10-01
Following the releasing instructions results in an error when pushing


mfisher87:
"protected branch hook declines"

"Changes must be made through a Pull Request"

Following the instructions as written only works if the releaser bypasses branch protections.

jhkennedy:
Note: There's a relevant discussion of automating the release process in #857

asteiker:
Hi @mfisher87 ! I saw this was mentioned in previously merged PR's. Is this bug still present and needs to be addressed?

mfisher87:
I'm not sure :) It's been some time since I tested. We should try to do our next release exactly following the doc, ideally have someone who's never released before do it without coaching. Then we'll know where our instructions are wrong/insufficient and can update them.

Overhauling the release process feels like a separate thing that we should do either way.

asteiker:
Good call! I know that @chuckwondo recently walked through our release docs for v0.18 and ran into a few issues. I believe you had submitted a small PR to update the release instructions. Do you know if perhaps you addressed this issue, or should this stay open?

chuckwondo:
I believe the instructions should be good now. If there are any rough edges remaining, they are certainly different from the cause of this old issue, so I suggest we close this as fixed, and open a new issue if there are any problems for the next person to attempt a release.

PR #834, opened 2024-10-01
Update releasing docs from run-through with Nikki

mfisher87:
We discovered an issue: #833 but chose to skip over that problem by giving Nikki temporary (already-revoked) privileges to bypass branch protection. The doc is updated assuming that one can actually push directly to `main`, but I think we should switch to doing all releases in PRs and update the doc accordingly. What does everyone think?

----
📚 Documentation preview 📚: https://earthaccess--834.org.readthedocs.build/en/834/


mfisher87:
This isn't rendering nicely in Mkdocs. Needs fix!

docs/contributing/releasing.md
```diff
@@ -3,26 +3,46 @@
 > :memo: The versioning scheme we use is [SemVer](http://semver.org/). Note that until
 > we agree we're ready for v1.0.0, we will not increment the major version.
 
-1. Ensure all desired features are merged to `main` branch and `CHANGELOG.md` is updated.
+1. Ensure all desired features are merged to `main` branch and `CHANGELOG.md` is
```
nikki-t, line 6: Should note that we needed to remove the brackets from `[Unreleased]` in order for `bumpy-my-version` to update the release for the CHANGELOG?

docs/contributing/releasing.md
```diff
@@ -3,26 +3,46 @@
...
+   updated. **Do not** edit the `[Unreleased]` header in the CHANGELOG -- the next step
+   will automatically update it.
 1. Use `bump-my-version` to increase the version number in all needed places, e.g. to
```
nikki-t, line 9: We should include the files that `bump-my-version` changes which is: CHANGELOG.md, pyproject.toml, and CITATION.cff

nikki-t (dismissed):
I left a few comments on some suggestion but I think the updated documentation on the release process look great!

docs/contributing/releasing.md
mfisher87: Good catch, the brackets shouldn't have been there in the first place :)

docs/contributing/releasing.md
```diff
@@ -3,26 +3,46 @@
 > :memo: The versioning scheme we use is [SemVer](http://semver.org/). Note that until
 > we agree we're ready for v1.0.0, we will not increment the major version.
 
-1. Ensure all desired features are merged to `main` branch and `CHANGELOG.md` is updated.
+1. Ensure all desired features are merged to `main` branch and `CHANGELOG.md` is
+   updated. **Do not** edit the `[Unreleased]` header in the CHANGELOG -- the next step
```
mfisher87, line 7:
```suggestion
   updated. **Do not** edit the `Unreleased` header in the CHANGELOG -- the next step
```

docs/contributing/releasing.md
mfisher87: I'm not sure how I feel about that. I could be convinced either way. The other side, in my mind, is then we have to keep that list in sync when we change the config. Perhaps something along the lines of "The files which will be edited by bump-my-version are configured in `pyproject.toml`"?

nikki-t: That makes sense. I like the idea of pointing folks to a list of files that will be edited (`pyproject.toml`) as it also provides some context for the `bump-my-version` command.

mfisher87: 3004088 What do you think?

mfisher87:
Fixed up the formatting in Mkdocs! 5b84afc

docs/contributing/releasing.md
nikki-t: Perfect!

nikki-t:
Looks great!

nikki-t (approved):

mfisher87:
Thanks, @nikki-t !

PR #862, opened 2024-11-01
Populate `earthaccess.login` guide in User Guide Authentications docs

### LINKED ISSUE #853: [BUG] Populate user_guide/authentication

### Current Behavior

User Guide Authentication page needs populating

### Expected Behavior

Contains complete documentation with examples for authenticating using earthaccess  

### Steps To Reproduce

See https://earthaccess.readthedocs.io/en/latest/user_guide/authenticate/

### Environment

```markdown
- OS:
- Python:
```

### Additional Context

_No response_

andypbarrett:
Populated authentication page in User Guide

This resolves #853 add (hopefully) a complete guide to using `earthaccess.login` and auth methods.

The following are included in the page:
- Descriptions of how to use `earthaccess.login` to manually login, and use netrc and environment variables.  
- Instructions to create netrc and environment variables in Linux, MacOX and Windows are included.
    -  These need to be tested, especially for MacOS and Windows.
- Instructions for authenticating different endpoints; e.g. UAT.
- Instructions for retrieving S3 credentials

Much of this information comes from the "How do I" section.  I suggest that we remove content from that section and use the "How do I..." to provide specific links to the User Guide Authentication.

----
📚 Documentation preview 📚: https://earthaccess--862.org.readthedocs.build/en/862/


docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,139 @@
...
+>>> auth = earthaccess.login()
+Enter your Earthdata Login username: your_username
+Enter your Earthdata password: 
+```
+
+You don't need to assign the result of `earthaccess.login()` to a variable but doing so enables access session information.  These are discussed in [Accessing Session Information]().
+
+Setting `strategy=interactive` will force a manual login.
+
+## Login using a `.netrc`
```
mfisher87, line 29: I think a note here might be really useful -- we don't (yet) support encrypted netrc files, and this strategy of writing credentials in plain text to disk **should not** be used on untrusted machines or shared user accounts.

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,139 @@
...
+    If you want to set these environmental variables permanently, add these two lines to your `.profile` file.
+    
+
+=== "Linux"
+    If you want to set the environment variables for the current shell session, type the following on the command line.
+    ```
+    export EARTHDATA_USERNAME="username"
+    export EARTHDATA_PASSWORD="password"
+    ```
+    If you want to set these environmental variables permanently, add these two lines to your `.bashrc` file.
```
mfisher87, line 92: `.bashrc` for Linux and `.profile` for Mac won't always be correct. I'm not very familiar with zsh yet, but I believe it uses a `.zprofile` file? Bash users can permanently set envvars in `.profile` as well, but that will only work for "login" shells. There isn't really a one-size-fits-all answer here, it depends on the shell.

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,139 @@
...
+    To set these environmental variables permanantly  
+    1. Open the start menu.  
+    2. Search for the "Advanced System Settings" control panel and click on it.  
+    3. Click on the "Environment Variables" button toward the bottom of the screen.  
+    4. Follow the prompts to add the variable to the user table.  
+
+
+## Accessing different endpoints
+
+### Earthdata User Acceptance Testing (UAT) environment
```
mfisher87, line 109:
```suggestion
### Earthdata User Acceptance Testing (UAT) endpoint
```

We use three terms in overlapping ways right now: "System", "environment", and "endpoint". I think since we use the term "system" for the API we should try and only use "system" in the docs. At the very least, we should eliminate the use of "environment"?

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,139 @@
...
+by setting the `system` argument at login, as follows:
+
+```python
+import earthaccess
+
+earthaccess.login(system=earthaccess.UAT)
+
+```
+
+## Using `earthaccess` to get credentials
```
mfisher87, line 122:
```suggestion
## Using `earthaccess` to get S3 credentials
```

docs/user_guide/authenticate.md
andypbarrett: So should we just say "add [these] to the appropriate configuration files for your operating system."

mfisher87: I think that makes sense as a fallback instruction; e.g.:

> If you use `bash` and would like to set these environment variables permanently, add these two lines to your `~/.profile` file: ... code ... . Otherwise, use the recommended method to persistently set environment variables for whichever shell you use.

Something like that?

andypbarrett: Added
```
    If you use base and would like to set these environmental variables permanently, add these two lines to your `~/.profile` file:
    ```
    EARTHDATA_USERNAME="username"
    EARTHDATA_PASSWORD="password"
    ```
    For other shells, use the recommended method to persistently set these environment variables for whichever shell you use.
```

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,157 @@
 # Authentication
 
-Introduces the `earthaccess.login` method for managing Earthdata Login and cloud credentials.
+You can use `earthaccess` to search for datasets and data without needing to log in.
+However, to access (download or stream) NASA Earth science data, whether from one of the NASA
+Distributed Active Archive Centers (DAACs) or from NASA Earthdata Cloud, you need
```
asteiker, line 5:
```suggestion
Distributed Active Archive Centers (DAACs) on-premises archive or from NASA Earthdata Cloud, you need
```

asteiker: data from NASA Earthdata Cloud still come from the DAAC. Just clarifying so that it doesn't sound like an either-or.

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,157 @@
...
+
+```
+>>> import earthaccess
+
+>>> auth = earthaccess.login()
+Enter your Earthdata Login username: your_username
+Enter your Earthdata password: 
+```
+
+You don't need to assign the result of `earthaccess.login()` to a variable but doing so enables access session information.  These are discussed in [Accessing Session Information]().
```
asteiker, line 32:
```suggestion
You don't need to assign the result of `earthaccess.login()` to a variable but doing so enables the access of session information.  These are discussed in [Accessing Session Information]().
```

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,157 @@
...
+    2. Search for the "Advanced System Settings" control panel and click on it.  
+    3. Click on the "Environment Variables" button toward the bottom of the screen.  
+    4. Follow the prompts to add the variable to the user table.  
+
+
+## Accessing different endpoints
+
+### Earthdata User Acceptance Testing (UAT) endpoint
+
+If your EDL account is authorized to access the User Acceptance Testing (UAT) system,
```
asteiker, line 129:
```suggestion
If your EDL account is authorized to access the User Acceptance Testing (UAT) environment,
```

asteiker (changes requested):
Just a few minor suggestions here. I did not manually test the different OS guidance but saw that Matt was on it!

docs/user_guide/authenticate.md
```diff
@@ -1,11 +1,139 @@
...
+    ```
+
+=== "Linux"
+    Type the following on your command line, replacing `<username>` and `<password>` with your
+    Earthdata Login credentials.
+    
+    ```
+    echo "machine urs.earthdata.nasa.gov login <username> password <password>" >> $HOME/.netrc
+    chmod 600 $HOME/.netrc
+    ```
```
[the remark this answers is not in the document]
andypbarrett: I'm torn on this suggestion.  On the one hand the approach is the same for MacOS and Linux so this avoids duplication.  But on the other hand, MacOS and Linux are/could be viewed by users and separate things.  It seems cleaner to have separate tabs.

asteiker (approved):

docs/user_guide/authenticate.md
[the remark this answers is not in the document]
mfisher87: I think it's a fairly common pattern to combine MacOS & Linux instructions where they would be the same, so I don't feel users would be thwarted by this. I also don't feel strongly! Whatever you prefer is fine with me :)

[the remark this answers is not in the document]
asteiker: @mfisher87 We are working on this synchronously in our hackday and I think we'll leave things as-is to get it merged today. Especially since we all don't feel strongly on this!

PR #866, opened 2024-11-04
Fix pqdm_kwargs integration tests

chuckwondo:
Integration tests are failing, with many errors of the following form:

```plain
FAILED tests/integration/test_cloud_open.py::test_***_can_open_onprem_collection_granules[daac1] - NotImplementedError: granules should be a list of DataGranule or URLs
```

This PR fixes this issue, although it's odd that the integration tests did not fail prior to the code landing on `main` that is now causing this problem.

It turns out to be an issue with use of singledispatch, and having registered functions with signatures that differed (accidentally, it seems) by more than simply a difference of type of the first argument.

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -1,4 +1,5 @@
 import logging
+from pathlib import Path
 
 import requests
 import s3fs
@@ -202,9 +203,10 @@ def login(strategy: str = "all", persist: bool = False, system: System = PROD) -
 
 def download(
     granules: Union[DataGranule, List[DataGranule], str, List[str]],
-    local_path: Optional[str],
+    local_path: Optional[Union[Path, str]] = None,
     provider: Optional[str] = None,
     threads: int = 8,
+    *,
     pqdm_kwargs: Optional[Mapping[str, Any]] = None,
 ) -> List[str]:
     """Retrieves data granules from a remote storage system.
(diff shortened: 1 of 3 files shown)


chuckwondo:
> Does it have something to do with our magic code that allows 10% of integration tests to fail? 😆

I don't think so. Something very odd seems to be happening, and I can't figure it out.

The code in this PR passes all unit and integration tests when I run them locally with `nox`, so I have no idea what's going on with all the CI failures.

It appears my attempt to "fix" the recent integration test failures has only produced even more failures. I'm still digging to see what's up.

mfisher87:
Want to pair on it sometime this week?

chuckwondo:
@mfisher87, I discovered one of the problems, which was that the unit tests that were added for the original pqdm_kwargs work should have been integration tests instead. The unit tests were failing because they were waiting for interactive input of creds for login. I have moved them (and tweaked them) to be integration tests, so now they are passing.

However, what's puzzling now is why are integration tests failing?

What I've discovered by looking at the most recent failure is that the integration test jobs are _not_ running the most recent code that I've pushed. I am coming to this conclusion because of the following:

- The new integration tests are not being executed (because they're not in whichever version of code is being run)
- This error shows that the reported line is _not_ the latest version of code that I've pushed:
   
    ```
    File "/opt/hostedtoolcache/Python/3.11.10/x64/lib/python3.11/site-packages/***/store.py", line 364, in open
      return self._open(granules, provider, pqdm_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ```

  That is _not_ the current code on line 364 of store.py.

Oddly, it appears that the code being run is coming from nsidc:main, not from chuckwondo:fix-pqdm-kwargs.

I might simply close this PR and open a new one to see if that fixes the problem.

PR #914, opened 2025-01-05
Add a workflow for generating Hackday Github Discussion threads

Sherwin-14:
I have replaced the CATEGORY_ID and REPOSITORY_ID, though, it remains untested. (It would be funny if Github allowed me to open a discussions thread in another repository :laughing:)

----
📚 Documentation preview 📚: https://earthaccess--914.org.readthedocs.build/en/914/

PR #917, opened 2025-01-06
Fix `CATEGORY_ID` in hack day discussion workflow

### LINKED ISSUE #801: Action to automatically create hack day discussion threads
We're currently making these manually:

https://github.com/nsidc/earthaccess/discussions/800

We should be able to set up their creation on a bi-weekly cron in GitHub Actions!

mfisher87:
See #914!

Resolves #801 

----
📚 Documentation preview 📚: https://earthaccess--917.org.readthedocs.build/en/917/


mfisher87:
Breaking the rules, already tested :)

PR #923, opened 2025-01-08
Fix small open_virtual_dataset bugs

### LINKED ISSUE #906: [BUG] incorrect typing on get_s3_filesystem has led open_virtual_mfdataset astray

### Current Behavior

The `earthaccess.open_virtual_mfdataset` function with `access="direct"` returns a `KeyError: 0`.

Here the `earthaccess.get_s3_filesystem` is given `granules[0]` when it should be given `granules`.
https://github.com/nsidc/earthaccess/blob/b9dec8bcbbe4bbdf7eadb318076ee2b2f67e2e69/earthaccess/dmrpp_zarr.py#L95

Here the typing is misleading and should be `Optional[list[DataGranule]]`.
https://github.com/nsidc/earthaccess/blob/b9dec8bcbbe4bbdf7eadb318076ee2b2f67e2e69/earthaccess/api.py#L402

### Expected Behavior

The `earthaccess.open_virtual_mfdataset` function with `access="direct"` returns a dataset.

### Steps To Reproduce

```python
import earthaccess

    short_name="MUR-JPL-L4-GLOB-v4.1",
    temporal=("2010-01-01", "2010-01-31"),
)
earthaccess.open_virtual_mfdataset(
    results,
    access="direct",
    load=False,
    concat_dim="time",
    coords="all",
    compat="override",
    combine_attrs="drop_conflicts",
)
```

### Environment

```markdown
- OS: openscapes.2i2c
- Python: 3.12.0
```

### Additional Context

_No response_

ayushnag:
- [x] Closes #906
- [x] Fixes load=True error by updating optional dependency group
- [x] Updated tests

----
📚 Documentation preview 📚: https://earthaccess--923.org.readthedocs.build/en/923/

### DIFF
--- a/earthaccess/dmrpp_zarr.py
+++ b/earthaccess/dmrpp_zarr.py
@@ -92,8 +92,8 @@ def open_virtual_mfdataset(
     import xarray as xr
 
     if access == "direct":
-        fs = earthaccess.get_s3_filesystem(results=granules[0])
-        fs.storage_options["anon"] = False  # type: ignore
+        fs = earthaccess.get_s3_filesystem(results=granules)  # type: ignore
+        fs.storage_options["anon"] = False
     else:
         fs = earthaccess.get_fsspec_https_session()
     if parallel:
@@ -114,7 +114,7 @@ def open_virtual_mfdataset(
                 filetype="dmrpp",  # type: ignore
                 group=group,
                 indexes={},
-                reader_options={"storage_options": fs.storage_options},  # type: ignore
+                reader_options={"storage_options": fs.storage_options},
             )
         )
     if preprocess is not None:
@@ -127,6 +127,7 @@ def open_virtual_mfdataset(
         vds = xr.combine_nested(vdatasets, **xr_combine_nested_kwargs)
     if load:
         refs = vds.virtualize.to_kerchunk(filepath=None, format="dict")
+        protocol = "s3" if "s3" in fs.protocol else fs.protocol
         return xr.open_dataset(
             "reference://",
             engine="zarr",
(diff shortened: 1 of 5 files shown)


ayushnag:
Note that this PR changes the dmrpp tests to no longer verify the result of `open_virtual_dataset` since that is checked in virtualizarr. The tests here now check that the expected dmrpp files are found, parsed, and can be loaded.

Mizokuiam:
Thank you for raising this issue! I'll look into it and try to help if I can.

betolink (approved):
Nice, sorry for the delay. Is this the last thing we need before adding example notebooks and releasing 0.13.0? cc @ayushnag

betolink:
oops, looks like there are some failing integration tests. e.g.
 ```
tests/integration/test_virtualizarr.py::test_open_virtual_dataset[AVHRR_OI-NCEI-L4-GLOB-v2.1] 
```
I wonder if it's related to JPL's outage. I'll debug locally.

ayushnag:
@betolink Thanks for reviewing and pushing the fixes! I believe zarr v3 doesn't break the parser, v3 is not yet compatible with kerchunk. However, kerchunk is needed for the `load=True` param so it is ok to pin to v2. However, we will have to solve this problem soon since virtualizarr will also soon require zarr v3. The alternatives are to use an icechunk in-memory store to load the data or use a Zarr store designed for reading manifests [like the one suggested here](https://github.com/zarr-developers/VirtualiZarr/issues/238#issuecomment-2585549691).

betolink:
Ahhh good to know! ok so let's get this merged and then work on the notebooks so we can have a virtualizarr release next week! @ayushnag

betolink (approved):

PR #946, opened 2025-02-10
Allow user to control behavior when login fails; warn by default

### LINKED ISSUE #888: [BUG] {{ No warning for incorrect login}}

### Current Behavior

Thanks for the work with `earthaccess`! I've been using it for a while with no issues, however, I'm setting up a new computer, and I noticed when I use 
```python
earthaccess.login()
```
and I'm prompted for my username and password, if I enter an incorrect value, I don't receive any error messages. I would expect an error for logging in failed. This seems to happen for both `interactive` and `environment` strategies, I haven't tried with the `netrc` strategy. 

### Expected Behavior

When logging in fails in `earthaccess.login()`, an error message is raised telling the users the login failed.

### Steps To Reproduce

```python
import earthaccess

earthaccess.login()
```
-> manually type in username and password

### Environment

```markdown
- OS: Linux (Fedora)
- Python: 3.12.7

- OS: Windows 10
- Python: 3.11.11
```

### Additional Context

_No response_

mfisher87:
Better distinguish between a failed login attempt and a skipped strategy.

Resolves #888 

- [ ] Please review our
      before getting started.
      example `closes #1`. See
      decorators, etc.

GitHub at this point!

</details>

will help you out!

</details>

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -4,9 +4,19 @@ from pathlib import Path
 import requests
 import s3fs
 from fsspec import AbstractFileSystem
-from typing_extensions import Any, Dict, List, Mapping, Optional, Union, deprecated
+from typing_extensions import (
+    Any,
+    Dict,
+    List,
+    Literal,
+    Mapping,
+    Optional,
+    Union,
+    deprecated,
+)
 
 import earthaccess
+from earthaccess.exceptions import LoginStrategyUnavailable
 from earthaccess.services import DataServices
 
 from .auth import Auth
@@ -161,7 +171,12 @@ def search_services(count: int = -1, **kwargs: Any) -> List[Any]:
     return query.get(hits if count < 1 else min(count, hits))
 
 
-def login(strategy: str = "all", persist: bool = False, system: System = PROD) -> Auth:
+def login(
+    strategy: str = "all",
+    persist: bool = False,
+    system: System = PROD,
+    on_failure: Literal["warn", "error"] = "warn",
+) -> Auth:
     """Authenticate with Earthdata login (https://urs.earthdata.nasa.gov/).
 
     Parameters:
@@ -174,6 +189,7 @@ def login(strategy: str = "all", persist: bool = False, system: System = PROD) -
             * **"environment"**: retrieve username and password from `$EARTHDATA_USERNAME` and `$EARTHDATA_PASSWORD`.
         persist: will persist credentials in a .netrc file
         system: the Earthdata system to access, defaults to PROD
+        on_failure: Whether to print a warning or raise an error when login fails.
 
     Returns:
         An instance of Auth.
@@ -186,16 +202,25 @@ def login(strategy: str = "all", persist: bool = False, system: System = PROD) -
         for strategy in ["environment", "netrc", "interactive"]:
             try:
                 earthaccess.__auth__.login(
-                    strategy=strategy, persist=persist, system=system
+                    strategy=strategy,
+                    persist=persist,
+                    system=system,
+                    on_failure=on_failure,
                 )
-            except Exception:
-                pass
+            except LoginStrategyUnavailable as err:
+                logger.debug(err)
+                continue
 
             if earthaccess.__auth__.authenticated:
                 earthaccess.__store__ = Store(earthaccess.__auth__)
                 break
     else:
-        earthaccess.__auth__.login(strategy=strategy, persist=persist, system=system)
+        earthaccess.__auth__.login(
+            strategy=strategy,
+            persist=persist,
+            system=system,
+            on_failure=on_failure,
+        )
         if earthaccess.__auth__.authenticated:
             earthaccess.__store__ = Store(earthaccess.__auth__)
 
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -6,15 +6,16 @@ import platform
 import shutil
 from netrc import NetrcParseError
 from pathlib import Path
-from typing import Any, Dict, List, Mapping, Optional
+from typing import Any, Dict, List, Literal, Mapping, Optional
 from urllib.parse import urlparse
 
 import requests  # type: ignore
 from tinynetrc import Netrc
 from typing_extensions import deprecated
 
-from .daac import DAACS
-from .system import PROD, System
+from earthaccess.daac import DAACS
+from earthaccess.exceptions import LoginAttemptFailure, LoginStrategyUnavailable
+from earthaccess.system import PROD, System
 
 try:
     user_agent = f"earthaccess v{importlib.metadata.version('earthaccess')}"
@@ -99,6 +100,8 @@ class Auth(object):
         strategy: str = "netrc",
         persist: bool = False,
         system: Optional[System] = None,
+        *,
+        on_failure: Literal["warn", "error"],
     ) -> Any:
         """Authenticate with Earthdata login.
 
@@ -112,6 +115,7 @@ class Auth(object):
                     Retrieve a username and password from $EARTHDATA_USERNAME and $EARTHDATA_PASSWORD.
             persist: Will persist credentials in a `.netrc` file.
             system: the EDL endpoint to log in to Earthdata, defaults to PROD
+            on_failure: Whether to print a warning or raise an error when login fails.
 
         Returns:
             An instance of Auth.
@@ -124,11 +128,11 @@ class Auth(object):
             return self
 
         if strategy == "interactive":
-            self._interactive(persist)
+            self._interactive(persist, on_failure=on_failure)
         elif strategy == "netrc":
-            self._netrc()
+            self._netrc(on_failure=on_failure)
         elif strategy == "environment":
-            self._environment()
+            self._environment(on_failure=on_failure)
 
         return self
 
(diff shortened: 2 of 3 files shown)


earthaccess/api.py
```diff
@@ -186,16 +202,25 @@ def login(strategy: str = "all", persist: bool = False, system: System = PROD) -
...
             try:
                 earthaccess.__auth__.login(
-                    strategy=strategy, persist=persist, system=system
+                    strategy=strategy,
+                    persist=persist,
+                    system=system,
+                    on_failure=on_failure,
                 )
-            except Exception:
-                pass
```
mfisher87, line 192: We don't want to skip no matter what error occurs; instead I'm using the exception type to communicate from the inner call whether we skip or not. There's a difference between "strategy unavailable" and "login failed", and we want the user to be able to tell!

mfisher87:
Any ideas why 3.13 integration tests are failing? I'm confused :)

chuckwondo (commented):
Thanks for picking up this issue.

Given that _not_ raising an error is what creates confusion and issue/discussion posts, I suggest we make "error" the default instead.

I suspect this will not only eliminate the bulk (if not all) of the confusion around the current behavior, but will be more in line with what is common behavior in other libraries that preemptively attempt to auth.

I would also suggest simply using a boolean (perhaps, `raise_on_failure: bool = True`). I think it would be a real stretch for there to be more than 2 ways to handle a failed login (i.e., either raise or don't).

I realize that making raising an error the default behavior qualifies as a breaking change, but in this case, I believe it would be the more familiar and less problematic default.

In other words, if we add this functionality with "warn" as the default, we're not breaking anything, but I would argue that we're perhaps not likely to reduce the number of questions about why things aren't working when login _appears_ to have succeeded.

Further, either way, some folks are likely to need/want to change code. If we make "warn" the default, folks who want to end their confusion/frustration must change their code to set the option to "error" (and _also_ continue to do so in future code). If we make "error" the default, the same users don't have to change code. Rather, they simply upgrade earthaccess.

Perhaps this is worthy of opening a discussion (unless I missed it).

mfisher87:
#947 for your second comment @chuckwondo

CHANGELOG.md
```diff
@@ -7,18 +7,25 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
...
+  ([**@mfisher87**](https://github.com/mfisher87))
+
 ## [v0.13.0] - 2025-01-28
 
 ### Changed
-- Integration tests: Test are no longer randomized! this means each fail should be reproducible, we are testing the most
-  popular datasets from all DAACs, see files under tests/integration/popular_collections.
+- Integration tests: Test are no longer randomized! this means each fail should be
+  reproducible, we are testing the most popular datasets from all DAACs, see files under
+  tests/integration/popular_collections.
```
chuckwondo, line 28: I thought we decided to exclude non-user-facing changes from the changelog.

docs/user_guide/backwards-compatibility.md
```diff
@@ -43,6 +43,8 @@ Our project follows the [SPEC0](https://scientific-python.org/specs/spec-0000/)
 
 ## Migration guides
 
-Under Development 🚧
+### 0.14.0
 
-This section will contain migration guides for future releases. Check back soon for updates!
+* `earthaccess.login()` will now fail by default if Earthdata Login rejects credentials.
```
chuckwondo, line 48:
```suggestion
* `earthaccess.login()` will now raise an exception by default if Earthdata Login rejects credentials.
```

docs/user_guide/backwards-compatibility.md
```diff
@@ -43,6 +43,8 @@ Our project follows the [SPEC0](https://scientific-python.org/specs/spec-0000/)
...
+  We recommend you adjust your code to handle this. If you'd really like to disable
```
chuckwondo, line 49:
```suggestion
  We recommend you adjust your code to handle this. If you want to disable
```

earthaccess/api.py
```diff
@@ -174,6 +189,7 @@ def login(strategy: str = "all", persist: bool = False, system: System = PROD) -
             * **"environment"**: retrieve username and password from `$EARTHDATA_USERNAME` and `$EARTHDATA_PASSWORD`.
         persist: will persist credentials in a .netrc file
         system: the Earthdata system to access, defaults to PROD
+        raise_on_failure: Whether to raise an error when login fails; if false, a warning will be printed
```
chuckwondo, line 192:
```suggestion
        raise_on_failure: Whether to raise an error when login fails; if `False`, only a warning will be printed
```

earthaccess/auth.py
```diff
@@ -112,6 +115,7 @@ def login(
                     Retrieve a username and password from $EARTHDATA_USERNAME and $EARTHDATA_PASSWORD.
             persist: Will persist credentials in a `.netrc` file.
             system: the EDL endpoint to log in to Earthdata, defaults to PROD
+            raise_on_failure: Whether to raise an error when login fails; if false, a warning will be printed
```
chuckwondo, line 118:
```suggestion
            raise_on_failure: Whether to raise an error when login fails; if `False`, only a warning will be printed
```

earthaccess/auth.py
```diff
@@ -234,63 +238,109 @@ class Session instance with Auth and bearer token headers
...
+            raise LoginStrategyUnavailable(f"No .netrc found at {netrc_loc}") from err
         except NetrcParseError as err:
-            raise NetrcParseError(f"Unable to parse .netrc file {netrc_loc}") from err
+            raise LoginStrategyUnavailable(
+                f"Unable to parse .netrc file {netrc_loc}"
+            ) from err
 
-        if (creds := my_netrc[self.system.edl_hostname]) is None:
+        creds = my_netrc[self.system.edl_hostname]
+        if creds is None:
```
chuckwondo, line 277: It seems that for consistency, we should also raise `LoginStrategyUnavailable` here too, no?

earthaccess/exceptions.py
```diff
@@ -0,0 +1,13 @@
...
+    This should be raised when a login strategy can't be attempted, for example because
+    "environment" was selected but the envvars are not populated.
+
+    DO NOT raise this exception when a login strategy is attempted and failed, for
+    example because credentials were rejected.
+    """
+
+
+class LoginAttemptFailure(Exception):
+    """The login attempt failed."""
```
chuckwondo, line 13: To avoid potential confusion with nothing but a docstring, I suggest adding `pass`:
```suggestion
    """The login attempt failed."""
    pass
```

earthaccess/exceptions.py
```diff
@@ -0,0 +1,13 @@
+class LoginStrategyUnavailable(Exception):
+    """The selected login strategy was skipped.
+
+    This should be raised when a login strategy can't be attempted, for example because
+    "environment" was selected but the envvars are not populated.
+
+    DO NOT raise this exception when a login strategy is attempted and failed, for
+    example because credentials were rejected.
+    """
```
chuckwondo, line 9:
```suggestion
    """
    pass
```

chuckwondo (dismissed):

earthaccess/exceptions.py
mfisher87: This isn't necessary since the docstring counts as the function body. If you still feel strongly about including `pass`, let me know!

earthaccess/exceptions.py
mfisher87: Same as above :)

earthaccess/auth.py
```diff
@@ -234,63 +238,109 @@ class Session instance with Auth and bearer token headers
...
             return False
```
mfisher87, line 278:
```suggestion
        if creds is None:
            raise LoginStrategyUnavailable(
                f"Earthdata Login hostname {self.system.edl_hostname} not found in .netrc file {netrc_loc}"
            )
```

earthaccess/auth.py
mfisher87: https://github.com/nsidc/earthaccess/pull/946/commits/bc187813e0e589b206fc43405da17ac0f5dfc577

mfisher87: Great catch :)

CHANGELOG.md
chuckwondo: Since you touched it, might as well just remove it, but not bothered if you leave it.

earthaccess/exceptions.py
chuckwondo: Right, I realize it's not necessary, but it may be confusing to others, so I suggest adding `pass` to avoid potential head-scratching by others who might stumble upon the code and wonder about its validity.

mfisher87: No worries! Adding.

CHANGELOG.md
mfisher87: OK I'll yank it since I want it gone anyway :)

mfisher87: faa295b

mfisher87:
pre-commit.ci autofix

CHANGELOG.md
```diff
@@ -7,18 +7,25 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
...
-### Changed
+### Added
 - `search_datasets` now accepts a `has_granules` keyword argument. Use
   `has_granules=False` to search for metadata about collections with no
   associated granules. The default value set in `DataCollections` remains `True`.
   ([#939](https://github.com/nsidc/earthaccess/issues/939))
   ([**@juliacollins**](https://github.com/juliacollins))
 
+### Changed
+- **Breaking**: Throw an error by default when login credentials are rejected. Pass
```
jhkennedy, line 18:
```suggestion
- **Breaking**: Earthaccess will throw an error by default when login credentials are rejected. Pass
```
?

earthaccess/exceptions.py
```diff
@@ -0,0 +1,17 @@
+class LoginStrategyUnavailable(Exception):
+    """The selected login strategy was skipped.
+
+    This should be raised when a login strategy can't be attempted, for example because
+    "environment" was selected but the envvars are not populated.
+
+    DO NOT raise this exception when a login strategy is attempted and failed, for
+    example because credentials were rejected.
```
jhkennedy, line 8:
```suggestion
    DO NOT raise this exception when a login strategy is attempted and fails. For
    example, this exception would not be thrown when credentials were rejected;
    a `LoginAttemptFailure` should be thrown instead.
```

earthaccess/exceptions.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+
+    pass
+
+
+class LoginAttemptFailure(Exception):
+    """The login attempt failed."""
```
jhkennedy, line 15:
```suggestion
    """The login attempt failed.
    
    This should be raised when a login attempt fails, for example, because
    the user's credentials were rejected.
    
    DO NOT raise this exception when a login strategy can't be attempted. For
    example, this exception would not be thrown when "environment" was selected
    but the envvars are not populated; a `LoginStrategyUnavailable` should be
    thrown instead.
    """
```

Should we be as descriptive as `LoginStrategyUnavailable`?

CHANGELOG.md
mfisher87: 5539a25

jhkennedy: I personally am on the other side of the house; I think changes that are useful for developers *or* users to know should be included as they are both "consumers" of earthaccess, and when I'm working on the code, the CHANGELOG is a lot closer than the releases.

But Common Chagelog defines it as "distributed form" not "repository" so, :shrug: . 

We did discuss it before, but I don't remember a definitive resolution. Maybe we should put a sentence at the top of the changelog about it?

earthaccess/exceptions.py
mfisher87: :100:

mfisher87:
pre-commit.ci autofix

jhkennedy (approved):
Looks good to me! @chuckwondo did you want to weigh in again?

mfisher87:
@chuckwondo if you're good, go ahead and merge. Otherwise I'll merge it tomorrow before releasing!

chuckwondo:
> @chuckwondo if you're good, go ahead and merge. Otherwise I'll merge it tomorrow before releasing!

Looks good, but of course I have to throw another (minor) wrench into the works :-)

I don't feel terribly strongly about the following, so I can likely be convinced to forget about it, but given that we are now raising an error by default, I'd actually suggest we _not_ add the `raise_on_failure` option.

My reasoning is that the language already provides a familiar way to handle this: try/except.

I'm taking a wild guess here, but I suspect nobody (or perhaps a nearly-zero percentage of users) check the result of the call to login to see if the Auth instance returned is authenticated (currently the only way to tell if login failed).

For example, I suspect (nearly) nobody is doing this, but please correct me if you've seen otherwise:

```python
import earthaccess

auth = earthaccess.login()

if not auth.authenticated:
    raise SomeException()

# do stuff requiring successful login
# ...
```

Instead, I suspect this is the common coding pattern, given the number of people who report having the problem that a failed login does "nothing":

```python
import earthaccess

earthaccess.login()

# do stuff requiring successful login
# ...
```

If the former pattern were common, (almost) nobody would be "complaining" about this issue.

If my suspicion is current, then (almost) nobody would then bother to make the following code change, once this PR (as it stands) were to land:

```python
import earthaccess

earthaccess.login(raise_on_error=False)

# do stuff requiring successful login
# ...
```

Conversely, if somebody does indeed use the latter approach (i.e., check the value of `auth.authenticated`), then they can simply remove the check and simply let the error propagate. Alternatively, if such a user is checking the value of `auth.authenticated` and taking a different course of action, then they would simply replace the check with a try/except block.

Bottom line: I don't believe there's a solid reason for us to bother adding a `raise_on_failure` option. Simply always raise on failure and allow the user to deal with it as they see fit via try/except (or not).

mfisher87:
That's an excellent point. We don't need to clutter our API for this. Thanks for the :brain: time Chuck :heart:

jhkennedy:
> I don't believe there's a solid reason for us to bother adding a raise_on_failure option. Simply always raise on failure and allow the user to deal with it as they see fit via try/except (or not).

Overall I agree with you @chuckwondo, and would discourage using `raise_on_failure=False`. I'm also having a hard time thinking of a legitimate use case for it... what usage path would need you to login but not be authenticated?

Even in places where I have seen `auth.authenticated` checked, it'd be more straightforward for the code to raise if it failed instead:
https://github.com/nasa-openscapes/earthdata-cloud-cookbook/blob/main/tutorials/earthaccess-demo.ipynb

mfisher87:
Thanks Joe, great points as well! Returning this to draft and re-working to always arise and log an error-level message.

mfisher87:
@jhkennedy @chuckwondo done!

earthaccess/api.py
```diff
@@ -161,7 +170,11 @@ def search_services(count: int = -1, **kwargs: Any) -> List[Any]:
     return query.get(hits if count < 1 else min(count, hits))
 
 
-def login(strategy: str = "all", persist: bool = False, system: System = PROD) -> Auth:
+def login(
+    strategy: str = "all",
+    persist: bool = False,
+    system: System = PROD,
+) -> Auth:
```
chuckwondo, line 177: Please add a `Raises` section to the docstring.

earthaccess/auth.py
chuckwondo, line 1: Please add a Raises section to the docstring of the `login` method.

chuckwondo (changes requested):
Looks much leaner now :-)

Just a couple of requests to update docstrings with exception info.

earthaccess/auth.py
mfisher87: 3e009e6

earthaccess/api.py
mfisher87: 3e009e6

chuckwondo (approved):

Issue #807, opened 2024-09-16
[BUG] s3 download does not check for existing file


meteodave:
### Is this issue already tracked somewhere, or is this a new report?

- [X] I've reviewed existing issues and couldn't find a duplicate for this problem.

### Current Behavior

All objects by s3 earthaccess.download are transferred even if they exist locally.

### Expected Behavior

I expected that the s3 earthaccess download would check the object is already available locally (similar to the HTTP) and skip the transfer (e.g., "File MYD04_3K.A2019214.1920.061.2019215152349.hdf already downloaded). 

### Steps To Reproduce

Compare the earthaccess.download via HTTP versus s3 for a cloud enabled data set.

### Environment

```markdown
- OS:Ubuntu 20.04.6
- Python:3.8.10
```

itcarroll:
Thank you, @meteodave. I can reproduce.

It looks like `earthaccess.Store. _download_file‎` implements a simple check for whether the path exists, which is not done in either `earthaccess.Store._get_granules` or `earthaccess.Store._get_urls` when using "direct" access logic.

Sherwin-14:
@itcarroll I am interested to work on this. I am thinking of adding `earthaccess.Store. _download_file‎` checks for the above two functions you listed. Anything else I need to take care of in your opinion?

PR #1002, opened 2025-05-02
Added checks for _get_granules and _get_urls which enables them to check for existing files.

Sherwin-14:
I have added checks for the functions as discussed in the [issue](https://github.com/nsidc/earthaccess/issues/807). Please let me know about any problems with this implementation.

----
📚 Documentation preview 📚: https://earthaccess--1002.org.readthedocs.build/en/1002/

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -587,8 +587,18 @@ class Store(object):
         if self.in_region and data_links[0].startswith("s3"):
             logger.info(f"Accessing cloud dataset using provider: {provider}")
             s3_fs = self.get_s3_filesystem(provider=provider)
-            # TODO: make this parallel or concurrent
+
+            _data_links = []
+            
             for file in data_links:
+                file_mod = file.split('/')[-1]
+                temp_local_path = local_path.parent.parent / 'data' 
+                file_paths = map(lambda x: x.split('/')[-1], set(glob.iglob(str(temp_local_path) + '/*/*', recursive = True)))
+                if not (file_mod in file_paths):
+                    _data_links.append(file)
+
+            # TODO: make this parallel or concurrent
+            for file in _data_links:
                 s3_fs.get(file, str(local_path))
                 file_name = local_path / Path(file).name
                 logger.info(f"Downloaded: {file_name}")
@@ -631,7 +641,7 @@ class Store(object):
         for file in _data_links:
             file_mod = file.split('/')[-1]
             temp_local_path = local_path.parent.parent / 'data' 
-            file_paths = map(lambda x: x.split('/')[-1], list(glob.iglob(str(temp_local_path) + '/*/*', recursive = True)))
+            file_paths = map(lambda x: x.split('/')[-1], set(glob.iglob(str(temp_local_path) + '/*/*', recursive = True)))
             if not (file_mod in file_paths):
                data_links.append(file)
 
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -1,3 +1,4 @@
+import os
 import datetime
 import logging
 import threading
@@ -621,11 +622,17 @@ class Store(object):
                 granule.data_links(access=access, in_region=self.in_region)
                 for granule in granules
             )
-        )
+        )   
         total_size = round(sum(granule.size() for granule in granules) / 1024, 2)    
         logger.info(
             f" Getting {len(granules)} granules, approx download size: {total_size} GB"
         )
+        for file in data_links:
+                file_mod = file.split('/')[-1]
+                file_name = local_path / Path(file_mod).name
+                raise ValueError(file,  file_name)
+                if os.path.exists(file_name):
+                    data_links.remove(file)
         if access == "direct":
             if endpoint is not None:
                 logger.info(
@@ -637,16 +644,7 @@ class Store(object):
                 s3_fs = self.get_s3_filesystem(provider=provider)
 
             local_path.mkdir(parents=True, exist_ok=True)
-            for links in data_links:
-                if "opendap" in links and links.endswith(".html"):
-                    links = links.replace(".html", "")
-                local_filename = links.split("/")[-1]
-                path = local_path / Path(local_filename)
 
-                if path.exists():
-                    data_links.remove(links)
-                else:
-                    pass    
             # TODO: make this async
             for file in data_links:
                 s3_fs.get(file, str(local_path))
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -623,6 +623,16 @@ class Store(object):
             )
         )
         total_size = round(sum(granule.size() for granule in granules) / 1024, 2)
+        for links in data_links:
+            if "opendap" in links and links.endswith(".html"):
+                links = links.replace(".html", "")
+            local_filename = links.split("/")[-1]
+            path = local_path / Path(local_filename)
+
+            if path.exists():
+                data_links.remove(links)
+            else:
+                pass        
         logger.info(
             f" Getting {len(granules)} granules, approx download size: {total_size} GB"
         )
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -622,17 +622,7 @@ class Store(object):
                 for granule in granules
             )
         )
-        total_size = round(sum(granule.size() for granule in granules) / 1024, 2)
-        for links in data_links:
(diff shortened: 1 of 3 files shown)


earthaccess/store.py
```diff
@@ -586,8 +587,21 @@ def _get_urls(
         if self.in_region and data_links[0].startswith("s3"):
             logger.info(f"Accessing cloud dataset using provider: {provider}")
             s3_fs = self.get_s3_filesystem(provider=provider)
-            # TODO: make this parallel or concurrent
+
+            _data_links: List = []
+
```
chuckwondo, line 592: This is unnecessary.
```suggestion
```

earthaccess/store.py
```diff
@@ -1,4 +1,5 @@
 import datetime
+import glob
```
chuckwondo, line 2: This is unnecessary.
```suggestion
```

earthaccess/store.py
```diff
@@ -586,8 +587,21 @@ def _get_urls(
...
             for file in data_links:
+                file_mod = file.split("/")[-1]
+                temp_local_path = local_path.parent.parent / "data"
+                file_paths = map(
+                    lambda x: x.split("/")[-1],
+                    set(glob.iglob(str(temp_local_path) + "/*/*", recursive=True)),
+                )
+                if file_mod not in file_paths:
+                    _data_links.append(file)
+
```
chuckwondo, line 602: Unnecessary
```suggestion
```

earthaccess/store.py
```diff
@@ -586,8 +587,21 @@ def _get_urls(
...
+            # TODO: make this parallel or concurrent
+            for file in _data_links:
                 s3_fs.get(file, str(local_path))
                 file_name = local_path / Path(file).name
                 logger.info(f"Downloaded: {file_name}")
```
chuckwondo, line 607:
```suggestion
                if not local_path.exists():
                    s3_fs.get(file, str(local_path))
                    file_name = local_path / Path(file).name
                    logger.info(f"Downloaded: {file_name}")
```

chuckwondo (changes requested):

earthaccess/store.py
```diff
@@ -626,6 +640,17 @@ def _get_granules(
         logger.info(
             f" Getting {len(granules)} granules, approx download size: {total_size} GB"
         )
+
```
itcarroll, line 643: I don't know why we are duplicating code here instead of letteing _get_granules call _get_urls.

earthaccess/store.py
itcarroll: Isn't `local_path` the directory? The check needs to be on `file_name.exists()` Otherwise, agreed.

Sherwin-14:
> This type of check should be all you need to do, and do so similarly in the `_get_granules` method.

 I don't think this will work. The local path will be different for every invocation of `earthaccess.download()`.
```
local_path: Local directory to store the remote data granules.  If not
            supplied, defaults to a subdirectory of the current working directory
            of the form `data/YYYY-MM-DD-UUID`, where `YYYY-MM-DD` is the year,
            month, and day of the current date, and `UUID` is the last 6 digits
            of a UUID4 value.
```

I tested this out and this is my file structure after first invocation of `earthaccess.download()`.

```
data/
└── 2025-05-05-110539
    ├── ATL06_20181014034354_02370106_006_02.h5
    ├── ATL06_20181014035224_02370107_006_02.h5
    ├── ATL06_20181014154641_02450101_006_02.h5
    ├── ATL06_20181014155343_02450102_006_02.h5
    ├── ATL06_20181015152102_02600101_006_02.h5
    ├── ATL06_20181015152804_02600102_006_02.h5
    ├── ATL06_20181016025235_02670106_006_02.h5
    └── ATL06_20181016145523_02750101_006_02.h5

2 directories, 8 files

```
The folder name for first invocation is `2025-05-05-110539`.  Then I tried another invocation of  `earthaccess.download()` and I got this value of `local_path`.

```
/home/$USER/Projects/earthaccess/data/2025-05-05-1e0ead
```

Even with @itcarroll  suggestion, this will not work as the folder is non existent.

chuckwondo:
> > This type of check should be all you need to do, and do so similarly in the `_get_granules` method.
> 
> I don't think this will work. The local path will be different for every invocation of `earthaccess.download()`.
> 
> ```
> local_path: Local directory to store the remote data granules.  If not
>             supplied, defaults to a subdirectory of the current working directory
>             of the form `data/YYYY-MM-DD-UUID`, where `YYYY-MM-DD` is the year,
>             month, and day of the current date, and `UUID` is the last 6 digits
>             of a UUID4 value.
> ```
> 
> I tested this out and this is my file structure after first invocation of `earthaccess.download()`.
> 
> ```
> data/
> └── 2025-05-05-110539
>     ├── ATL06_20181014034354_02370106_006_02.h5
>     ├── ATL06_20181014035224_02370107_006_02.h5
>     ├── ATL06_20181014154641_02450101_006_02.h5
>     ├── ATL06_20181014155343_02450102_006_02.h5
>     ├── ATL06_20181015152102_02600101_006_02.h5
>     ├── ATL06_20181015152804_02600102_006_02.h5
>     ├── ATL06_20181016025235_02670106_006_02.h5
>     └── ATL06_20181016145523_02750101_006_02.h5
> 
> 2 directories, 8 files
> ```
> 
> The folder name for first invocation is `2025-05-05-110539`. Then I tried another invocation of `earthaccess.download()` and I got this value of `local_path`.
> 
> ```
> /home/$USER/Projects/earthaccess/data/2025-05-05-1e0ead
> ```
> 
> Even with @itcarroll suggestion, this might not work.

When the user does not supply a local path, then yes, you are correct, due to the default logic that generates a different local path on _every_ invocation. This is exactly why I was never in favor of this default behavior. In fact, I was never in favor of using any sort of default location, and still feel that the local path should always be supplied by the user to avoid situations such as this one. However, that is perhaps worthy of opening a new discussion for.

For this particular issue, while the additional logic that you originally proposed might avoid duplicate downloads, the problem is that it makes it very hard (and likely very frustrating, as I imagine it) for a user to realize that the most recent default download path does _not_ contain the files they requested because the files were previously downloaded to some other default path that was created the first time they requested the files.

I imagine this would be the source of yet more issues/bugs being reported by users, along the lines of "download succeeded, but did not actually download my files" because it would be completely non-obvious that the files are sitting in a different directory.

Therefore, for this PR, I suggest that the logic remain very simple. The simplified code suggestions I made will do the following:

1. keep things as simple as possible
2. avoid confusing users as I just described (this will duplicate downloads, but will avoid unnecessary additional bug reporting as mentioned above)
3. require users to specify a download path in order to avoid duplicate downloads
4. will still work even if we decide to change the "unique download path" logic

Separately, I suggest we open a new discussion regarding the unique download paths that get generated by default.

cc: @mfisher87, @jhkennedy, @betolink

earthaccess/store.py
chuckwondo: > Isn't `local_path` the directory? The check needs to be on `file_name.exists()` Otherwise, agreed.

Ah, yes, I didn't look closely enough.

Sherwin-14:
@chuckwondo @itcarroll Thanks for explaining all of this! Should we add this fact (duplicating downloads only for `default_path` and not when `local_path` is provided by the user) into our docs as well? I guess this would not be intuitive to users, or maybe I am wrong, what do you guys think?

chuckwondo (commented):
Thanks @Sherwin-14. The changes look good. I've triggered a re-run of the integration tests to make sure this doesn't break anything.

@mfisher87, @itcarroll, @betolink, or @jhkennedy, do any of you have any leanings one way or the other about the need to add some tests for this change?

itcarroll (approved):
Minimal and gets the bug fixed.

I do think the suggestion to enhance the documentation is a good one. Perhaps add the sentence "Provide the optional `local_path` argument to prevent repeated downloads." to all the relevant download methods.

Sherwin-14:
> Provide the optional `local_path` argument to prevent repeated downloads."

How about this "If set manually, it prevents repeated downloads.". I'll add this to the docstring.

itcarroll:
I like it mentioned in the description, not just in the argument summary, which is why I gave the argument name.

itcarroll (approved):

chuckwondo (approved):
Thanks Sherwin!

PR #1006, opened 2025-05-06
Write PR comment when int. tests fail on insufficient perms

### LINKED ISSUE #1005: Automatic PR comment fails
Our attempts to automatically add a PR comment when a user has insufficient permission to run integration tests in CI [still doesn't work](https://github.com/nsidc/earthaccess/actions/runs/14862911271/job/41732206403), but I believe we are a bit closer than we were before.

chuckwondo:
Yet another attempt to get auto-commenting on PRs to work.

Fixes #1005

----
📚 Documentation preview 📚: https://earthaccess--1006.org.readthedocs.build/en/1006/

### DIFF
--- a/.github/workflows/integration-test-review.yml
+++ b/.github/workflows/integration-test-review.yml
@@ -19,11 +19,52 @@ jobs:
           require: write
           username: ${{ github.triggering_actor }}
 
-      - name: Add PR comment when user does not have write permission
+      # The remaining steps are run only if the triggering actor (user) does NOT have
+      # write permission for the repo, in which case, the purpose of the remaining
+      # steps is to add a comment to the PR indicating that the actor (user) does
+      # NOT have permissions to run integration tests, and thus a maintainer must
+      # make sure the PR is "safe", and if so, re-run the "failed" integration tests
+      # (because maintainers DO have the necessary write permission).
+
+      - name: Download PR number artifact
         # The name of the output require-result is a bit confusing, but when its value
         # is 'false', it means that the triggering actor does NOT have the required
         # permission.
-        if: steps.permission.outputs.require-result == 'false'
+        if: ${{ !env.ACT && steps.permission.outputs.require-result == 'false' }}
+        uses: actions/github-script@v7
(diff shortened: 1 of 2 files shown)


mfisher87:
Is there a possibility of a race condition when 2 PRs from forks come in around the same time? My gut feeling is that there _must_ be an alternative to writing an artifact for this. My brain has no alternative to offer, however :laughing:

chuckwondo:
> Is there a possibility of a race condition when 2 PRs from forks come in around the same time? My gut feeling is that there _must_ be an alternative to writing an artifact for this. My brain has no alternative to offer, however 😆

Regarding a race condition, no, 2 separate PRs will have 2 separate, non-interfering artifacts.

Regarding a different approach, unfortunately, I don't think so. This appears to be one of those annoying github workflow situations for which we have to do some awkward contortions. The [previous problem](https://github.com/nsidc/earthaccess/actions/runs/14862911271/job/41732206403) was that the expression `context.issue.number` resulted in an empty value (in the context of a `workflow_run` event), which is why the failure was due to a "404". Instead of resulting in a URL like `https://api.github.com/repos/nsidc/earthaccess/issues/NNN`, where `NNN` would be the issue number, we simply get `https://api.github.com/repos/nsidc/earthaccess/issues/` because `conext.issue.number` is empty, and that resulted in the 404 error.

The solution I've implemented here is straight from the docs on `workflow_run`: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#using-data-from-the-triggering-workflow (and I included that URL in a comment in `.github/workflows/integration-test.yml` for reference).

mfisher87:
Looks promising, let's try :D

mfisher87 (approved):

mfisher87:
Thanks for continuing to work to find a fix for this, Chuck!

chuckwondo:
Fingers crossed! I believe this might be more than the 3rd try, so I think we're past the point of, "Third time's the charm!"

PR #1013, opened 2025-05-16
Fix path to pr_number.zip artifact

chuckwondo:
Hopefully this is the last fix to get PR comments to work. The workflow was attempting to unzip the pr_number.zip artifact at the wrong path, so was failing the workflow to add the PR comment.

These changes _should_ allow the unzip to work now, which should then allow the PR number to be read from the unzipped artifact, which should then allow a comment to be added to the PR.

----
📚 Documentation preview 📚: https://earthaccess--1013.org.readthedocs.build/en/1013/

PR #1012, opened 2025-05-15
Add an exception `DownloadFailure` for handling failure.

Sherwin-14:
I have added the logic for handling download failure as discussed [here](https://github.com/nsidc/earthaccess/issues/996).  Also I observed that we have same block for Changed twice in CHANGELOG, is this done on purpose?

----
📚 Documentation preview 📚: https://earthaccess--1012.org.readthedocs.build/en/1012/

PR #1014, opened 2025-05-17
Fix link to PR files in PR auto-comment

chuckwondo:
The auto-comment on PRs is now working! See #1013 for the fix, and #1012 for correctly added comments.

However within the auto-generated comment, the link on the text "code changes in this pull request" is broken, so it doesn't properly link to the files (diff) view of the PR.

This PR attempts to fix the broken link in the generated comment.

----
📚 Documentation preview 📚: https://earthaccess--1014.org.readthedocs.build/en/1014/

### DIFF
--- a/.github/workflows/integration-test-review.yml
+++ b/.github/workflows/integration-test-review.yml
@@ -57,6 +57,9 @@ jobs:
             fs.writeFileSync(path.join(temp, 'pr_number.zip'), Buffer.from(download.data));
 
       - name: Unzip downloaded PR number artifact
+        # The name of the output require-result is a bit confusing, but when its value
+        # is 'false', it means that the triggering actor does NOT have the required
+        # permission.
         if: ${{ !env.ACT && steps.permission.outputs.require-result == 'false' }}
         run: unzip "${{ runner.temp }}/artifacts/pr_number.zip" -d "${{ runner.temp }}/artifacts"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-test-review.yml
```diff
@@ -80,15 +83,15 @@ jobs:
             const { owner, repo } = context.repo;
 
             // Read the PR number from the downloaded and unzipped artifact.
-            const issue_number = Number(fs.readFileSync(path.join(temp, 'pr_number')));
+            const pr_number = Number(fs.readFileSync(path.join(temp, 'pr_number')));
 
             // Get the URL of the PR so we can add a link in the PR comment.
-            const { html_url } = await github.rest.issues.get({ owner, repo, issue_number });
+            const { html_url } = await github.rest.pulls.get({ owner, repo, pr_number });
```
mfisher87, line 89: Interesting that the REST API cares about issues vs PRs; the web interface treats them the same, i.e. `.../issues/1014` would redirect here. :shrug:

mfisher87 (approved):

.github/workflows/integration-test-review.yml
chuckwondo: Yes, it's rather confusing. I noticed the redirection too, but for some reason, the value of `html_url` was coming back as `undefined`, thus causing the broken link. I _think_ that using `pulls` instead of `issues` should give us back the expected URL.

mfisher87: Awesome job getting the workflow working ;)

PR #1015, opened 2025-05-23
Fix method resolution bug when base class methods override getattr

### LINKED ISSUE #610: method resolution surprise on `EarthAccessFile`

alexandervladsemenov:
This PR closes #610.

The issue was that the `__getattr__` method does not override methods from the base class. As a result, EarthAccessFile methods were inherited from the base class fsspec.spec.AbstractBufferedFile.

- To address this, `__getattribute__` is used instead of `__getattr__`. The implementation is designed to avoid infinite recursion.

- The base class has been changed from fsspec.spec.AbstractBufferedFile to io.IOBase to ensure correct reading from in-memory files using xarray.

- An additional test has been added to verify in-memory reading with xarray.

----
📚 Documentation preview 📚: https://earthaccess--1015.org.readthedocs.build/en/1015/


itcarroll:
Thanks for the PR! I will work out how to make the integration-tests run on GitHub Actions.

@betolink @mfisher87 I'm surprised there is no previous test using `xarray.open_*`, so where should such a test go? This PR has it with `tests/unit/test_store` but it seems to me like a new file called `tests/integration/test_xarray.py` would be better.

itcarroll:
Have confirmation from in [zulipchat](https://earthaccess.zulipchat.com/#narrow/channel/480557-general/topic/xarray.20tests/with/521760667) that `test_earthaccess_xarray_access` should be moved under `tests/integration`. I suggest a `test_xarray` module, and renaming to `test_open_dataset`.

alexandervladsemenov:
> Have confirmation from in [zulipchat](https://earthaccess.zulipchat.com/#narrow/channel/480557-general/topic/xarray.20tests/with/521760667) that `test_earthaccess_xarray_access` should be moved under `tests/integration`. I suggest a `test_xarray` module, and renaming to `test_open_dataset`.

@itcarroll   I've created "test_xarray" integration test and pushed an update.

betolink:
thanks for the PR @alexandervladsemenov! and sorry for the late late review, this summer has been a little hectic. Question:  do you know if io.IOBase buffers are serializable in the same way `AbstractBufferedFile` are? this helper class exists mostly to "switch" the file system on the fly if we open data out of region where S3 is not allowed. I'll test this with Coiled, I assume it should be the same. The integration tests fail for other reasons, I'll wait for @chuckwondo opinion and we should merge this early next week if everything works as expected!

Also cc to @martindurant on this `AbstractBufferedFile` fsspec question.

alexandervladsemenov:
> thanks for the PR @alexandervladsemenov! and sorry for the late late review, this summer has been a little hectic. Question: do you know if io.IOBase buffers are serializable in the same way `AbstractBufferedFile` are? this helper class exists mostly to "switch" the file system on the fly if we open data out of region where S3 is not allowed. I'll test this with Coiled, I assume it should be the same. The integration tests fail for other reasons, I'll wait for @chuckwondo opinion and we should merge this early next week if everything works as expected!
> 
> Also cc to @martindurant on this `AbstractBufferedFile` fsspec question.

Hi @betolink No,  io.IOBase is not serializable. However,  since the `__reduce__` is implemented 

```py
    def __reduce__(self) -> Any:
        return make_instance, (
            type(self.f),
            self.granule,
            earthaccess.__auth__,
            dumps(self.f),
        )
 ```       
 
 and `f` is serializable, I believe it will  work with Dask/Coiled distributed computing. You can check it if it fails when Coiled sends them to workers

**Comment by itcarroll** [comment]
@betolink I'm going to try to figure out and add a test today that will check dask's ability to reopen these file-likes.

**Comment by betolink** [comment]
@itcarroll this one is tricky I think running `open()` using Coiled would be the only "real world" scenario. I did some testing yesterday and it seems like it works as expected (with the https links)

A change we should consider soon is to get rid of `pqdm` in favor of a generic parallel executor (with a progress bar) like what VirtualiZarr did here: https://github.com/zarr-developers/VirtualiZarr/blob/main/virtualizarr/parallel.py 

Then our `open()` and `download()` won't be limited to local execution and that counts when we would like to process 1000+ files. I'll open a ticket after these changes and the next release.

**Comment by martindurant** [comment]
Following.
I didn't really understand the thread in the issue, what exactly fsspec's File is doing wrong here. A couple of notes:
- using getattribute is annoying and fragile, but still quite common in the fsspec codebase for this kind of thing. Whenever you implement a new method on your subclass, you need to update the set of strings in getattribute AND care about methods they call in turn, or get long error tracebacks
- assigning a bound method is probably not great; couldn't you have a single read() method that dispatches instead?
- fsspec files and OpenFile objects are designed to be used in a context (`with`); the contract is, that it must be serialisable _outside_ the context

**Comment by betolink** [comment]
>  fsspec's File is doing wrong here

I don't think this is a fsspec doing something wrong, this is an optimization @jrbourbeau implemented awhile back. NASA Earthdata has 2 distribution links for a file, `s3://` only works if we are in region (bucket policy restrictions), `https://` works out of region and when we use remote compute this trick reopens the file using S3 even though started as an HTTP link. This is a very nice but rarely used feature, mostly if we have a Dask cluster already running in AWS (e.g. Coiled), at the time James said that it speed up things by 2X. If this works as expected it may worth keeping it around. Then @itcarroll ran into some issue with the way this thin wrapper was implemented #610

**Comment by itcarroll** [comment]
Thanks for comments, @betolink and @martindurant. Defintely not an `fsspec` issue.

I've sent suggestions to my colleague @alexandervladsemenov, and think we're almost there. I'm suggesting:

1. a partial check on serialization using `dask.distributed.Client.submit`
    https://github.com/itcarroll/earthaccess/blob/2f5f01e5a672aa1e6a31788d6e5171147a9657a2/tests/integration/test_xarray.py#L25-L29

3. to address some of @martindurant's first concern (fragility of `__getattribute__`) by casing `__getattribute__` behaviour on the contents of a `__dict__` rather than manually curating.
   https://github.com/itcarroll/earthaccess/blob/2f5f01e5a672aa1e6a31788d6e5171147a9657a2/earthaccess/store.py#L78-L84

~~Unfortunately, I think my test is running into dask/distributed#9013.~~

**Comment by betolink** [comment]
@itcarroll I don't think we need xarray/h5py to test this, just reading some raw bytes with the fsspec instance should be enough. I just saw a similar issue Ryan reported but it's more recent, seems like using xarray<2025.6 should work  https://github.com/pydata/xarray/issues/10712

**Comment by betolink** [comment]
Should we merge this or do we want to do some refactoring and more testing? it's all working now for what we need it, let me know. @itcarroll @alexandervladsemenov

File: tests/integration/test_xarray.py [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+    ds = xr.Dataset({"y": ("x", list(range(8)))})
+    fs = fsspec.filesystem("memory")
+    path = "foo"
+    with fs.open(path, mode="wb") as f:
+        ds.to_netcdf(f, engine="h5netcdf")
+    f = fs.open(path)
+    earthaccess_file = EarthAccessFile(f, granule="foo")
+    # confirm backend detection
+    backends = xr.backends.list_engines()
+    assert backends["h5netcdf"].guess_can_open(earthaccess_file)
```
betolink, line 16: Seems like this is triggering the scipy engine which for some reason is not installed for the tests, maybe we need to explicitly install all the xarray back ends in the test group? 

image

itcarroll:
Failure from `xarray.to_netcdf` is picking `scipy` ... I'm trying to figure out why it's not writing with `h5netcdf` in the test environment.

itcarroll:
My test would pass with `xarray>=2025.8` due to https://github.com/pydata/xarray/pull/10557. We are pinned to `xarray<2025.6` but with no explanation.

betolink:
Should we unpin xarray? the reason to pin it was to avoid the cloud pickle regression. @itcarroll

earthaccess/store.py
```diff
@@ -153,7 +163,7 @@ def make_instance(
     ):
         # NOTE: This uses the first data_link listed in the granule. That's not
```
betolink, line 164: I wonder if this note needs to be revised, if I recall correctly there was a mapping granule to URLs method to zip them and all of the assets/files in the granue under GET_DATA are opened. Testing HSL or some of the collections that have multiple files will clarify the behavior.

tests/unit/test_store.py
```diff
@@ -209,15 +209,11 @@ def mock_download_file(url):
...
-    reason="Expected failure: Reproduces a bug (#610) that has not yet been fixed."
-)
 def test_earthaccess_file_getattr():
     fs = fsspec.filesystem("memory")
-    with fs.open("/foo", "wb") as f:
-        earthaccess_file = EarthAccessFile(f, granule="foo")  # type: ignore
-        assert f.tell() == earthaccess_file.tell()
-    # cleanup
+    with fs.open("foo", "wb") as f:
+        earthaccess_file = EarthAccessFile(f, granule="foo")
```
betolink, line 215: nice!

itcarroll:
Yeah! Tests passing. I'd updated the test so your cloud pickle dodge isn't needed. I'm just adding the changelog entry, but I think it's good to go.

betolink (approved):
Great PR, approving pending docs.

tests/integration/test_xarray.py
itcarroll: Since xarray 2025.8 the correct backend (h5netcdf, which is installed) get picked up rather than scipy.

PR #1011, opened 2025-05-13
Decision record for #231: in region checking method

jhkennedy:

Issue #929, opened 2025-01-22
Fork to independent organization


mfisher87:
This project is community-owned and I propose that we reflect that by forking to an independent organization with ownership shared between interested individuals and organizations.

mfisher87:
We could use this fork: https://github.com/earth-observation-community/earthaccess

Is there a better org name available / reserved?

JessicaS11:
Who is in/on the earth-observation-community org? There are no public members, so I'd want to make sure we've communicated with them directly around adding a few folks as maintainers if we're going to use this as the primary copy of earthaccess.

I know there's some ongoing discussion elsewhere with the name (can't recall where ATM), but I'll also put forth https://github.com/earthaccess-contrib

mfisher87:
>Who is in/on the earth-observation-community org?

Just one other person has accepted the invitation and has owner rights. I prefer not to name drop on this sensitive issue in a public place, but I'm in the process of inviting every earthaccess collaborator with the same rights on this org.

No strong feelings on the name personally. Do we want to build an org that's more earthaccess-focused or do we want to help other projects that aspire to be community owned incubate in the new org? 🤷

mfisher87:
Invites are out.

betolink:
I think moving the project into a neutral organization would be a reasonable approach. IMO this would be better than forking and may serve the interests of all parties involved—unless the community and maintainers feel that forking is the better path forward.

I don’t think this change should impact the overall goals of the project, In any case, I fully support the idea that these decisions should be made collectively by the community, with the advancement of science as our shared priority.

mfisher87:
If we move we can keep our issues and discussions and stuff, but not if we fork. It's a tough call; I can imagine a future where we have to deal with negative consequences because we didn't make a clean enough break. What if, for example, there is a dispute over the ownership of the name _earthaccess_?

betolink:
The name is not a NASA brand, it was changed to avoid this from earthdata.py to earthaccess.py. I think it's worth considering the idea proposed by @chuckwondo on exploring EODAG https://eodag.readthedocs.io/en/stable/ seems like they have the architecture in place to expand the work we are doing here. 

In the other hand, I recognize that a lot of the value we provide to the research community is our shared knowledge on the quirks of CMR and how DAACs operate and just migrating to a different code base will be difficult. But it could be a valid parallel effort.

mfisher87:
Exploring EODAG would be a fun hackathon activity that we could do as a large group!

[...]

PR #1030, opened 2025-06-10
Add documentation of the Decision Committee process

jhkennedy:
In the May 13th hackday, we discussed the decision committee a bit and started to develop what the actual process could/should look like. 

We decided to try and use standard GitHub PR workflows to facilitate a decision, which I've outlined here, and use #1011 to work through the processes and try it out.

I'm leaving this as draft while we work through the process in #1011 as I expect we will update it as we go, but this will at least provide a frame of reference. 

TODOs:
- [ ] Write a GitHub action that creates a draft decision record PR that is triggered on labeling an issue or PR that it https://github.com/nsidc/earthaccess/labels/needs%3A%20decision
   - [ ] :question: Possibly also 'on: workflow_dispatch` to create one manually
   - [ ] :question: use a ruleset or step in the PR to auto-assign the PR (see: [enabling auto assignment](https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#about-auto-assignment) for one way to do this) 
- [ ] Ensure CODEOWNERS is accurate and everyone listed is on board with how they are listed in the doc
- [ ] :question: Add this and any decisions to our website

--- 

> [!NOTE]
> This would be a lot easier to manage using teams, which would also be a lot easier to manage in an org, so we might want to make a decision for #929 as well

### DIFF
--- a/docs/governance/governance-process.md
+++ b/docs/governance/governance-process.md
@@ -1,0 +1,38 @@
+# Earthaccess Governance
+
+Earthaccess is an open source project devoted to providing easy access to NASA's Earth data. Earthaccess, importantly, is developed _for_ and _by_ the community.  
+
+As with any project, from time to time, decisions will need to be made about direction, features, engineering choices. Eathaccess values reaching a community consensus on those decisions and including as many community perspectives as possible.
+
+Examples of when the decision committee may need to be invoked:
+* Issues that relate to **significant changes** to earthaccess framework, or way we develop earthaccess
+* When there is a **disagreement** on a path forward
+
+In order to reach consensus, Earthacess has a developed an _asynchronous_ decision processes, that in brief looks like:
+
+1. GitHub issues can be labeled `needs: decision` by anyone to kick off the decision committee process
+2. A PR will be automatically created with a blank decision record document and a member of the decision committee will be assigned at random to shepherd the decision through the processes
+3. The assignee will create a _first_ draft of the decision record document, summarizing the relevant issues, discussion, and pull requests, and then mark it as ready for review
+4. The decision committee will use PR review mechanics to discuss the decision and suggest changes or edit the decision record to reflect the decision and discussion
+   - Decision committee members endorse a decision as written by approving the PR
+   - Decision committee members can reject a decision _as written_ by requesting changes in the PR
+   - Decision committee member can express neutral discussion points by leaving PR comments 
+5. Once N decision committee members endorse a decision (approve the PR), the decision is considered "decided" and the PR should be merged to record the decisions
+6. The PR assignee should update any relevant GitHub issues, discussion, and pull requests with the decision and a link back to the decision record PR
+
+## Decision Committee Members
+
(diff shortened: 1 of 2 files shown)


jhkennedy:
One thing we'll need to look out for is that whoever creates the PR can't approve or request changes:
image

So they effectively can't vote as I've outlined the process. We can use a "bot" user for the GitHub Actions to get around that, or we might be able to specify it in some other way. 

Another option would be to have it appear that whoever is assigned the decisions (and shepherding it) created the PR so they can't vote (neutral) and they can cast the final "tie-breaker" deciding vote by merging the PR, depending on how we set up required reviews and whatnot...

mfisher87:
I love using GHA to create the PR from a label or manual trigger! [Here's a workflow I use for creating meeting notes documents using a Jinja template](https://github.com/mfisher87/hackmd-meeting-notes-action/blob/main/.github/workflows/create-meeting-notes-pr.yml). Something similar could be used to pre-populate a decision template from an issue.

mfisher87:
Also, I think the template could be the place where we keep the source of truth for the process. Then every time a PR is opened from that template, the process is right there for everyone to be on the same page about endorsement threshold, etc. And as the process changes, we have a record of which version of the process was used for each decision.

itcarroll:
I don't know what GHAs are capable of, but it would also be nice to have one comment with the kind of [table](https://github.com/nsidc/earthaccess/pull/1011#issuecomment-3228417640) @jhkennedy used to remind everyone of the approvals before they get marked as stale.

mfisher87:
:100: Perhaps that table could be pre-populated from CODEOWNERS in the PR description?

docs/governance/governance-process.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. GitHub issues can be labeled `needs: decision` by anyone to kick off the decision committee process
+2. A PR will be automatically created with a blank decision record document and a member of the decision committee will be assigned at random to shepherd the decision through the processes
+3. The assignee will create a _first_ draft of the decision record document, summarizing the relevant issues, discussion, and pull requests, and then mark it as ready for review
+4. The decision committee will use PR review mechanics to discuss the decision and suggest changes or edit the decision record to reflect the decision and discussion
+   - Decision committee members endorse a decision as written by approving the PR
+   - Decision committee members can reject a decision _as written_ by requesting changes in the PR
+   - Decision committee member can express neutral discussion points by leaving PR comments 
+5. Once N decision committee members endorse a decision (approve the PR), the decision is considered "decided" and the PR should be merged to record the decisions
```
itcarroll, line 20: When we can return to this, I think it will be important to remember that we created ambiguity in #1047 between "voting" and "approving". We had the good idea of creating a table to track preferences, but it was unclear (at least to me) whether that was a "straw poll" or the vote of record. We merged the PR with a single "approval", so in effect the table was the official record. Or we went with N==1, which seems insufficient. Let's ensure the process eliminates ambiguity here.

mfisher87: Great call, I totally forgot about this :laughing:  I was thinking of the table as the vote of record, which I shouldn't have.

PR #1033, opened 2025-06-23
Removed `has_granules` and `include_granule_counts` option default values from CMR collection queries

### LINKED ISSUE #884: [BUG] Inconsistent results from search_datasets

### Current Behavior

I am trying to search the NASA [OPERA](https://www.jpl.nasa.gov/go/opera/) data products using earthaccess. Using the `OPERA` keyword to search datasets can retrieve the eight OPERA data products correctly. However, using the `*` keyword to search only returns seven OPERA data products. The [OPERA_L3_DSWX-S1_V1](https://search.earthdata.nasa.gov/search/granules/collection-details?p=C2949811996-POCLOUD&pg[0][v]=f&pg[0][gsk]=-start_date&q=DSWx&tl=1732851361.392!3!!&lat=-0.140625) data product is missing from the search results. This is problematic because I rely on using the `*` keyword to retrieve the entire NASA Earth Data as a CSV file for the [NASA-Earth-Data](https://github.com/opengeos/NASA-Earth-Data) repo.

Relevant repo: [OPERA_Applications](https://github.com/OPERA-Cal-Val/OPERA_Applications) @alhandwerger

### Expected Behavior

Using the `*` keyword to search should return all eight OPERA data products.

### Steps To Reproduce

This code snippet can retrieve the eight OPERA data products correctly. 
```python
datasets = earthaccess.search_datasets(keyword="OPERA")
for dataset in datasets:
    print(dataset["umm"]["ShortName"])
```
image

This code snippet only returns seven OPERA data products. The `OPERA_L3_DSWX-S1_V1` data product is missing. 
```python
datasets = earthaccess.search_datasets(keyword="*")
for dataset in datasets:
    if "OPERA" in dataset["umm"]["ShortName"]:
        print(dataset["umm"]["ShortName"])
```
image

### Environment

```markdown
- OS: Manjaro Linux
- Python: 3.12
```

### Additional Context

_No response_

Sherwin-14:
I have made the changes as discussed [here](https://github.com/nsidc/earthaccess/issues/884). Resolves #884 .

----
📚 Documentation preview 📚: https://earthaccess--1033.org.readthedocs.build/en/1033/


CHANGELOG.md
```diff
@@ -49,6 +49,11 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 ### Fixed
 - Corrected Harmony typo in notebooks/Demo.ipynb([#995](https://github.com/nsidc/earthaccess/issues/995))([stelios-c](https://github.com/stelios-c))
 
+- Removed `self.params["has_granule"]` and `self.params["include_granule_counts"]`
+  from `DataCollections` to avoid future gotchas.
+  ([#884](https://github.com/nsidc/earthaccess/issues/884))
+  ([@Sherwin-14](https://github.com/Sherwin-14))
+
```
chuckwondo: Under the "Unreleased" section, I see that there are 2 "Added" sub-sections, so I'd like to have you do the following:

- Combine the 2 "Added" sub-sections into one
- Order the sub-sections according to what is suggested here: https://common-changelog.org/#24-change-group
- Move your new entry under the "Removed" sub-section and place it as the first item since breaking changes should be listed first within a sub-section.
- Add a `**Breaking:** ` prefix to your new entry, as described here: https://common-changelog.org/#244-prefixes

```
- **Breaking:** Removed `has_granules=true` and `include_granule_counts=true`
  as default parameters upon creation of a `DataCollections` instance.
  ([#884](https://github.com/nsidc/earthaccess/issues/884))
  ([@Sherwin-14](https://github.com/Sherwin-14))

```

tests/unit/test_collection_queries.py
```diff
@@ -59,8 +59,6 @@ def test_querybuilder_can_handle_has_granules():
     assert not query.params["has_granules"]
     query = DataCollections().has_granules(True)
     assert query.params["has_granules"]
-    query = DataCollections().has_granules(None)
-    assert "has_granules" not in query.params
```
chuckwondo: Let's **keep these lines**.

tests/unit/test_collection_queries.py
chuckwondo, line 1: Let's add a new test to this file:

```python
def test_no_default_params():
    query = DataCollections()
    assert len(query.params) == 0
```

mfisher87:
How do we feel about this change to the title?

tests/unit/test_collection_queries.py
```diff
@@ -60,8 +65,7 @@ def test_querybuilder_can_handle_has_granules():
     query = DataCollections().has_granules(True)
     assert query.params["has_granules"]
     query = DataCollections().has_granules(None)
-    assert "has_granules" not in query.params
-
+    assert query.params["has_granules"] is None
```
chuckwondo: Ah, now I see why you had to change this to get the test to pass again. It's because it revealed an existing bug in the `DataCollections.has_granules` method that we should fix. The problem is due to these lines in that method:

```python
        if has_granules is None and "has_granules" in self.params:
            del self.params["has_granules"]
        else:
            self.params["has_granules"] = has_granules
```

Since we have removed the line in the `__init__` method that sets an initial value for the `has_granules` parameter, when we call `has_granules(None)`, `"has_granules"` is _not_ in `self.params`, so the `else` branch is taken, which should _not_ be the case when `has_granules` is `None`.

In other words, the logic should _never_ explicitly set the value to `None`.

Therefore, the lines above in `DataCollections.has_granules` should be modified like so:

```python
        if has_granules is not None:
            self.params["has_granules"] = has_granules
        elif "has_granules" in self.params:
            del self.params["has_granules"]
```

In addition, revert the line above in this test function to its previous value, because we want to make sure the key `"has_granules"` is not in the params dict.

chuckwondo (changes requested):
We're getting there. I added another comment. Also, it looks like there's some formatting you'll need to address: see https://results.pre-commit.ci/run/github/399867529/1750751061.-QrOBrC9SJG-O5yW-7OzIw

chuckwondo:
> How do we feel about this change to the title?

Thanks for some fine-tuning. I further refined it to be explicit about removing _default_ values.

chuckwondo (changes requested):
@Sherwin-14, this looks good. Thank you. One last request to see if we can resolve the pre-commit failure.

When I run `pre-commit run -a`, I am able to reproduce the pre-commit error. After I run `uv lock`, which updates the `uv.lock` file, running `pre-commit run -a` no longer produces the error.

Please run `uv lock` and commit/push the changes to the `uv.lock` file to see if that clears up the CI pre-commit error. If not, no worries. I'll approve anyway, as this has been an ongoing problem.

chuckwondo:
@Sherwin-14, I ran the integration tests and they are failing due to the recordings in the following files:

- `tests/integration/fixtures/vcr_cassettes/TestServices.test_services.yaml`
- `tests/unit/fixtures/vcr_cassettes/TestResults.test_collections_less_than_2k.yaml`
- `tests/unit/fixtures/vcr_cassettes/TestResults.test_collections_more_than_2k.yaml`
- `tests/unit/fixtures/vcr_cassettes/TestServices.test_service_results.yaml`

You must remove `has_granules=true&include_granule_counts=true&` from various `uri` values in the files above.

It looks like you removed some, but not all.

Sherwin-14:
> You must remove `has_granules=true&include_granule_counts=true&` from various `uri` values in the files above.
> 
> It looks like you removed some, but not all.

I have made the changes in all possible files, please let me know if I missed something.

Sherwin-14:
> Thanks @Sherwin-14! The integration tests still fail, but now it's only because of tests for an "on prem" collection that no longer exists. I'm going to open a separate issue to address this so we can land this PR.

I opened a discussion regarding integration tests [here](https://earthaccess.zulipchat.com/#narrow/channel/480557-general/topic/Failing.20Integration.20Tests/with/525750237). I will need your guidance there :)

chuckwondo:
@mfisher87, @betolink, or @jhkennedy, this is a small PR, if any of you want to take a quick look before I merge it.

The integration tests failed only because they attempt to query granules from obsolete collections, for which I opened a separate issue: #1036

betolink (dismissed):
Looks good, are the cassettes from VCR autogenerated?

chuckwondo:
> Looks good, are the cassettes from VCR autogenerated?

Yes, they were generated a while back. Technically, we should regenerate them rather than hand-modifying the `uri` values, but this is such a minor change to the parameters that the difference in results (if any) would not affect the tests.

chuckwondo:
@mfisher87, after #1038 landed, int. tests are better, but still flaky. At least now, re-running failed int. tests tends to succeed, but I'm still not a fan of the fact that we do such testing across numerous collections, primarily because of the flakiness, but also because this feels too much like testing things that are not our responsibility to test. I feel we need to find a better approach.

chuckwondo:
> @mfisher87, after #1038 landed, int. tests are better, but still flaky. At least now, re-running failed int. tests tends to succeed, but I'm still not a fan of the fact that we do such testing across numerous collections, primarily because of the flakiness, but also because this feels too much like testing things that are not our responsibility to test. I feel we need to find a better approach.

@mfisher87, it appears that our failure "threshold" is no longer in place. What happened to that? We're now failing the workflow even if there's only 1 failure.

mfisher87:
> @mfisher87, after #1038 landed, int. tests are better, but still flaky. At least now, re-running failed int. tests tends to succeed, but I'm still not a fan of the fact that we do such testing across numerous collections, primarily because of the flakiness, but also because this feels too much like testing things that are not our responsibility to test. I feel we need to find a better approach.

I definitely agree a better approach is needed. I do think there's value in this type of testing, because it notifies us of  changes to external systems that we would need to respond to. But I don't see much value in running those types of tests on every PR. Perhaps instead they should run monthly, or something like that? If there's a breakage that impacts many users and needs to be addressed urgently, users will let us know. Breakages with smaller impacts may slip through the cracks, silently impacting users who don't open reports on github, without this style of testing.

What do you think about bringing this discussion to Zulip or an issue?

chuckwondo:
> What do you think about bringing this discussion to Zulip or an issue?

Yes, that sounds good.

mfisher87:
> @mfisher87, it appears that our failure "threshold" is no longer in place. What happened to that? We're now failing the workflow even if there's only 1 failure.

Correct, we intentionally removed that because the threshold applied to all integration tests, not just the flaky ones, so we had some tests that failed 100% of the time slipping by. That was the motivation for changing our approach from testing randomly-selected collections and allowing some failures (because we would often select broken collections) with this new approach of testing a smaller number of very popular collections, which presumably would be broken less frequently.

mfisher87:
Context here: https://github.com/nsidc/earthaccess/issues/215

mfisher87:
Thanks @Sherwin-14 ! :heart:

PR #1039, opened 2025-06-30
Added `status` method for checking status of NASA APIs

### LINKED ISSUE #161: Add feture to check the status of NASA APIs
Currently there is no way of checking the status of the APIs used by the library, it would be useful having a `status()` function (maybe in the module __init__) to verify that CMR and EDL are working and available.

Sherwin-14:
I've added the status function as discussed [here](https://github.com/nsidc/earthaccess/issues/161). Resolves #161 

----
📚 Documentation preview 📚: https://earthaccess--1039.org.readthedocs.build/en/1039/

### DIFF
--- a/.github/workflows/test-mindeps.yml
+++ b/.github/workflows/test-mindeps.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v4.1.1
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6.3.1
+        uses: astral-sh/setup-uv@v6.4.3
         with:
           enable-cache: true
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,14 +30,14 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.2
+    rev: v0.12.7
     hooks:
       - id: ruff
         args: ["--fix", "--exit-non-zero-on-fix"]
       - id: ruff-format
 
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: "0.7.19"
+    rev: "0.8.4"
     hooks:
       # Asserts that the lockfile is valid for pyproject.toml.
       # If it fails, update the lockfile by running:


earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
+
+    Examples:
+        ```python
+        result = earthaccess.status(system=UAT)
+        ```
+    """
+    urls = {
+        PROD: "https://status.earthdata.nasa.gov/api/v1/statuses",
+        UAT: "https://status.uat.earthdata.nasa.gov/api/v1/statuses",
+    }
```
chuckwondo: Let's move the URLs into the `System` class itself.

In `earthaccess/system.py`, consider adding new types:

```python
StatusURL = NewType("StatusURL", str)
StatusApiURL = NewType("StatusApiURL", str)
```

Then, add corresponding attributes to the `System` class, and a corresponding arguments to the 2 calls at the end of the file that assign values to the `PROD` and `UAT` instances.

The status URLs are the URLs intended for browser rendering:

- https://status.earthdata.nasa.gov/
- https://status.uat.earthdata.nasa.gov/

and the status API URLs are for programmatic use:

- https://status.earthdata.nasa.gov/api/v1/statuses
- https://status.uat.earthdata.nasa.gov/api/v1/statuses

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
 
 
+def status(system: System = PROD) -> Dict[str, str] | None:
+    """Gets the status of NASA's Earthdata services.
+
+    Parameters:
+        system: The Earthdata system to access, defaults to PROD.
+
+    Returns:
+        A dictionary containing the status of Earthdata services or None.
```
chuckwondo:
```suggestion
        A dictionary containing the status of Earthdata services.
```

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
 from .search import CollectionQuery, DataCollections, DataGranules, GranuleQuery
 from .store import Store
-from .system import PROD, System
+from .system import PROD, UAT, System
 from .utils import _validation as validate
 
 logger = logging.getLogger(__name__)
 
 
+def status(system: System = PROD) -> Dict[str, str] | None:
```
chuckwondo:
```suggestion
def status(system: System = PROD) -> Mapping[str, str]:
```

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
+    Returns:
+        A dictionary containing the status of Earthdata services or None.
+
+    Raises:
+        Exception: The request to the Earthdata status API failed.
+
+    Examples:
+        ```python
+        result = earthaccess.status(system=UAT)
+        ```
```
chuckwondo:
```suggestion
        >>> earthaccess.status()  # doctest: +SKIP
        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}
        >>> earthaccess.status(earthaccess.UAT)  # doctest: +SKIP
        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}
```

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
+
+    Parameters:
+        system: The Earthdata system to access, defaults to PROD.
+
+    Returns:
+        A dictionary containing the status of Earthdata services or None.
+
+    Raises:
+        Exception: The request to the Earthdata status API failed.
+
```
chuckwondo: We won't (intentionally) raise any exceptions:

```suggestion
```

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
+                name = name.replace("(PROD)", "").strip()
+
+            if name in earthdata_services:
+                result[name] = status["status"]
+
+        return result
+
+    except Exception as e:
+        logger.error(f"Failed to retrieve Earthdata service status for {system}: {e}")
+        return None
```
chuckwondo:
```suggestion
    services = ("Earthdata Login", "Common Metadata Repository")
    statuses = {service: "Unknown" for service in services}
    
    try:
        with requests.get(system.status_api_url) as r:
            r.raise_for_status()

            for entry in r.json().get("statuses", []):
                name = entry.get("name", "")
                
                if (service := next(filter(name.startswith, services), None)):
                    statuses[service] = entry.get("status", "Unknown")
    except (json.JSONDecodeError, requests.exceptions.RequestException):
        logger.error(
            f"Unable to retrieve Earthdata service status for {system}."
            "  Try again later, or visit {system.status_url} to check service status."
        )

    return statuses
```

tests/integration/test_api.py
```diff
@@ -78,6 +79,17 @@ def test_download(tmp_path, selection, use_url):
     assert all(Path(f).exists() for f in files)
 
 
+@pytest.mark.parametrize("system", [PROD, UAT])
+def test_earthdata_status(system):
+    result = earthaccess.status(system)
```
chuckwondo: We don't want to make real network calls in unit tests. For HTTP testing, make use of the `responses` library. For guidance, see `tests/unit/test_auth.py`, but if you need further assistance, let me know.

We want to test the following responses, at a minimum:

- happy path: we get a "normal" response (with status code 200)
- status 200, but malformed json in the response (should return the "defaults" mapping defined in the `status` function, due to json decoding error)
- non-200 status, say 500, (should also return "defaults", due to raising for status)

CHANGELOG.md
```diff
@@ -43,6 +43,9 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
   environment variable `EARTHDATA_TOKEN`
   ([#484](https://github.com/nsidc/earthaccess/issues/484))
   ([@kgrimes2](https://github.com/kgrimes2))
+- Added `status` method to check status of NASA APIs
```
chuckwondo:
```suggestion
- Added top level `status` function to check the statuses of NASA Earthdata services
```

earthaccess/api.py
```diff
@@ -22,12 +22,57 @@
...
+
+    Returns:
+        A dictionary containing the status of Earthdata services or None.
+
+    Raises:
+        Exception: The request to the Earthdata status API failed.
+
+    Examples:
+        ```python
+        result = earthaccess.status(system=UAT)
```
mfisher87:
```suggestion
        nasa_status = earthaccess.status(system=UAT)
        if nasa_status["Earthdata Login"] != "OK" or nasa_status["Common Metadata Repository"] != "OK":
            raise Exception(
                "NASA APIs unavailable, please try again later."
                " You can check status at https://status.earthdata.nasa.gov"
            )
```

Nitpick; I don't like `result` as a variable name and we're using it for a ton of other things in this doc. 

I think it'd be nice to show how it might be used as well.

earthaccess/api.py
mfisher87: Nitpick: This is one of my least favorite ways to use the walrus operator. The assignment happens after the usage! I'd personally sacrifice a bit of conciseness for readability here, but that's just me.

earthaccess/api.py
mfisher87: Nice! I think we should be doing more doctests (and agree this one should be skipped since it does an HTTP request). This comment came through while I still had a review in progress, so I have some related thoughts in a separate comment, to the tune of:

I like the doctest approach, but also like providing a bit more hand-holding of how a user might use this in a real program.

itcarroll (changes requested):
I'd like to request a (new?) page in the docs under "User Guide" for this new top-level function, please.

earthaccess/api.py
chuckwondo: > Nitpick: This is one of my least favorite ways to use the walrus operator. The assignment happens after the usage! I'd personally sacrifice a bit of conciseness for readability here, but that's just me.

I agree that the readability of usage appearing before assignment is less than ideal, but don't blame the lovely walrus operator :-)

Such a readability issue is typical of _any_ type of collection comprehension even in the absence of the walrus operator anywhere in the comprehension. Specifically, note that the element expression (or part of it) typically refers to a variable defined in the `for` subexpression, which appears _after_ the element expression.

Granted, in this case, the walrus operator is one line farther away, but I'd argue that refactoring this for readability would be due to use of a comprehension in general, not specifically for the walrus operator (which perhaps just makes for a stronger case against the comprehension in general).

Either way, I have no issue if someone wants to refactor this for improved readability. (I just wanted to defend the lonely walrus operator.)

mfisher87: Understood, but I think that mechanism is fairly well-understood by Python developers! My concern is that the walrus operator is new and not often used, and I think that it may introduce more confusion when used in this way, as opposed to when it's used more straightforwardly, for example in a conditional expression (and I really like the walrus operator in this case). I don't feel strongly about this, so if a third person feels like being a tie-breaker we can go by their preference ;)

chuckwondo: I'm in agreement with you. I'm just saying, as many times as I've written comprehensions, when I look at them, I _still_ often find myself looking _above_ for where a variable was declared, only to realize it's _after_ the element expression, in the `for` subexpression.

Admittedly, I perhaps have an excessive fondness for comprehensions (in spite of what I said above), but I have no qualms if we don't use one here, so we can avoid the additional reduction in readability caused by my throwing the walrus operator into the mix.

chuckwondo: @mfisher87, I removed the comprehension in the suggested code change. Let me know if you find that more readable. It seems better to me.

mfisher87: I think that's a good readability improvement, thanks!  :heart:  I also love comprehensions myself... perhaps out of that love I don't even notice I often read them backwards :laughing: Maybe it's because I kind of read them in english, but when an operator symbol is introduced suddenly I think order matters :laughing: 

Maybe I'm being a bit overly critical about readability. I feel readability is the most important thing so that tends to be what I think about the most :grin:

chuckwondo: > Maybe I'm being a bit overly critical about readability. I feel readability is the most important thing so that tends to be what I think about the most 😁

I don't think you're being overly critical about readability. I agree that it is very important, and thanks for the reminder. I tend to let my preference for a more functional style of programming and my disdain for `for` loops blind me. 😬

earthaccess/api.py
mfisher87: I hate to go back and forth on this, but I think our return types should be exact -- i.e. this function returns a `dict`, a more specific case of `Mapping`.

I like the idea of having looser, more general argument types that describe _what we do_ with the thing passed in, i.e. if we do string lookups on an argument object, we can accept any `Mapping` object from the user.

But I think it should be the opposite for return types -- tell the user exactly what they're getting from the code they don't control.

chuckwondo: I have to disagree with you here. The return type should indicate the "interface" we're returning to the user, thus giving us the freedom to change the implementation without introducing a breaking change.

Of course, in this specific case, the impact is certainly not critical, but generally speaking, this is my preference.

mfisher87: My concern is that we have no idea what the user is doing with the returned object. If we type our return loosely, then changing the object we return would be a breaking change even though we don't have to update the annotation. Having a concrete annotation wouldn't change whether the change is breaking or not; it just changes what we communicate to the user, and even helps users recognize and adapt to breaking changes if they use a type checker!

The official Python typing best practices recommend this approach:

https://typing.python.org/en/latest/reference/best_practices.html#arguments-and-return-types

chuckwondo: > My concern is that we have no idea what the user is doing with the returned object. If we type our return loosely, then changing the object we return would be a breaking change even though we don't have to update the annotation.

Sure, we don't know what the user will do with the returned value, but that's the point of specifying an interface. 

By specifying an interface, our "contract" to the user is that we will always return a value that implements the interface. Thus, if the user chooses to rely upon our returning a specific implementation of that interface, they do so at their own risk because the contract makes no promise as to the specific implementation returned.

In other words, by specifying an interface, changing the concrete type we return does not represent a breaking change. The user should not rely on the concrete type, nor should they need to. The interface specifies the behavior users can expect without any need for them to care about how that behavior is implemented.

Only if we were to decide to change which interface we return would we be introducing a breaking change.

> Having a concrete annotation wouldn't change whether the change is breaking or not; it just changes what we communicate to the user, and even helps users recognize and adapt to breaking changes if they use a type checker!

Yes, it would. By specifying an "interface," we can change the implementation of that interface without having to communicate anything at all to the user, because we have not broken our contract.

Further, type checking remains completely unaffected because the type signature remains unchanged.

> The official Python typing best practices recommend this approach:
> 
> https://typing.python.org/en/latest/reference/best_practices.html#arguments-and-return-types

Unfortunately, that "best practice" guide doesn't explain why it recommends the following:

> For return values, prefer concrete types (list, dict, etc.) for concrete implementations.

I _strongly_ disagree with that bit, precisely because of what I explained above.

chuckwondo: However, in this particular case, I would not be utterly opposed to annotating the return type with `dict` rather than `Mapping` because this function is isolated and not on any sort of "critical path" of logic, where we really would have any concern over the possibility of breaking changes.

In other areas, however, I will be less inclined to concede this point.

jhkennedy: Well, since there's a bit of discussion, I'm going to weigh in. 

I also prefer to specify what we're actually returning (`dict[str, str]`) over a higher-level generic type. For example, I'd be pretty annoyed by an API that type-hinted `float` but returned `int`, and this feels the same to me.

That said, I also think, in this particular case, the discussion is a bit moot as there is only one standard mapping type, the `dict`.

> A [mapping](https://docs.python.org/3/glossary.html#term-mapping) object maps [hashable](https://docs.python.org/3/glossary.html#term-hashable) values to arbitrary objects. Mappings are mutable objects. There is currently only one standard mapping type, the dictionary.

If we were returning a Mapping-like object other than `dict`, I generally would expect us to specify that object specifically, as the additional or changed methods/attributes from dict would be valuable (otherwise we'd just use a dict). Using `Mapping` for output gives users no valuable information other than "I expect you to use it like a dict". Anywhere we'd take advantage of those additional or changed methods/attributes in an object that's passed in, we'd need to type that class specifically, and we wouldn't be telling users where they get said type.

So I can see using `Mapping` as an input type, but I think it's actually harmful as an output type.

mfisher87: I think I get you -- I also read a StackOverflow post where you were having a similar conversation with another commenter. I want to change your mind on this, but I respect that you're an expert, and we can have different opinions informed by our distinct experiences and that's OK. I'm not completely closed to having my mind changed, but I think I understand why this best practice is defined as it is. It's a shame the document doesn't explain itself.

> Unfortunately, that "best practice" guide doesn't explain why it recommends the following:
>> For return values, prefer concrete types (list, dict, etc.) for concrete implementations.

If we don't come to agreement and nobody else weighs in on this, I'd like to resolve this discussion by deferring to the documented best practice, if that's OK with you!

My attempt to change your mind:

semver.org does not provide an official definition of "breaking change" :sob: This feels like another missed opportunity.

The way I define "breaking change" is "doing something that can cause code in the wild to cease functioning". Or "a change that will require **any** users to make a change in response". I think our definitions may be different. It would be helpful if you could provide your definition!

To me, the purpose of classifying changes as "breaking" or "not breaking" is to notify users when they might need to make those changes in response to software releases. And it gives us a framework to understand how we can aid those users in adapting.

So, for example, if we return an object which is an instance of `Foo` which conforms to the `Mapping<str, str>` type, but that object also has some other methods on it like `.bar()`, users may be using `.bar()` in their code. If we changed from returning a `Foo` to returning a `dict`, users who use `.bar()` on the object we return will have their code break.

Only by annotating the return with a concrete type can we communicate to the user that we are going to break their code.

Because our code is a black box to many users, I feel the logical conclusion of all of this is that we should tell the user exactly what _we_ need from their inputs (i.e. the loosest possible type as determined by what we're doing inside the black box, where we have control), and tell them exactly what _they_ should expect to get out on the other side of the black box, as what they do with that thing is outside of our control, and changing that thing means they may need to handle it differently.

This means that we would be potentially subject to more frequent changes in what we're advertising the user to expect from our code, but with the benefit that we're giving them much more information about what to expect and more effectively protecting them from frustration and breakage.

To me, the tradeoff is a cost to smaller number of people (maintenance burden of having a definition of "breaking change" that is more strict) with a payoff for a larger number of people (greater knowledge of how our changes will affect their code).

What do you think?

chuckwondo: I think we'll just have to agree to disagree here (although to @jhkennedy's point, in Python, this point is indeed somewhat moot given that `dict` is effectively the only mapping type anybody deals with).

Given that this particular case is very low impact, I'll concede to being outvoted.

If this topic should arise again, we can open a discussion, since this is more about the general topic, not something blocking this particular piece of code.

mfisher87: Sounds like a plan. Although I do still want to know if you agree with my definition of "breaking change", and if not, what definition you're using!

earthaccess/api.py
```diff
@@ -22,12 +22,52 @@
...
 from .search import CollectionQuery, DataCollections, DataGranules, GranuleQuery
 from .store import Store
-from .system import PROD, System
+from .system import PROD, UAT, System
 from .utils import _validation as validate
 
 logger = logging.getLogger(__name__)
 
 
+def status(system: System = PROD) -> Mapping[str, str]:
```
mfisher87:
```suggestion
def status(system: System = PROD) -> dict[str, str]:
```

Sorry @Sherwin-14 to flip flop you here :) See the other thread for context :)

earthaccess/api.py
chuckwondo: I classify a breaking change as a change that breaks (changes) the "contract" defined by our public API. The public API communicates to the user what we expect _from_ the user and what we provide _to_ the user. There are no guarantees to the user for anything outside of the public API.

If _we_ break (change) the contract (e.g., change the type of an input or an output in a non-backwards compatible way), it's a breaking change because it has the potential to break user code that fully adheres to the previous contract.

If the _user_ breaks the contract (e.g., passes the wrong type of value for an input, or treats an output as a type that is not compatible with what's defined in the public API), that's a bug in their code (if it causes their code to break or behave unexpectedly).

Conversely, when a user relies on something that is _not_ part of the public API, they do so at their own risk, meaning that if we make a change to non-public code that causes a break in the user's code that is _directly_ relying on such non-public code, that's the user's problem. Thus, it is not classified as a breaking change.

Not to go back down the rabbit hole, but this is why I'm in favor of return types as interfaces, because the interfaces are part of that public API contract, but the concrete implementations are not (in cases where we expose interfaces as part of the public API, rather than concrete types), meaning that we can change the implementation without breaking the contract (as long as the implementation still implements the interface), and thus not introduce a breaking change.

mfisher87: Thanks for providing some more detail about your position. I have lots of thoughts and would like to continue the discussion, but I don't want to dive deeper here.

tests/integration/test_api.py
Sherwin-14: Since both PROD and UAT return different JSON values, should I use them literally or only use the `statuses` field?

mfisher87: I'm not sure I fully understand the question, can you give examples? I was thinking that we should normalize the response so the returned dict from earthaccess always looks the same, and Chuck's code implements that!

Sherwin-14: > I'm not sure I fully understand the question, can you give examples? I was thinking that we should normalize the response so the returned dict from earthaccess always looks the same, and Chuck's code implements that!

I was talking about the tests
```py
@pytest.mark.parametrize("system", [PROD, UAT])
@responses.activate
def test_earthdata_status_happy_path(system):
    responses.add(
        responses.GET,
        system.status_api_url,
        json=
        status=200,
    )
    ...
```
    
   I was asking about what to set in the json field, both `PROD` and `UAT` return different values when you hit the respective urls but they both have the `statuses` field in common.
   
   I was wondering if I should *just* use the statuses field or use the whole response for `PROD` and `UAT`.

mfisher87: I would use a whole real response if it were me!

chuckwondo: @Sherwin-14, parametrize the json value along with the system value, using a full json response. This will allow you to provide a PROD response value that is different from a UAT response value.

docs/user_guide/status.md
```diff
@@ -0,0 +1,20 @@
...
+
+Earthaccess relies on the Common Metadata Repository (CMR) API and Earthdata Login (EDL) service to provide functionality. You can check the status of these NASA Earthdata services manually or programmatically.
+
+## Manual Status Check
+
+You can check the status of NASA Earthdata services manually for [PROD](https://status.earthdata.nasa.gov/) and [UAT](https://status.uat.earthdata.nasa.gov/) systems.
+
+## Programmatic Status Check
+
+You can check the status of EarthData services  programmatically this way.
```
itcarroll:
```suggestion
You can check the status of Earthdata services  programmatically this way.
```

docs/user_guide/status.md
```diff
@@ -0,0 +1,20 @@
...
+
+```py
+import earthaccess
+
+nasa_status = earthaccess.status(system=UAT)
```
itcarroll:
```suggestion
nasa_status = earthaccess.status()
```
The example does not define `UAT`, and I think that's kind of "low-level" to show in this example of basic use. I didn't look at what happens if you use the default `system`, but would it be good enough for this example?

itcarroll (changes requested):
Thank you for the docs page!

docs/user_guide/status.md
Sherwin-14: I added the example for `UAT` assuming that the user already knows about `UAT` and `PROD` env.

docs/user_guide/status.md
```diff
@@ -0,0 +1,20 @@
+# Status
```
mfisher87, line 1: 🤩

tests/integration/test_api.py
```diff
@@ -32,6 +34,89 @@
...
+        service["name"].replace(" (PROD)", "").replace(" (UAT)", "").strip(): service[
+            "status"
+        ]
+        for service in nasa_statuses[system]["statuses"][-2:]
+    }
+
+    responses.add(
+        responses.GET,
+        system.status_api_url,
+        json=nasa_statuses[system],
```
mfisher87: This is accessing the `nasa_statuses` object from the module scope; we should be accessing it from the parameterization arguments (currently, `kwargs` is unused). I think we should restructure this a bit so that we only use the test function's parameters. E.g. (pseudocode, forgive me if it's not quite right, but I think it shows what I mean):

```python
@pytest.mark.parametrize("system,nasa_response,expected", [
    (
        PROD,
        { ... nasa response dict here ... },
        {"Earthdata Login": "OK", "Common Metadata Repository": "OK"},
    ),
    (
        UAT,
        { ... nasa response dict here ... }
        {"Earthdata Login": "OK", "Common Metadata Repository": "OK"},
    ),
])
def test_earthdata_status_happy_path(system, nasa_response, expected):
    responses.add(
        ...,
        json=nasa_response,
        ...,
    )
    actual = earthaccess.status(system)
    assert actual == expected
```

tests/integration/test_api.py
```diff
@@ -32,6 +34,89 @@
...
+        status=200,
+    )
+
+    result = earthaccess.status(system)
+    assert isinstance(result, dict)
+    assert result == statuses
+
+
+@pytest.mark.parametrize("system", [PROD, UAT])
```
mfisher87: We don't need to parametrize this one as there is no difference in the response data between prod and UAT!

```suggestion
```

tests/integration/test_api.py
```diff
@@ -32,6 +34,89 @@
...
+        body='{"x": 5,}',
+        status=200,
+    )
+
+    result = earthaccess.status(system)
+    assert isinstance(result, dict)
+    assert result == statuses
+
+
+@pytest.mark.parametrize("system", [PROD, UAT])
```
mfisher87:
```suggestion
```

tests/integration/test_api.py
Sherwin-14: On removing this, the tests would start failing as `system` is used to access corresponding `status_api_url`.
```py
 responses.add(
        responses.GET,
        system.status_api_url,
        body='{"x": 5,}',
        status=200,
 )
```
If we remove this, then should we hardcode the `PROD`  url inplace of `status_api_url` ?

docs/user_guide/status.md
```diff
@@ -0,0 +1,20 @@
...
+
+You can check the status of Earthdata services  programmatically this way.
+
+```py
+import earthaccess
+
+nasa_status = earthaccess.status()
+
+if any(service_status != 'OK' for service_status in nasa_status.values()):
+   raise Exception("NASA APIs unavailable, please try again later.")
```
chuckwondo: @mfisher87, what do you think about adding an optional keyword parameter to the `status` function, named something like `raise_on_outage` (boolean, default `False`) that will raise an exception (perhaps something like`ServiceOutage`) when any status is not `"OK"` and `raise_on_outage` is `True`?

earthaccess/api.py
```diff
@@ -28,6 +29,42 @@
 logger = logging.getLogger(__name__)
 
 
+def status(system: System = PROD) -> Dict[str, str]:
+    """Gets the status of NASA's Earthdata services.
```
chuckwondo:
```suggestion
    """Get the statuses of NASA's Earthdata services.
```

earthaccess/api.py
```diff
@@ -28,6 +29,42 @@
...
+
+    Parameters:
+        system: The Earthdata system to access, defaults to PROD.
+
+    Returns:
+        A dictionary containing the status of Earthdata services.
```
chuckwondo:
```suggestion
        A dictionary containing the statuses of Earthdata services.
```

earthaccess/api.py
```diff
@@ -28,6 +29,42 @@
...
+
+            for entry in r.json().get("statuses", []):
+                name = entry.get("name", "")
+
+                if service := next(filter(name.startswith, services), None):
+                    statuses[service] = entry.get("status", "Unknown")
+    except (json.JSONDecodeError, requests.exceptions.RequestException):
+        logger.error(
+            f"Unable to retrieve Earthdata service status for {system}."
+            f"Try again later, or visit {system.status_url} to check service status."
```
chuckwondo:
```suggestion
            f"Unable to retrieve Earthdata service statuses for {system}."
            f"Try again later, or visit {system.status_url} to check service statuses."
```

earthaccess/system.py
```diff
@@ -8,15 +8,34 @@
 
 CMRBaseURL = NewType("CMRBaseURL", str)
 EDLHostname = NewType("EDLHostname", str)
+StatusURL = NewType("StatusURL", str)
+StatusApiURL = NewType("StatusApiURL", str)
+
+STATUS_PROD_URL = "https://status.earthdata.nasa.gov/"
+STATUS_UAT_URL = "https://status.uat.earthdata.nasa.gov/"
+STATUS_API_PROD_URL = "https://status.earthdata.nasa.gov/api/v1/statuses"
+STATUS_API_UAT_URL = "https://status.uat.earthdata.nasa.gov/api/v1/statuses"
```
chuckwondo: Let's just hard-code these in the only places they're used below in the declarations of the `PROD` and `UAT` variables. There's no need to make constants here, as their values are used only once, and they should be accessed via the `PROD` and `UAT` instances rather than directly via constants.

chuckwondo (changes requested):

docs/user_guide/status.md
mfisher87: I think that sounds good, nice convenience for users who may not care about the structure of the data this function returns, they just want to stop their code in case of outage.

If we do that let's update our docs page to use that as the example, since I think it may be the most common case!

earthaccess/api.py
```diff
@@ -28,6 +29,45 @@
 logger = logging.getLogger(__name__)
 
 
+def status(
+    system: System = PROD, raise_on_outage: Optional[bool] = False
```
chuckwondo:
```suggestion
    system: System = PROD, raise_on_outage: bool = False
```

earthaccess/api.py
```diff
@@ -28,6 +29,45 @@
...
+) -> Dict[str, str]:
```
chuckwondo:
```suggestion
) -> dict[str, str]:
```

earthaccess/api.py
```diff
@@ -28,6 +29,45 @@
...
+        raise_on_outage: If True, raises exception on errors or outages.
+
+    Returns:
+        A dictionary containing the statuses of Earthdata services.
+
+    Examples:
+        >>> earthaccess.status()  # doctest: +SKIP
+        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}
+        >>> earthaccess.status(earthaccess.UAT)  # doctest: +SKIP
+        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}
```
chuckwondo, line 46:
```suggestion
        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}

    Raises:
        ServiceOutage: if at least one service status is not `"OK"`
```

earthaccess/api.py
```diff
@@ -28,6 +29,45 @@
...
+    """
+    services = ("Earthdata Login", "Common Metadata Repository")
+    statuses = {service: "Unknown" for service in services}
+    msg = f"Unable to retrieve Earthdata service statuses for {system}.Try again later, or visit {system.status_url} to check service statuses."
```
chuckwondo:
```suggestion
    msg = (
        f"Unable to retrieve Earthdata service statuses for {system}."
        "  Try again later, or visit {system.status_url} to check service statuses."
    )
```

earthaccess/api.py
```diff
@@ -28,6 +29,45 @@
...
+                name = entry.get("name", "")
+
+                if service := next(filter(name.startswith, services), None):
+                    statuses[service] = entry.get("status", "Unknown")
+                    if raise_on_outage and statuses[service] != "OK":
+                        raise ServiceOutage(msg)
+    except (json.JSONDecodeError, requests.exceptions.RequestException):
+        logger.error(msg)
+
+    return statuses
```
chuckwondo:
```suggestion
    try:
        with requests.get(system.status_api_url) as r:
            r.raise_for_status()
            statuses_json = r.json()

        for entry in statuses_json.get("statuses", []):
            name = entry.get("name", "")

            if service := next(filter(name.startswith, services), None):
                statuses[service] = entry.get("status", "Unknown")
    except (json.JSONDecodeError, requests.exceptions.RequestException):
        logger.error(msg)

    if (
        raise_on_outage and
        any(status not in {"OK", "Unknown"} for status in statuses.values())
    ):
        msg = f"At least 1 service is unhealthy: {services}"
        raise ServiceOutage(msg)

    return statuses
```

tests/integration/test_api.py
```diff
@@ -32,6 +35,111 @@
...
+
+
+@responses.activate
+def test_earthdata_status_service_outage():
+    responses.add(
+        responses.GET,
+        "https://status.earthdata.nasa.gov/api/v1/statuses",
+        json={
+            "statuses": [
+                {"name": "Earthdata Login", "status": "geospatial"},
```
```suggestion
                {"name": "Earthdata Login", "status": "Degraded"},
```

tests/integration/test_api.py
```diff
@@ -32,6 +35,111 @@
...
+        },
+        status=200,
+    )
+    with pytest.raises(ServiceOutage):
+        earthaccess.status(raise_on_outage=True)
+
+
+@responses.activate
+def test_earthdata_status_malformed_json():
+    services = ("Earthdata Login", "Common Metadata Repository")
```
chuckwondo: Let's move this out of the function and to the top of the file instead of duplicating it in multiple test functions.

tests/integration/test_api.py
```diff
@@ -32,6 +35,111 @@
...
+        responses.GET,
+        "https://status.earthdata.nasa.gov/api/v1/statuses",
+        body="Internal Server Error",
+        status=500,
+    )
+
+    result = earthaccess.status()
+    assert isinstance(result, dict)
+    assert result == expected
+
```
chuckwondo: Let's add a similar test function (perhaps named `test_earthdata_status_fetch_fail_raise`, but where `raise_on_error` is set to `True`, making sure `ServiceOutage` is _not_ raised in this case.

@mfisher87, alternatively, would we want `ServiceOutage` to be raised when the status service itself returns a server error response? I'm on the fence, since an error fetching the statuses does not necessarily mean that the services themselves are experiencing any problems.

docs/user_guide/status.md
```diff
@@ -0,0 +1,18 @@
...
+You can check the status of NASA Earthdata services manually for [PROD](https://status.earthdata.nasa.gov/) and [UAT](https://status.uat.earthdata.nasa.gov/) systems.
+
+## Programmatic Status Check
+
+You can check the status of Earthdata services  programmatically this way.
+
+```py
+import earthaccess
+
+nasa_status = earthaccess.status(system = PROD, raise_on_outage = True)
```
chuckwondo:
```suggestion
# Equivalent to earthaccess.status(system=PROD, raise_on_outage=False)
nasa_status = earthaccess.status()
```

chuckwondo (changes requested):

chuckwondo:
@Sherwin-14, please merge `main` into your branch to get your branch up to date.

tests/integration/test_api.py
mfisher87: Yes, I'd say just run this on the PROD url!

tests/integration/test_api.py
mfisher87: Yeah, I'd view that as a different outcome; but maybe we should keep it simple. If the status service is down, the status of Earthdata services is treated as "unknown". So, do we want to raise a `ServiceOutage` when status is not "OK", or do we want to raise it only when the status is explicitly down? I think it'd be OK to raise `ServiceOutage` on any status other than "OK", and listen for feedback from users from there.

Sherwin-14: So, should I proceed with adding the test `test_earthdata_status_fetch_fail_raise`?

earthaccess/api.py
```diff
@@ -28,6 +29,54 @@
...
+
+            if service := next(filter(name.startswith, services), None):
+                statuses[service] = entry.get("status", "Unknown")
+    except (json.JSONDecodeError, requests.exceptions.RequestException):
+        logger.error(msg)
+
+    if raise_on_outage and any(
+        status not in {"OK", "Unknown"} for status in statuses.values()
+    ):
+        msg = f"At least 1 service is unhealthy: {services}"
```
chuckwondo:
```suggestion
        msg = f"At least 1 service is in an unhealthy/unknown state: {services}"
```

earthaccess/api.py
```diff
@@ -28,6 +29,54 @@
...
+        {'Earthdata Login': 'OK', 'Common Metadata Repository': 'OK'}
+
+    Raises:
+        ServiceOutage: if at least one service status is not `"OK"`
+    """
+    services = ("Earthdata Login", "Common Metadata Repository")
+    statuses = {service: "Unknown" for service in services}
+    msg = (
+        f"Unable to retrieve Earthdata service statuses for {system}."
+        "  Try again later, or visit {system.status_url} to check service statuses."
```
chuckwondo:
```suggestion
        f"  Try again later, or visit {system.status_url} to check service statuses."
```

chuckwondo (changes requested):
@Sherwin-14, just a couple more fixes to error messages, but you can skip writing the extra tests we had discussed before.

earthaccess/api.py
mfisher87: Ooo good catch.

mfisher87:
@betolink are the tests failing because a new set of ECS collections have been migrated to the cloud? Do we need to re-generate our collection lists?

mfisher87:
See https://github.com/nsidc/earthaccess/pull/1061/files#r2257844461 for integration test failure details. 

Let's update the integration tests on main and then merge this PR after? **or** I'm fine with merging it now knowing the reason for the integration tests failure. 

@chuckwondo any last minute thoughts?

chuckwondo:
> See https://github.com/nsidc/earthaccess/pull/1061/files#r2257844461 for integration test failure details.
> 
> Let's update the integration tests on main and then merge this PR after? **or** I'm fine with merging it now knowing the reason for the integration tests failure.
> 
> @chuckwondo any last minute thoughts?

Either way is fine with me.

mfisher87:
Alright let's get this PR off the board. Thank you, @Sherwin-14 !!

mfisher87 (approved):

PR #1061, opened 2025-08-04
fsspec fetching improvements

betolink:
<details><summary>Pull Request (PR) draft checklist - click to expand</summary>

- [x] Please review our
      [contributing documentation](https://earthaccess.readthedocs.io/en/latest/contributing/)
      before getting started.
- [x] Populate a descriptive title. For example, instead of "Updated README.md", use a
      title such as "Add testing details to the contributor section of the README".
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [x] Populate the body of the pull request with:
    - A clear description of the change you are proposing.
    - Links to any issues resolved by this PR with text in the PR description, for
      example `closes #1`. See
      [GitHub docs - Linking a pull request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
- [x] Update `CHANGELOG.md` with details about your change in a section titled
      `## Unreleased`. If such a section does not exist, please create one. Follow
      [Common Changelog](https://common-changelog.org/) for your additions.
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [x] Update the documentation and/or the `README.md` with details of changes to the
      earthaccess interface, if any. Consider new environment variables, function names,
      decorators, etc.

Click the "Ready for review" button at the bottom of the "Conversation" tab in GitHub
once these requirements are fulfilled. Don't worry if you see any test failures in
GitHub at this point!

</details>

<details><summary>Pull Request (PR) merge checklist - click to expand</summary>

Please do your best to complete these requirements! If you need help with any of these
requirements, you can ping the `@nsidc/earthaccess-support` team in a comment and we
will help you out!

- [x] Add unit tests for any new features.
- [x] Apply formatting and linting autofixes. You can add a GitHub comment in this Pull
      Request containing "pre-commit.ci autofix" to automate this.
- [x] Ensure all automated PR checks (seen at the bottom of the "conversation" tab) pass.
- [x] Get at least one approving review.

</details>

[...]

PR #1065, opened 2025-08-08
Show progress bar in interactive sessions (Jupyter, REPL) by default

betolink:
This is a minor patch to restore the progress bar when we open and download files, the default is to show it if we are in a notebook and there is a new parameter to mute it with `hide_progress` in both download() and open(). 

----
📚 Documentation preview 📚: https://earthaccess--1065.org.readthedocs.build/en/1065/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Changed
 
-- By default, _disable_ progress bars during downloading.
+- Allow users to disable progress bar if needed, show for interactive sessions.
   ([#612](https://github.com/nsidc/earthaccess/issues/612))
+  ([#1065](https://github.com/nsidc/earthaccess/pull/1065))
   ([@Sherwin-14](https://github.com/Sherwin-14))
 - Updated bug and triage label names in bug Issue template.
   ([#998](https://github.com/nsidc/earthaccess/pull/998))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Changed
 
-- Allow users to disable progress bar if needed, show for interactive sessions.
+- Add `show_progress` argument to `earthaccess.download()` to let the user control display of progress bars. Defaults to true for interactive sessions, otherwise false.
   ([#612](https://github.com/nsidc/earthaccess/issues/612))
   ([#1065](https://github.com/nsidc/earthaccess/pull/1065))
   ([@Sherwin-14](https://github.com/Sherwin-14))
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -301,8 +301,7 @@ def download(
             of a UUID4 value.
         provider: if we download a list of URLs, we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
-        show_progress: if False, earthaccess will not show the progress bar for downloads. True if
-            the session is interactive.
+        show_progress: whether to display a progress bar. Defaults to `True` for interactive sessions, otherwise `False`.
         pqdm_kwargs: Additional keyword arguments to pass to pqdm, a parallel processing library.
             See pqdm documentation for available options. Default is to use immediate exception behavior
             and the number of jobs specified by the `threads` parameter.
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -252,7 +252,7 @@ def download(
             of a UUID4 value.
         provider: if we download a list of URLs, we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
-        hide_progress: if True, will not show the progress bar for downloads. True if 
+        hide_progress: if True, will not show the progress bar for downloads. True if
             the session is non-interactive.
         pqdm_kwargs: Additional keyword arguments to pass to pqdm, a parallel processing library.
             See pqdm documentation for available options. Default is to use immediate exception behavior
@@ -273,7 +273,12 @@ def download(
 
     try:
         return earthaccess.__store__.get(
-            granules, local_path, provider, threads, hide_progress=hide_progress, pqdm_kwargs=pqdm_kwargs
+            granules,
+            local_path,
+            provider,
+            threads,
+            hide_progress=hide_progress,
+            pqdm_kwargs=pqdm_kwargs,
         )
     except AttributeError as err:
         logger.error(
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -283,7 +283,7 @@ def download(
     provider: Optional[str] = None,
     threads: int = 8,
     *,
-    hide_progress: bool = False,
+    show_progress: Optional[bool] = None,
     pqdm_kwargs: Optional[Mapping[str, Any]] = None,
 ) -> List[Path]:
     """Retrieves data granules from a remote storage system. Provide the optional `local_path` argument to prevent repeated downloads.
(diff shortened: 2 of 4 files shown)


betolink:
We need to merge this before #1061 also the integration tests are failing due NSIDC ECS, they got hidden recently in favor of cloud copies but... that's coming in #1061 so I don't mind merging this with them failing.

earthaccess/api.py
```diff
@@ -283,6 +283,7 @@ def download(
     provider: Optional[str] = None,
     threads: int = 8,
     *,
+    hide_progress: bool = False,
```
mfisher87: I don't feel we should add a new kwarg here. Currently `pqdm_kwargs.disable` allows the user to control this behavior, and perhaps this PR should just update the default value instead of introducing a new API element?

mfisher87: I.e.

```python
pqdm_kwargs = {
    "n_jobs": threads,
    "disable": not _is_interactive(),
    **(pqdm_kwargs or {}),
}
```

betolink: makes sense, I feel like pqdm is kind of "obscure" and `hide_progress` is very explicit.

mfisher87: That's definitely true. You've kinda convinced me :) this is end-user-facing and we want it to be user-friendly. Maybe it should be `show_progress` instead? `hide_progress=False` is a double-negative.

betolink: done!

betolink: @mfisher87 it will be great if we can merge this to unblock #1061 and try to release v0.15.0 at the end of the week!

CHANGELOG.md
```diff
@@ -9,8 +9,9 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Changed
 
-- By default, _disable_ progress bars during downloading.
+- Allow users to disable progress bar if needed, show for interactive sessions.
```
mfisher87:
```suggestion
- Add `show_progress` argument to `earthaccess.download()` to let the user control display of progress bars. Defaults to true for interactive sessions, otherwise false.
```

earthaccess/api.py
```diff
@@ -300,6 +301,8 @@ def download(
             of a UUID4 value.
         provider: if we download a list of URLs, we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
+        show_progress: if False, earthaccess will not show the progress bar for downloads. True if
+            the session is interactive.
```
mfisher87:
```suggestion
        show_progress: whether to display a progress bar. Defaults to `True` for interactive sessions, otherwise `False`.
```

I felt "earthaccess" and "for downloads" were not necessary to include :)

earthaccess/store.py
```diff
@@ -61,6 +61,28 @@ def __repr__(self) -> str:
...
+    try:
+        from IPython import get_ipython  # type: ignore
+
+        # IPython Notebook or REPL:
+        if get_ipython() is not None:
+            return True
+    except ImportError:
+        pass
+
+    import sys
```
mfisher87: Move import to the top of the module?

earthaccess/store.py
```diff
@@ -71,13 +93,6 @@ def multi_thread_open(data: tuple[str, Optional[DataGranule]]) -> EarthAccessFil
...
         return EarthAccessFile(fs.open(url), granule)  # type: ignore
 
-    pqdm_kwargs = {
-        "exception_behaviour": "immediate",
-        "n_jobs": 8,
-        "disable": True,
-        **(pqdm_kwargs or {}),
-    }
-
     return pqdm(url_mapping.items(), multi_thread_open, **pqdm_kwargs)
```
mfisher87, line 93: `pqdm_kwargs` could be `None` at this point, so it's possible the user will be `NoneType is not a mapping` on this line.

earthaccess/store.py
```diff
@@ -355,12 +371,22 @@ def open(
...
             provider: e.g. POCLOUD, NSIDC_CPRD, etc.
+            show_progress: If False, progress bars will be hidden. Default is False.
             pqdm_kwargs: Additional keyword arguments to pass to pqdm, a parallel processing library.
                 See pqdm documentation for available options. Default is to use immediate exception behavior.
 
         Returns:
             A list of "file pointers" to remote (i.e. s3 or https) files.
         """
+        if not _is_interactive() and show_progress is None:
+            show_progress = False
```
mfisher87: If `show_progress is None`, we want it to be `True` if `_is_interactive()`. Maybe I'm missing something, but I think that's not handled. 

```suggestion
    if show_progress is None:
        show_progress = _is_interactive()
```

?

earthaccess/store.py
```diff
@@ -529,9 +557,12 @@ def get(
             uuid = uuid4().hex[:6]
             local_path = Path.cwd() / "data" / f"{today}-{uuid}"
 
+        if not _is_interactive() and show_progress is None:
+            show_progress = False
```
mfisher87: Same as above

earthaccess/store.py
```diff
@@ -529,9 +557,12 @@ def get(
...
+
         pqdm_kwargs = {
             "n_jobs": threads,
-            "disable": True,
+            "disable": not show_progress,  # True if show_progrss is False
```
mfisher87:
```suggestion
            "disable": not show_progress,
```

earthaccess/store.py
betolink: yeah the param in pqdm is disable so show_progress has to be negated then.

earthaccess/store.py
betolink: I think that's why the `or {}` is there no?

betolink:
Hi @chuckwondo if you have time, it would be great to have your input on this one. Just a minor edit on how we expose the progress bar. I'm waiting on this to keep working on #1061

earthaccess/store.py
mfisher87: Agreed, we should still do `disable: not show_progress` as we do now, but I think this code still needs to be changed so `show_progress` can be true by default. Right now, I think that isn't possible because of the `and`.

betolink: Yeah, I pushed that commit. I also included Chuck for a review in case you run out of time today.

chuckwondo:
Thanks @betolink! I'll take a look at this today.

earthaccess/store.py
```diff
@@ -27,6 +27,28 @@
...
+    except ImportError:
+        pass
+
+    import sys
+
+    # Python REPL
+    if hasattr(sys, "ps1"):
+        return True
+
+    return False
```
```suggestion
    # Python REPL
    return hasattr(sys, "ps1")
```

earthaccess/store.py
chuckwondo: @betolink, correct, that's what the `or {}` is for.

earthaccess/api.py
```diff
@@ -300,6 +301,7 @@ def download(
             of a UUID4 value.
         provider: if we download a list of URLs, we need to specify the provider.
         threads: parallel number of threads to use to download the files, adjust as necessary, default = 8
+        show_progress: whether to display a progress bar. Defaults to `True` for interactive sessions, otherwise `False`.
```
chuckwondo:
```suggestion
        show_progress: whether or not to display a progress bar. If not specified, defaults to `True` for interactive sessions (i.e., in a notebook or a python REPL session), otherwise `False`.
```

Please also duplicate this description to the rest of the places where `show_progress` was added so that the description is consistent everywhere. (There's a fair bit of variation/inconsistency across the other spots.)

chuckwondo:
@betolink, be sure to merge main into your branch, and address the pre-commit error.

betolink:
so weird, locally pre-commit passes 
image

earthaccess/api.py
betolink: done!  I'm taking a look at the pre-commit error.

betolink:
Ah my origin/main is not  in sync with earthaccess/main, fixing that now!

chuckwondo (approved):
Thanks @betolink! Looks good to me!

earthaccess/store.py
mfisher87: @betolink I show that line as deleted, so can't it still be `None`?

betolink: (paraphrasing an LLM) you're totally right! I think we should include the `or {}` in this line as well, technically this is not a public method and the methods that call it have the `or {}`, should we open an issue for it?

PR #1109, opened 2025-10-07
Multi-feature search support added to CMR api calls

tiagodc:
This pull request adds multi-feature querying support for all spatial types. The python_cmr library already supports querying with multiple features simultaneously, which are parsed from list inputs in the [`_build_url`](https://github.com/nasa/python_cmr/blob/develop/cmr/queries.py) function. This CMR functionality enables easy multipolygon [...] support with minimal code changes.

I created five new methods in the `DataGranules` class to enable explicit search of granules from multi-feature spatial inputs:
- `multi_bounding_box`
- `multipolygon`
- `multipoint`
- `multicircle`
- `multiline`

Those functions run the single-feature methods internally and update the `params` attribute inherited from the `Query` class of python_cmr, thus bulding single GET request URLs following the CMR standard.

Tests are provided for all the new methods and the documentation was updated to illustrate how to use them. 

I recently found out about `earthaccess`. It's an amazing tool and I hope this PR will be useful :)

### DIFF
--- a/earthaccess/search.py
+++ b/earthaccess/search.py
@@ -991,6 +991,25 @@ class DataGranules(GranuleQuery):
         self.options['line'] = {'or': True}
         return self
 
+    def multicircle(self, multi_circles: Sequence[Tuple[FloatLike,FloatLike,FloatLike]]) -> Self:
+        """Filter by granules that overlap any circle from an input list.
+
+        Parameters:
+            multi_circles: list of tuples of (lon, lat, radius)
+
+        Returns:
+            self
+        """
+        circles = []
+
+        for circle in multi_circles:
+            self.circle(*circle)
+            circles.append(self.params.pop('circle'))
+
+        self.params['circle'] = circles
+        self.options['circle'] = {'or': True}
+        return self
+
     @override
     def downloadable(self, downloadable: bool = True) -> Self:
         """Only match granules that are available for download. The inverse of this
--- a/earthaccess/search.py
+++ b/earthaccess/search.py
@@ -851,6 +851,26 @@ class DataGranules(GranuleQuery):
         """
         return super().point(lon, lat)
 
+    def multipoint(self, lon_lat_pairs: Sequence[PointLike]) -> Self:
+        """Filter by granules that include multiple geographic points.
+
+        Parameters:
+            lon_lat_pairs: sequence of (lon, lat) tuples
+
+        Returns:
+            self
+        """
+        points = []
+
+        for x, y in lon_lat_pairs:
+            self.point(x, y)
+            points.append(self.params.pop('point')[0])
+        
+        self.params['point'] = points
+        self.options['point'] = {'or': True}        
+        return self
+
+
     @override
     def polygon(self, coordinates: Sequence[PointLike]) -> Self:
         """Filter by granules that overlap a polygonal area. Must be used in combination
@@ -868,6 +888,25 @@ class DataGranules(GranuleQuery):
                 first and last coordinate pairs are not equal.
         """
         return super().polygon(coordinates)
+
+    def multipolygon(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
+        """Filter by granules that overlap any polygonal area from an input list.
+
+        Parameters:
+            multi_coordinates: list of lists of (lon, lat) tuples
+
+        Returns:
+            self
+        """
+        polygons = []
+                
+        for polygon in multi_coordinates:
+            self.polygon(polygon)
+            polygons.append(self.params.pop('polygon'))
+        
+        self.params['polygon'] = polygons
+        self.options['polygon'] = {'or': True}        
+        return self
 
     @override
     def bounding_box(
@@ -895,6 +934,25 @@ class DataGranules(GranuleQuery):
         return super().bounding_box(
             lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat
         )
+        
+    def multi_bounding_box(self, boxes: Sequence[Tuple[FloatLike, FloatLike, FloatLike, FloatLike]]) -> Self:
+        """Filter by granules that overlap any bounding box from an input list.
+
+        Parameters:
+            boxes: list of tuples of (lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat)
+
+        Returns:
+            self
+        """
+        bboxes = []
+                
+        for box in boxes:
+            self.bounding_box(*box)
+            bboxes.append(self.params.pop('bounding_box'))
+        
+        self.params['bounding_box'] = bboxes
+        self.options['bounding_box'] = {'or': True}
+        return self
 
     @override
     def line(self, coordinates: Sequence[PointLike]) -> Self:
@@ -913,6 +971,25 @@ class DataGranules(GranuleQuery):
                 pairs, or a coordinate could not be converted to a float.
         """
         return super().line(coordinates)
+    
+    def multiline(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
+        """Filter by granules that overlap any series of connected points from an input list.
+
+        Parameters:
+            multi_coordinates: a list of lists of (lon, lat) tuples
+
+        Returns:
+            self
+        """
+        lines = []
+                
+        for line in multi_coordinates:
+            self.line(line)
+            lines.append(self.params.pop('line'))
+        
+        self.params['line'] = lines
+        self.options['line'] = {'or': True}
+        return self
 
     @override
     def downloadable(self, downloadable: bool = True) -> Self:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 - Obstore and VirtualiZarr should not be required([#1097](https://github.com/nsidc/earthaccess/issues/1097))([@betolink](https://github.com/betolink))
 
+## Added:
+
+- Multi-feature search support (multi_bounding_box, multipolygon, multipoint, multicircle, multiline) from a single API call following the [CMR](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#g-polygon) standards.
+
 ## [v0.15.0] - 2025-09-16
 
 ### Changed
--- a/docs/user_guide/search.md
+++ b/docs/user_guide/search.md
@@ -394,6 +394,63 @@ results = earthaccess.search_data(
     )
 ```
 
+### Multi feature support
+
+`earthaccess` supports multi-feature searches for various spatial types, including polygons, points, lines, and more. To query multiple features of the same type, use the prefix multi followed by the feature type, such as:
+
+- `multi_bounding_box`
+- `multipolygon`
+- `multipoint`
+- `multicircle`
+- `multiline`
+
+When using `earthaccess.search_data`, the query will return granules that intersect any of the specified features. For example, to search using multiple polygons, you can structure your code as follows:
+
+```python
+polygons = [
+    # same polygon used in the single query
+    [
+        (-49.64860422604741, 69.23553485026147),
+        (-49.667876114626296, 69.07309059285959),
+        (-49.1722491331669, 69.03175841820749),
+        (-47.53552489113113, 69.03872918462292),
+        (-47.35616491854395, 69.22149993224824),
+        (-48.1447695277283, 69.33507802083219),
+        (-49.178671242118384, 69.29455117736225),
+        (-49.64860422604741, 69.23553485026147),
+    ],    
+    # a second polygon over the Eyjafjallajökull volcano in Iceland
+    [
+        (-19.61490317965708, 63.63370144220765), 
+        (-19.61490317965708, 63.61370144220765),
+        (-19.59490317965708, 63.61370144220765),
+        (-19.59490317965708, 63.63370144220765),
+        (-19.61490317965708, 63.63370144220765),
+    ]
+]
+
+results = earthaccess.search_data(
+    short_name="ATL06",
+    multipolygon=polygons,
+    )
+```
+
+Similarly, to query multiple points, you can use:
+
+```python
+lon_lat_pairs = [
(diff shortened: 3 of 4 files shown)


asteiker:
Thank you very much @tiagodc ! We appreciate your first contribution to `earthaccess` and for your positive words about the library. I linked to the related #490 enhancement Issue in our backlog.

earthaccess/search.py
```diff
@@ -851,6 +851,26 @@ def point(self, lon: FloatLike, lat: FloatLike) -> Self:
...
+        Returns:
+            self
+        """
+        points = []
+
+        for x, y in lon_lat_pairs:
+            self.point(x, y)
+            points.append(self.params.pop('point')[0])
+        
+        self.params['point'] = points
```
chuckwondo, line 869: This should be unnecessary because the `point` method already performs an append each time it is called. That is, the `point` method is already implemented to support being called multiple times, adding a new point to the list of points on each call.

Therefore, this should be all that's required:

```suggestion
        for x, y in lon_lat_pairs:
            self.point(x, y)
        
```

earthaccess/search.py
```diff
@@ -851,6 +851,26 @@ def point(self, lon: FloatLike, lat: FloatLike) -> Self:
         """
         return super().point(lon, lat)
 
+    def multipoint(self, lon_lat_pairs: Sequence[PointLike]) -> Self:
```
chuckwondo, line 854: Please add an or_ kwarg so the caller can explicitly set the option, instead of assuming it should be set to True (and please update the docstring to include it):

```suggestion
    def points(self, lon_lat_pairs: Sequence[PointLike], *, or_: bool | None = None) -> Self:
```

I would prefer `or_` to be a kwarg because passing a bool value without a keyword is not intuitive.

However, making it a positional arg allows for it to be set like so (whereas this is not possible when `or_` is a kwarg):

```python
earthaccess.search_data(
    short_name="ATL06",
    polygons=(polygons, True),
)
```

Again, it's not intuitive as to what `True` means here, so I'd like some input from others on this point (cc: @betolink, @mfisher87, @jhkennedy).

I don't like the "bare" boolean argument, but on the other hand, I don't like that (sub)kwargs cannot be specified via `search_data`.

In order to pass `or_` when it is a kwarg, the caller cannot use `search_data`, and must instead directly construct a `DataGranules` instance, like so:

```python
query = (
    DataGranules()
    .short_name("ATL06")
    .polygons(polygons, or_=True)
)
```

(Alternatively, renaming `or_` to `any_` might make it even more readable -- i.e., any/all is more intuitive than or/and, no?)

earthaccess/search.py
```diff
@@ -851,6 +851,26 @@ def point(self, lon: FloatLike, lat: FloatLike) -> Self:
...
+            self
+        """
+        points = []
+
+        for x, y in lon_lat_pairs:
+            self.point(x, y)
+            points.append(self.params.pop('point')[0])
+        
+        self.params['point'] = points
+        self.options['point'] = {'or': True}
```
chuckwondo, line 870: I would prefer not to assume the caller wants the `or` option to be set to `True`, so let's add an `or_` kwarg, defaulted to `None`:

```suggestion
        self.option("point", "or", or_)
```

earthaccess/search.py
```diff
@@ -869,6 +889,25 @@ def polygon(self, coordinates: Sequence[PointLike]) -> Self:
         """
         return super().polygon(coordinates)
 
+    def multipolygon(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
```
chuckwondo, line 892: Please add an `or_` kwarg, as I've described in my comment on the `multipoint` method:

```suggestion
    def polygons(self, polygons_: Sequence[Sequence[PointLike]], *, or_: bool | None = None) -> Self:
```

earthaccess/search.py
```diff
@@ -869,6 +889,25 @@ def polygon(self, coordinates: Sequence[PointLike]) -> Self:
...
+            self
+        """
+        polygons = []
+                
+        for polygon in multi_coordinates:
+            self.polygon(polygon)
+            polygons.append(self.params.pop('polygon'))
+        
+        self.params['polygon'] = polygons
+        self.options['polygon'] = {'or': True}
```
chuckwondo, line 908:
```suggestion
        self.option("polygon", "or", or_)
```

earthaccess/search.py
```diff
@@ -895,6 +934,25 @@ def bounding_box(
         return super().bounding_box(
             lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat
         )
+        
+    def multi_bounding_box(self, boxes: Sequence[Tuple[FloatLike, FloatLike, FloatLike, FloatLike]]) -> Self:
```
chuckwondo, line 938:
```suggestion
    def bounding_boxes(self, bboxes: Sequence[tuple[FloatLike, FloatLike, FloatLike, FloatLike]], *, or_: bool | None = None) -> Self:
```

earthaccess/search.py
```diff
@@ -895,6 +934,25 @@ def bounding_box(
...
+            self
+        """
+        bboxes = []
+                
+        for box in boxes:
+            self.bounding_box(*box)
+            bboxes.append(self.params.pop('bounding_box'))
+        
+        self.params['bounding_box'] = bboxes
+        self.options['bounding_box'] = {'or': True}
```
chuckwondo, line 954:
```suggestion
        self.option("bounding_box", "or", or_)
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
                 pairs, or a coordinate could not be converted to a float.
         """
         return super().line(coordinates)
+    
+    def multiline(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
```
chuckwondo, line 975:
```suggestion
    def lines(self, lines_: Sequence[Sequence[PointLike]], *, or_: bool | None = None) -> Self:
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+            self
+        """
+        lines = []
+                
+        for line in multi_coordinates:
+            self.line(line)
+            lines.append(self.params.pop('line'))
+        
+        self.params['line'] = lines
+        self.options['line'] = {'or': True}
```
chuckwondo, line 991:
```suggestion
        self.option("line", "or", or_)
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+        return self
+
+    def multicircle(self, multi_circles: Sequence[Tuple[FloatLike,FloatLike,FloatLike]]) -> Self:
```
chuckwondo, line 994:
```suggestion
    def circles(self, circles_: Sequence[Tuple[FloatLike,FloatLike,FloatLike]], *, or_: bool | None = None) -> Self:
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+            self
+        """
+        circles = []
+
+        for circle in multi_circles:
+            self.circle(*circle)
+            circles.append(self.params.pop('circle'))
+
+        self.params['circle'] = circles
+        self.options['circle'] = {'or': True}
```
chuckwondo, line 1010:
```suggestion
        self.option("circle", "or", or_)
```

earthaccess/search.py
```diff
@@ -869,6 +889,25 @@ def polygon(self, coordinates: Sequence[PointLike]) -> Self:
...
+        Returns:
+            self
+        """
+        polygons = []
+                
+        for polygon in multi_coordinates:
+            self.polygon(polygon)
+            polygons.append(self.params.pop('polygon'))
+        
+        self.params['polygon'] = polygons
```
chuckwondo, line 907:
```suggestion
        self.params["polygon"] = [
            self.polygon(polygon).params.pop("polygon") for polygon in polygons_
        ]
```

earthaccess/search.py
```diff
@@ -869,6 +889,25 @@ def polygon(self, coordinates: Sequence[PointLike]) -> Self:
         """
         return super().polygon(coordinates)
 
+    def multipolygon(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
+        """Filter by granules that overlap any polygonal area from an input list.
+
+        Parameters:
+            multi_coordinates: list of lists of (lon, lat) tuples
```
chuckwondo, line 896:
```suggestion
            polygons_: list of lists of (lon, lat) tuples
```

earthaccess/search.py
```diff
@@ -895,6 +934,25 @@ def bounding_box(
...
+        Returns:
+            self
+        """
+        bboxes = []
+                
+        for box in boxes:
+            self.bounding_box(*box)
+            bboxes.append(self.params.pop('bounding_box'))
+        
+        self.params['bounding_box'] = bboxes
```
chuckwondo, line 953:
```suggestion
        self.params["bounding_box"] = [
            self.bounding_box(*bbox).params.pop("bounding_box") for bbox in bboxes
        ]
```

earthaccess/search.py
```diff
@@ -895,6 +934,25 @@ def bounding_box(
         return super().bounding_box(
             lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat
         )
+        
+    def multi_bounding_box(self, boxes: Sequence[Tuple[FloatLike, FloatLike, FloatLike, FloatLike]]) -> Self:
+        """Filter by granules that overlap any bounding box from an input list.
+
+        Parameters:
+            boxes: list of tuples of (lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat)
```
chuckwondo, line 942:
```suggestion
            bboxes: list of tuples of (lower_left_lon, lower_left_lat, upper_right_lon, upper_right_lat)
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+        Returns:
+            self
+        """
+        lines = []
+                
+        for line in multi_coordinates:
+            self.line(line)
+            lines.append(self.params.pop('line'))
+        
+        self.params['line'] = lines
```
chuckwondo, line 990:
```suggestion
        self.params["line"] = [
            self.polygon(line).params.pop("line") for line in lines_
        ]
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
                 pairs, or a coordinate could not be converted to a float.
         """
         return super().line(coordinates)
+    
+    def multiline(self, multi_coordinates: Sequence[Sequence[PointLike]]) -> Self:
+        """Filter by granules that overlap any series of connected points from an input list.
+
+        Parameters:
+            multi_coordinates: a list of lists of (lon, lat) tuples
```
chuckwondo, line 979:
```suggestion
            lines_: a list of lists of (lon, lat) tuples
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+        Returns:
+            self
+        """
+        circles = []
+
+        for circle in multi_circles:
+            self.circle(*circle)
+            circles.append(self.params.pop('circle'))
+
+        self.params['circle'] = circles
```
chuckwondo, line 1009:
```suggestion
        self.params["circle"] = [
            self.circle(*circle).params.pop("circle") for circle in circles_
        ]
```

earthaccess/search.py
```diff
@@ -913,6 +971,44 @@ def line(self, coordinates: Sequence[PointLike]) -> Self:
...
+        
+        self.params['line'] = lines
+        self.options['line'] = {'or': True}
+        return self
+
+    def multicircle(self, multi_circles: Sequence[Tuple[FloatLike,FloatLike,FloatLike]]) -> Self:
+        """Filter by granules that overlap any circle from an input list.
+
+        Parameters:
+            multi_circles: list of tuples of (lon, lat, radius)
```
chuckwondo, line 998:
```suggestion
            circles_: list of tuples of (lon, lat, radius)
```

chuckwondo (changes requested):
@tiagodc, thank you for your contribution, and apologies that it has taken so long to get it reviewed.

I've made comments, but I've also solicited some other maintainers for input as well, so let's see what they have to say about my suggestions and questions.

tiagodc:
Thank you for reviewing it @chuckwondo. I'll wait to see what the other reviewers say before committing further changes - mainly the `_or` kwarg. 

I set up the `or` to be true by default as I didn't see a way to enable a kwarg in `search_data` without updating the method itself - which i tried to avoid in a 1st PR, so I assumed `or` to be the most common use case. Setting the argument explicitly by the user would be ideal - and perhaps setting `_any=True` by default? 

Regardless if it may require further changes to `search_data`, I'll be happy to go with whatever solution works best in your view.

chuckwondo:
@mfisher87, @betolink, @jhkennedy, any of you able to look over this? I took a first pass, but I added a question to get input from others about how we want to approach something: https://github.com/nsidc/earthaccess/pull/1109/files#r2491099493.

@tiagodc, in the meantime, would you mind merging the latest changes from the `main` branch into your branch?

tiagodc:
Hi all, any updates on the review status? @mfisher87, @betolink, @jhkennedy, @chuckwondo

chuckwondo:
> Hi all, any updates on the review status? @mfisher87, @betolink, @jhkennedy, @chuckwondo

Apologies for the delay.

@mfisher87, @betolink, @jhkennedy, I took an initial pass over this a while back, but included some questions for you in several comments regarding how we might want to deal with some query options in an intuitive manner. I'm not completely happy with what I initially suggested, which is why I'd like more eyes on this with perhaps some alternative suggestions.

It's been a while since I looked at this myself, so I'll take another pass, hopefully with fresher eyes and perhaps a slightly different perspective.

Issue #203, opened 2023-02-20
Add methods to return DOI and/or data citation


andypbarrett:
To make good data citation practices easier to follow, it would be great to have a method or methods that return the DOI and data citation for a dataset.  This would allow citations to be printed in notebooks or exported to bibliographies.
```
>results.get_doi()
10.7265/jj4s-tq79
>results.get_citation(format="APA")
Walsh, J. E., W. L. Chapman, F. Fetterer, and J. S. Stewart. (2019). Gridded Monthly Sea Ice Extent and Concentration, 1850 Onward, Version 2 [Data Set]. Boulder, Colorado USA. National Snow and Ice Data Center. https://doi.org/10.7265/jj4s-tq79. Date Accessed 02-20-2023.
>results.get_citation(format="bibtex")
@misc{Walsh, J. E., W. L. Chapman, F. Fetterer, and J. S. Stewart._2019, title={Gridded Monthly Sea Ice Extent and Concentration, 1850 Onward, Version 2}, url={https://nsidc.org/data/G10010/versions/2}, DOI={10.7265/jj4s-tq79}, publisher={National Snow and Ice Data Center}, author={Walsh, J. E., W. L. Chapman, F. Fetterer, and J. S. Stewart.}, year={2019} }
```

mfisher87:
I love the idea of making it easy for people to cite. I imagine we'd want to generate citations from individual granules or individual collections? E.g. `results.get_doi()`, even if we have a `Results` class (#860), wouldn't always map to a single DOI.

Sherwin-14:
@mfisher87 I would love to collaborate on this one. I tried retrieving **DOI** and **Citations** for _collections_ (although we are yet to decide whether this should be for granules and collections or any one of them), but I do have a questions.

I made a request for the concept_id C2789815241-ENVIDAT. In response there is UMM dictionary that contains two elements within it, one named `DOI` which contains **DOI** for the _collection_ and the other one named `CollectionCitations` which contains **Citations**. Below I have attached the request as well as the UMM part of the response.

Request:

```curl
curl "https://cmr.earthdata.nasa.gov/search/collections.umm_json?concept_id=C2789815241-ENVIDAT" | jq .
``` 

Response:

```JSON
{
  "umm": {
   ....
    "CollectionCitations": [
      {
        "Creator": "Katrin Meusburger, Elisabeth Graf Pannatier, Marcus Schaub",
        "Editor": "Katrin Meusburger",
[...]

PR #1112, opened 2025-10-14
Added methods for DOI and Citation

Sherwin-14:
I've added methods to get DOI and citations for collections as discussed [here](https://github.com/nsidc/earthaccess/issues/203). Please let me know if there are improvements upon this.

----
📚 Documentation preview 📚: https://earthaccess--1112.org.readthedocs.build/en/1112/

### DIFF
--- a/earthaccess/results.py
+++ b/earthaccess/results.py
@@ -13,21 +13,6 @@ from .services import DataServices
 
 @cache
 def _citation(*, doi: str, format: str, language: str) -> str:
-    """Generate a formatted citation for a collection using its DOI.
-
-    Parameters:
-        doi: The DOI of the collection.
-        format: Citation format style (e.g., 'apa', 'bibtex', 'ris').
-             For a full list of valid formats, visit <https://citation.doi.org/>
-        language: Language code (e.g., 'en-US').
-             For a full list of valid language codes, visit <https://citation.doi.org/>
-
-    Returns:
-         The formatted citation as a string.
-
-    Raises:
-         requests.RequestException: if fetching citation information from citations.doi.org failed.
-    """
     response = requests.get(
         "https://citation.doi.org/format",
         params={"doi": doi, "style": format, "lang": language},
@@ -147,7 +132,7 @@ class DataCollection(CustomDict):
         return None
 
     def citation(self, *, format: str, language: str) -> str | None:
-        """Generate a formatted citation for this collection using its DOI.
+        """Fetch a formatted citation for this collection using its DOI.
 
         Parameters:
             format: Citation format style (e.g., 'apa', 'bibtex', 'ris').


earthaccess/results.py
```diff
@@ -107,6 +109,32 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        Returns:
+             The formatted citation as a string.
+        """
+        if not self.get_doi():
+            raise ValueError(
+                "The collection is missing a DOI, citation generation is not supported."
+            )
+        return requests.get(
+            f"https://citation.doi.org/format?doi={self.get_doi()}&style={format}&lang={language}"
+        ).text
```
danielfromearth: Perhaps we should check for the status of the HTTP response explicitly, for example using `response.raise_for_status()` before attempting to access the `.text` attribute?


```suggestion
        response = requests.get(
            f"https://citation.doi.org/format?doi={self.get_doi()}&style={format}&lang={language}"
        )
        response.raise_for_status()  # Raise exception for HTTP errors
        return response.text
```

earthaccess/results.py
```diff
@@ -107,6 +109,32 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+    def get_doi(self) -> str:
+        """Placeholder.
+
+        Returns:
+            A collection's DOI information.
+        """
+        return self["umm"].get("DOI", "").get("DOI", "")
+
+    def get_citation(self, format: str, language: str) -> str:
+        """Placeholder.
```
danielfromearth:
```suggestion
        """Generate a formatted citation for this collection using its DOI.
```

earthaccess/results.py
```diff
@@ -107,6 +109,32 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
         """
         return self["umm"].get(umm_field, "")
 
+    def get_doi(self) -> str:
+        """Placeholder.
```
danielfromearth:
```suggestion
        """Retrieve the Digital Object Identifier (DOI) for this collection.
```

earthaccess/results.py
Sherwin-14: @danielfromearth I was thinking of making this change for all of the methods in a separate PR. Would that be fine with you?

danielfromearth: okay, that works for me 👍

danielfromearth (dismissed):

earthaccess/results.py
chuckwondo: I have to disagree here. When you add new code, add complete docstrings.

If you want to updated _existing_ docstrings, open a separate PR, but don't add new code with incomplete docstrings.

chuckwondo: Also, please include in the docstring that the full lists of valid values for format and language can be found by visiting https://citation.doi.org/

Sherwin-14:
> Also, please add some unit tests.

The file `test_results.py` contains VCR cassettes test so should I add the unit tests (for citation and DOI) in the same file?

chuckwondo:
> > Also, please add some unit tests.
> 
> The file `test_results.py` contains VCR cassettes test so should I add the unit tests (for citation and DOI) in the same file?

I'm not looking at the repo at the moment, but if there is code in this file with corresponding tests in `test_results.py`, then that's what I would do. Generally speaking, when adding/changing code that has tests in a test file, add/change test functions in that same test file.

chuckwondo:
Hi @Sherwin-14. Any chance you can wrap this up soon?

Sherwin-14:
> Hi @Sherwin-14. Any chance you can wrap this up soon?

Apologies! I am going through a very busy period, but I'll try to complete this by the end of the week. I have just one question (sorry for repeating this again, as I wasn't clear with my question above): the `test_results.py` file contains no tests for the `DataCollections` class. Should I add them as individual test functions or should I create a class and add them there?

chuckwondo:
> > Hi @Sherwin-14. Any chance you can wrap this up soon?
> 
> Apologies! I am going through a very busy period, but I'll try to complete this by the end of the week. I have just one question (sorry for repeating this again, as I wasn't clear with my question above): the `test_results.py` file contains no tests for the `DataCollections` class. Should I add them as individual test functions or should I create a class and add them there?

No worries. Just add individual test functions. No need for the complexities of a test class (which I recommend against in general -- IMHO, it's a questionable recreation of JUnit [from the Java world] in Python).

earthaccess/results.py
```diff
@@ -107,6 +109,39 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+    def get_doi(self) -> str:
+        """Retrieve the Digital Object Identifier (DOI) for this collection.
+
+        Returns:
+            A collection's DOI information.
+        """
+        doi = self["umm"].get("DOI", {})
+        if isinstance(doi, dict):
+            return doi.get("DOI", "")
+        return ""
```
chuckwondo: I'm thinking we should return `None` rather than an empty string when there is no DOI value. Alternatively, we might want to raise an error when there is no DOI value, to be consistent with `get_citation`.

Either way, I think we should make `get_doi` and `get_citation` consistent with each other: they should either both return empty strings when there is no DOI, or they should both return `None`, or they should both raise an exception.

@betolink, @mfisher87, @jhkennedy, what do you think?

chuckwondo (changes requested):
@Sherwin-14, nice work on the unit tests!

I've asked other for some input on making `get_doi` and `get_citation` consistent with each other on returning `None` or raising exceptions, so let's see what kind of feedback we get.

earthaccess/results.py
betolink: I like `None` or a clear message `DOI not available`

betolink:
General question, I remember this already existed for `search_data()` not just for collections, was it removed? I don't think it works anymore, in practice it saved the user a query, it did a collection by `doi` first.

chuckwondo:
> General question, I remember this already existed for `search_data()` not just for collections, was it removed? I don't think it works anymore, in practice it saved the user a query, it did a collection by `doi` first.

It still exists, but this method is different. It's for _extracting_ the DOI from an item in a result list, not for _querying_ on DOI.

earthaccess/results.py
chuckwondo: > I like `None` or a clear message `DOI not available`

I'm okay with either returning `None` or raising an exception, but opposed to returning a string that represents an error when a string also represents a valid return value. That makes is too easy to _not_ see that something went wrong.

jhkennedy: I agree `None` or an exception. Python best practices would typically say raise an exception over optional returns (None | str), but thinking about client code, I'd rather handle and unexpected None than have it crash out with an exception, so I vote `None`.

chuckwondo: @Sherwin-14, a few PRs have landed on `main`, so please pull the latest from `main` and merge into your branch, and please update the 2 new functions to return `None`, rather than returning an empty (or error) string or raising an exception.

earthaccess/results.py
```diff
@@ -107,6 +109,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
         """
         return self["umm"].get(umm_field, "")
 
+    def get_doi(self) -> Optional[str]:
```
chuckwondo: Let's avoid the `get_` prefix. The other methods don't include such a prefix, except for `get_data`, but that's because it returns the "GET DATA" links, so "get" is in the name of the thing it's getting the value of.

Also, let's use a pipe (`|`) instead of `Optional`, since our min python version supports them. (As we update code, we should replace `Optional` with `|`.)

```suggestion
    def doi(self) -> str | None:
```

earthaccess/results.py
```diff
@@ -107,6 +109,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        """Retrieve the Digital Object Identifier (DOI) for this collection.
+
+        Returns:
+            A collection's DOI information or None.
```
chuckwondo:
```suggestion
            This collection's DOI information, or `None`, if it has none.
```

earthaccess/results.py
```diff
@@ -107,6 +109,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        """
+        doi = self["umm"].get("DOI", {})
+        if isinstance(doi, dict):
+            return doi.get("DOI", None)
+        return None
+
+    def get_citation(self, format: str, language: str) -> Optional[str]:
```
chuckwondo: Let's drop the `get_` prefix here too, use `|` instead of `Optional`, and use kw-only params since they are both type `str`, which makes it too easy to pass them in the wrong order without forcing keyword usage:

```suggestion
    def citation(self, *, format: str, language: str) -> str | None:
```

earthaccess/results.py
```diff
@@ -107,6 +109,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        """Generate a formatted citation for this collection using its DOI.
+
+        Parameters:
+            format: Citation format style (e.g., 'apa', 'bibtex', 'ris').
+                 For a full list of valid formats, visit https://citation.doi.org/
+            language: Language code (e.g., 'en-US').
+                 For a full list of valid language codes, visit https://citation.doi.org/
```
chuckwondo: In Markdown, bare URLs should be enclosed in angle brackets (`<` and `>`):

```suggestion
                 For a full list of valid formats, visit <https://citation.doi.org/>.
            language: Language code (e.g., 'en-US').
                 For a full list of valid language codes, visit <https://citation.doi.org/>.
```

earthaccess/results.py
```diff
@@ -107,6 +109,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+
+        Returns:
+             The formatted citation as a string or None.
+        """
```
chuckwondo, line 148:
```suggestion

        Raises:
            requests.RequestException: If fetching citation information from citation.doi.org failed.
        """
```

tests/unit/test_results.py
```diff
@@ -224,3 +226,77 @@ def assert_is_using_search_after(self, cass):
...
+
+    citation = collection.get_citation(format="apa", language="fr-FR")
+
+    assert (
+        citation
+        == "Meusburger, K., Graf Pannatier, E., & Schaub, M. (2019). 10-HS Pfynwald (Version 2019) [Jeu de données]. EnviDat. https://doi.org/10.16904/ENVIDAT.LWF.34"
+    )
+
+
+def test_get_citation_raises_error_when_doi_missing():
```
chuckwondo:
```suggestion
def test_get_citation_returns_none_when_doi_missing():
```

tests/unit/test_results.py
```diff
@@ -224,3 +226,77 @@ def assert_is_using_search_after(self, cass):
...
+    collection = DataCollection({"umm": {}, "meta": {}})
+
+    assert collection.get_citation(format="apa", language="en-US") is None
+
+
+def test_get_citation_raises_error_when_doi_empty():
```
chuckwondo:
```suggestion
def test_get_citation_returns_none_when_doi_empty():
```

CHANGELOG.md
```diff
@@ -9,7 +9,12 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [v0.15.1] - 2025-09-16
 
-## Fixed:
+### Added
+
+- Added methods to return DOI and data citation ([#203](https://github.com/nsidc/earthaccess/issues/203))
```
chuckwondo:
```suggestion
- Added methods `doi` and `citation` to `DataCollection` class ([#203](https://github.com/nsidc/earthaccess/issues/203)) (@Sherwin-14, @chuckwondo)
```

CHANGELOG.md
```diff
@@ -9,7 +9,12 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [v0.15.1] - 2025-09-16
 
-## Fixed:
+### Added
```
chuckwondo, line 10: Move this section under `[Unreleased]`.

chuckwondo (changes requested):

earthaccess/results.py
```diff
@@ -107,6 +109,41 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        """Generate a formatted citation for this collection using its DOI.
+
+        Parameters:
+            format: Citation format style (e.g., 'apa', 'bibtex', 'ris').
+                 For a full list of valid formats, visit <https://citation.doi.org/>
+            language: Language code (e.g., 'en-US').
+                 For a full list of valid language codes, visit <https://citation.doi.org/>
+
+        Returns:
+             The formatted citation as a string or `None`.
```
chuckwondo:
```suggestion
             The formatted citation as a string, or `None`, if this collection does not have a DOI.
```

earthaccess/results.py
```diff
@@ -107,6 +109,41 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+             The formatted citation as a string, or `None`, if this collection does not have a DOI.
+
+        Raises:
+             requests.RequestException: if fetching citation information from citations.doi.org failed.
+        """
+        if not self.doi():
+            return None
+
+        response = requests.get(
+            f"https://citation.doi.org/format?doi={self.doi()}&style={format}&lang={language}"
```
chuckwondo: Sorry, I missed this earlier, but prefer passing `params` rather than manually constructing the URL with query params:

```suggestion
            "https://citation.doi.org/format",
            params={"doi": self.doi(), "style": format, "lang": language},
```

earthaccess/results.py
```diff
@@ -107,6 +109,41 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+
+        Returns:
+            This collection's DOI information, or `None`, if it has none.
+        """
+        doi = self["umm"].get("DOI", {})
+        if isinstance(doi, dict):
+            return doi.get("DOI", None)
+        return None
+
+    def citation(self, *, format: str, language: str) -> str | None:
```
chuckwondo, line 134: Let's cache the citations to avoid making multiple requests for the same citation:

```suggestion
    @cache
    def citation(self, *, format: str, language: str) -> str | None:
```

earthaccess/results.py
```diff
@@ -2,6 +2,8 @@
 import uuid
 from typing import Any, Dict, List, Optional, Union
```
chuckwondo, line 4:
```suggestion
from functools import cache
from typing import Any, Dict, List, Optional, Union
```

earthaccess/results.py
Sherwin-14: DataCollections does not have the ` __hash__` method, so @cache won't work.

chuckwondo: Ah, right, because `DataCollections` is a `dict`.

In that case, let's create a top-level function in this file, and we'll decorate that with `@cache`, and have this method call the function, like so:

```python
@cache
def _citation(*, doi: str, format: str, language: str) -> str:
    response = requests.get(
        "https://citation.doi.org/format",
        params={"doi": doi, "style": format, "lang": language},
    )
    response.raise_for_status()
    return response.text
```

Please copy the docstring from the `citation` method to the new `_citation` function above, and adjust it accordingly since the `doi` is now a parameter, and the return type is `str` instead of `str | None` since `doi` is `str` instead of `str | None`.

Then, call it from the `citation` method, like so (keeping the current docstring as is):

```python
def citation(self, *, format: str, language: str) -> str | None:
    return (
        None
        if (doi := self.doi()) is None
        else _citation(doi=doi, format=format, language=language)
    )
```

chuckwondo (approved):
Thanks for your patience @Sherwin-14! Nicely done.

@betolink, @mfisher87, and @jhkennedy, would at least one of you please give this a second review?

earthaccess/results.py
```diff
@@ -107,6 +135,38 @@ def get_umm(self, umm_field: str) -> Union[str, Dict[str, Any]]:
...
+        Returns:
+            This collection's DOI information, or `None`, if it has none.
+        """
+        doi = self["umm"].get("DOI", {})
+        if isinstance(doi, dict):
+            return doi.get("DOI", None)
+        return None
+
+    def citation(self, *, format: str, language: str) -> str | None:
+        """Generate a formatted citation for this collection using its DOI.
```
itcarroll: Let's be clear we are not generating the citation ourselves.

```suggestion
        """Fetch a formatted citation for this collection using its DOI.
```

itcarroll (changes requested):
A nice feature, thank you!

The answer may be a quick no, but is it possible we could avoid the module level `_citation` function by defining `__hash__` as we did for `DataGranule`? Or perhaps `DataGranule.__hash__` should be moved to `CustomDict` so both `DataGranule` and `DataCollection` are hashable in the same way?

I believe the reason for `_citation` is the inability to use `@cache` on `DataCollection.citation`, but the current solution appears to require duplicating a docstring which is bad for maintainability. I noticed this when suggesting a word change in the docstring.

chuckwondo:
> The answer may be a quick no, but is it possible we could avoid the module level `_citation` function by defining `__hash__` as we did for `DataGranule`? Or perhaps `DataGranule.__hash__` should be moved to `CustomDict` so both `DataGranule` and `DataCollection` are hashable in the same way?
> 
> I believe the reason for `_citation` is the inability to use `@cache` on `DataCollection.citation`, but the current solution appears to require duplicating a docstring which is bad for maintainability. I noticed this when suggesting a word change in the docstring.

Fair point regarding the duplication of the docstring.

TL;DR: I suggest simply removing the docstring from the `_citation` function since it is a "private" function, not intended (at least not at the moment) to be called by the user.

Regarding implementing `__hash__`, we may want to do so, although I think that's a separate can of worms. In brief, although this is perhaps not likely to cause problems in our case, I recommend against making anything that is mutable to also be hashable. If you change the value upon which the hash is based, you have problems. More broadly, we should refactor `DataGranule` and `DataCollection` to be immutable in order for the `__hash__` methods to be reliable.

Further, even if mutability were not an issue, I would suggest not using the `DataCollection` (specifically it's hash value) as part of the cache key, because the citation specifically relies on the DOI, not the value of the collection's hash value (perhaps concept ID), so I recommend that the cache key uses the more specific values that the citation relies upon (DOI, format, and language).

chuckwondo:
@itcarroll, since you gave a thumbs-up to my reply to your comment, are you okay with approving this PR now, so we can unblock merging?

earthaccess/results.py
```diff
@@ -1,13 +1,26 @@
...
 
+import requests
+
 import earthaccess
 
 from .formatters import _repr_granule_html
 from .services import DataServices
 
 
+@cache
```
betolink, line 14: Nice!

PR #1129, opened 2025-11-06
Use only specified login strategy, if not "all"

### LINKED ISSUE #945: [BUG] earthaccess.login tries other strategies when it should not

### Current Behavior

When calling `earthaccess.login` with a strategy other than `"all"`, it _might_ call one or more other strategies.

### Expected Behavior

When calling `earthaccess.login` with a strategy other than `"all"`, it should attempt _only_ the specified strategy.

### Steps To Reproduce

The following should _ignore_ the env vars as well as netrc, and _only_ prompt for creds, but does not.

It uses the env vars to login, even though the strategy is `"interactive"`, not `"all"` (of course, replace `username` and `password` with valid Earthdata Login creds):

```
EARTHDATA_USERNAME=username EARTHDATA_PASSWORD=password python -c 'import earthaccess; earthaccess.login("interactive")'
```

If the creds are valid, you are not prompted for creds, as the env vars are used to successfully login, and thus the "interactive" strategy is skipped.

Conversely, if you use _invalid_ creds for the env vars, login will still attempt to use them, but since login will fail with them, the "netrc" strategy will be tried. If you have a valid netrc entry, login will succeed without prompting you for creds.

_Only_ when neither valid env vars nor a valid netrc entry exists will you will be prompted to enter creds. However, _neither_ of the other strategies should be attempted at all, not even if a user enters invalid creds interactively.

### Environment

```markdown
- OS: macOS 15.3
- Python: 3.10.15
```

### Additional Context

_No response_

chuckwondo:
Fixes #945

- [x] Please review our
      before getting started.
      example `closes #1`. See
      decorators, etc.

GitHub at this point!

</details>

will help you out!

- [x] Ensure all automated PR checks (seen at the bottom of the "conversation" tab) pass.

</details>

----
📚 Documentation preview 📚: https://earthaccess--1129.org.readthedocs.build/en/1129/

### DIFF
--- a/earthaccess/api.py
+++ b/earthaccess/api.py
@@ -326,7 +326,9 @@ def login(
     does login fail.
 
     Parameters:
-        strategy: One of the following:
+        strategy:
+            An authentication method.
+
             * `"all"`: try each of the following methods, in order, until one
               succeeds.
             * `"environment"`: retrieve either an Earthdata login token from the


earthaccess/__init__.py
```diff
@@ -86,21 +86,4 @@ def __getattr__(name):  # type: ignore
...
-                            f"Automatic authentication with {strategy=} was successful"
-                        )
-                        break
-                except Exception as e:
-                    logger.debug(
-                        f"An error occurred during automatic authentication with {strategy=}: {str(e)}"
-                    )
-
-        return _auth if name == "__auth__" else _store
+    return _auth if name == "__auth__" else _store
```
betolink: this will remove the authentication on import, right? making `earthaccess.login()` required before accessing a file.

CHANGELOG.md
```diff
@@ -788,29 +802,29 @@ _Conception!_
...
+[0.8.2]: https://github.com/nsidc/earthaccess/releases/tag/v0.8.2
+[0.9.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.9.0
+[0.10.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.10.0
+[0.11.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.11.0
+[0.12.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.12.0
+[0.13.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.13.0
+[0.14.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.14.0
+[0.15.0]: https://github.com/nsidc/earthaccess/releases/tag/v0.15.0
+[0.15.1]: https://github.com/nsidc/earthaccess/releases/tag/v0.15.1
+[Unreleased]: https://github.com/nsidc/earthaccess/compare/v0.15.1...HEAD
```
mfisher87, line 848: I kind of liked the newest-to-oldest order we had. What's the motivation for changing this? I don't want to bikeshed on this portion of the PR though, it's not that important. Maybe it should be separated to another PR.

earthaccess/api.py
```diff
@@ -319,18 +319,28 @@ def login(
 ) -> Auth:
     """Authenticate with Earthdata login (https://urs.earthdata.nasa.gov/).
 
+    Attempt to login via _only_ the specified strategy, unless the `"all"`
+    strategy is used (default), in which case each of the individual strategies
+    is attempted in the following order, until one succeeds: `"environment"`,
+    `"netrc"`, `"interactive"`.  In this case, only when all strategies fail
+    does login fail.
+
```
mfisher87, line 327: Huge improvement :100:

earthaccess/api.py
```diff
@@ -319,18 +319,28 @@ def login(
...
-            * **"all"**: (default) try all methods until one works
-            * **"interactive"**: enter username and password.
-            * **"netrc"**: retrieve username and password from ~/.netrc.
-            * **"environment"**: retrieve either a username and password pair from
-              the `EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD` environment variables,
-              or an Earthdata login token from the `EARTHDATA_TOKEN` environment variable.
-        persist: will persist credentials in a .netrc file
-        system: the Earthdata system to access, defaults to PROD
+        strategy: One of the following:
+            * `"all"`: (default) try each of the following methods, in order,
```
mfisher87: I think we should remove the default indicators from this docstring as those are specified and generated by the type annotations already:

image

We're likely to let this information fall out of sync if we duplicate it.

earthaccess/__init__.py
```diff
@@ -86,21 +86,4 @@ def __getattr__(name):  # type: ignore
...
-                        logger.debug(
-                            f"Automatic authentication with {strategy=} was successful"
-                        )
-                        break
-                except Exception as e:
-                    logger.debug(
-                        f"An error occurred during automatic authentication with {strategy=}: {str(e)}"
-                    )
-
-        return _auth if name == "__auth__" else _store
```
mfisher87, line 106: It's possible some may rely on this implicit/automatic login behavior. I don't think we should necessarily keep this behavior around, but I do think we should call this out as a breaking change in the changelog if we're removing it.

And let's also update the title of this PR to mention the removal of automatic login behavior!

CHANGELOG.md
chuckwondo: The motivation is that as this list continues to grow, it becomes increasingly annoying to get to the correct line to add a new entry when the list is in reverse-chronological order. Whereas when it is in chronological order (the order I modified it to be), the correct line is _always_ the next to last line in the file (along with updating the very last line). It's simply a matter of convenience for updating.

earthaccess/api.py
chuckwondo: Agreed. Will do.

earthaccess/__init__.py
chuckwondo: Yes, it is possible, although I would be very surprised if this broke anybody's code. I don't think I've ever seen any example code where someone did _not_ include a call to `earthdata.login()`.

Regardless, I agree that this does represent a breaking change, as it is certainly possible that someone is (unknowingly) relying upon the (undocumented?) automatic login behavior to successfully download files.

I will also comb the docs to see if we actually document the automatic login behavior. I don't think we do, but if we do, I'll need to update the docs as part of this PR.

CHANGELOG.md
mfisher87: I wonder why we even have this instead of inlining the links :thinking: I guess it doesn't matter right now... please ignore me! :)

chuckwondo: That's a good question. That probably makes much better sense, as I don't think we make multiple references to any of those links.

chuckwondo:
@mfisher87, I updated the changelog with the "breaking change" called out. Let me know how that reads (I simply used the format suggested in common changelog conventions), and if anything else needs tweaking.

chuckwondo:
> The change looks good to me, but I'm curious if you feel we have consensus around removing automatic login behavior. It sounds good to me, but I haven't asked around about it.

Fair question. I've copied others to solicit more input, so I won't merge this PR yet, but here's my take on it all:

Unfortunately, unless I'm missing something, I don't think there's any completely non-breaking way to address the myriad issues we have around auth.

In this case, if we leave the auto-login behavior in place, we remain unable to address the issue this PR resolves. That is, with auto-login as it currently stands, we cannot prevent, for example, attempting login (implicitly) via env vars or netrc when the user specifically wants to ignore them and (explicitly) force _only_ prompting the user for creds.

Of course, that's just one example. Generally speaking, I find this implicit, non-overrideable behavior highly problematic (and may very well be the source of other auth issues/bugs in our backlog).

I believe keeping this _implicit_ behavior is worse than the alternative, which is to remove the auto-login behavior and run the risk of breaking existing code.

In the case of removing the auto-login behavior, there are only 2 scenarios where this may have an impact:

1. **search:** All of our `search_*` functions end up causing auto-login (because they all access `earthaccess.__auth__`, which is what triggers the behavior). The worst case scenario created by removing auto-login is that search results _might_ contain fewer results, but only in cases where specific search criteria would include results available only to certain users. This almost certainly won't break code (i.e., won't raise an exception that would otherwise not be raised), but rather _might_ produce different results.
2. **read:** This is the more concerning scenario. Without being logged in, any attempt to read data will result in an auth exception.

Anecdotally, I suspect _most_ users always call `earthdata.login` first (at least before attempting to read, if not also before attempting to search), so I also suspect there will be few issues opened as a result of problems encountered by removing the auto-login behavior.

However, of the few (I suspect) such breaking scenarios that do arise, none (I suspect) will be related to search, and thus all will be related to read scenarios where users "forget" to call `earthdata.login` before reading. If the error that the user encounters isn't enough for them to instantly realize for themselves that they must add a call to `earthdata.login`, we should be able to quickly help them resolve it.

I have not recently looked at our docs, so this may be well covered, but if and when we do see users raising issues due to removal of auto-login, we may want to review our docs to see if we can improve the explicitness and clarity of our docs on auth/login.

To our benefit in removing this auto-login behavior, the behavior is _not_ documented anywhere (as far as I could tell), so I would be very surprised if anybody is even aware that this behavior exists. (I certainly didn't realize it until recently, and I've even read that very code many times, but never had it sink in that auto-login was happening. Maybe I just have a thicker skull than most.)

Longer term, I want to incrementally continue improving our auth process to further eliminate any remaining _implicit_ behavior that may be problematic, and provide more _explicit_ controls/overrides for users, and relatedly with respect to session management.

cc: @betolink, @jhkennedy, @itcarroll

betolink:
Thanks Chuck! I've used the implicit behavior knowing it's there and I think some users will run into this but with a clear explanation I think we are OK.

mfisher87:
I'm 100% sold on this change, but I do want to say I'm most worried about the latter part of this statement:

> This almost certainly won't break code (i.e., won't raise an exception that would otherwise not be raised), but rather _might_ produce different results

If any users are relying on implicit login to turn up restricted data as part of their search, this could lead to a sneaky bug. Ideally, they're only searching for restricted data, and so they go from N results to 0 results, and they'd instantly know there's a problem. But if they are searching for both restricted and unrestricted data, they would go from N to <N results and their analysis may be unknowingly incorrect. Maybe the best thing we can do about this is (1) provide a (very simple) migration guide and (2) announce the release on Zulip and Slack with a warning and link to the migration guide.

chuckwondo:
> I'm 100% sold on this change, but I do want to say I'm most worried about the latter part of this statement:
> 
> > This almost certainly won't break code (i.e., won't raise an exception that would otherwise not be raised), but rather _might_ produce different results
> 
> If any users are relying on implicit login to turn up restricted data as part of their search, this could lead to a sneaky bug. Ideally, they're only searching for restricted data, and so they go from N results to 0 results, and they'd instantly know there's a problem. But if they are searching for both restricted and unrestricted data, they would go from N to <N results and their analysis may be unknowingly incorrect. Maybe the best thing we can do about this is (1) provide a (very simple) migration guide and (2) announce the release on Zulip and Slack with a warning and link to the migration guide.

Agreed. This is perhaps the worse of the 2 cases because of its subtlety. At least the other case will be immediately obvious by raising an exception.

Hopefully this "reduced search results" case will be a rare case, but I agree that we should provide greater visibility and clarity around this change along the lines of your suggestions to reduce the likelihood of surprises without having a place to point folks to enable quick rectification.

Would you be open to allowing this PR to land (without the guide) so that it allows tackling other auth issues possibly in parallel to developing the short migration guide you're suggesting? If so, would you mind opening a new "docs" issue for the guide and coordinating announcement?

We could hold off on any new version release until the migration guide also lands.

chuckwondo:
> Thanks Chuck! I've used the implicit behavior knowing it's there and I think some users will run into this but with a clear explanation I think we are OK.

@betolink, I would say that you might not be a representative user in this case because you're the one who wrote it (and thus are certainly aware of the undocumented behavior) 😄

mfisher87:
> Would you be open to allowing this PR to land (without the guide) so that it allows tackling other auth issues possibly in parallel to developing the short migration guide you're suggesting? If so, would you mind opening a new "docs" issue for the guide and coordinating announcement?

Sounds good, on it!

CHANGELOG.md
jhkennedy: It's "preferred" by common-changelog: 

> Preferably use [reference-style links](https://daringfireball.net/projects/markdown/syntax#link) to keep the unrendered Markdown form of the changelog readable.

https://common-changelog.org/#22-release

PR #1144, opened 2025-11-14
Prevent formatting symlink file

### LINKED ISSUE #1143: [BUG] pre-commit on Windows modifying main branch files

### Current Behavior

When `pre-commit run -a` on a main branch on a Windows platform the yml files are getting modified.
On Linux fies are staying unchanged.

Similar to: https://github.com/google/yamlfmt/issues/263

Changed files:
- .pre-commit-config.yaml
- mkdocs.yml
- publish.yml
- LICENSE.txt

`pre-commit` version: 4.4.0
latest `main` commit was run at: 589e79ec90dfaabdf6d6a8f06eef53307019b2c8

### Expected Behavior

Files from `main` remain unchanged on Windows.

### Steps To Reproduce

Fetch `main` on Windows.
Setup conda environment.
Install `pre-commit`
Run `pre-commit run -a`

### Environment

```markdown
- OS: Windows 11 Home, OS build 26200.6899
- Python: 3.14.0
```

### Additional Context

_No response_

ana-sher:
Prevent fixing symlink file
resolves #1143 LICENSE.txt fixing issue with end-of-file-fixer hook on Windows

----
📚 Documentation preview 📚: https://earthaccess--1144.org.readthedocs.build/en/1144/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 - Use only specified login strategy to attempt login, when strategy other than
   "all" is specified ([#945](https://github.com/nsidc/earthaccess/issues/945))
   (@chuckwondo)
+- Fix undesirable pre-commit changes when running on Windows
+  ([#1143](https://github.com/nsidc/earthaccess/issues/1143)) (@ana-sher)
 
 ## [0.15.1] - 2025-09-16
 


.pre-commit-config.yaml
```diff
@@ -13,18 +13,19 @@ repos:
         exclude: ".*/vcr_cassettes/.*\\.yaml"
       - id: check-toml
       - id: check-json
-      # Check for common mistakes
       - id: check-added-large-files
+        # Check for common mistakes
```
chuckwondo: Thanks @ana-sher! It looks like you may have forgotten to skip the yamlfmt hook when you committed your change, but I think that's because you used VSCode to commit this change, rather than making the commit from the command line, where you would be able to specify `SKIP`.

Can you revert the changes in this file, other than the single line for excluding `docs/LICENSE.txt`?

chuckwondo:
@ana-sher, thanks again for your first contribution!

It looks like your branch is out of date with the `main` branch. Can you pull in the latest changes from main?

.pre-commit-config.yaml
ana-sher: Thank you @chuckwondo, yes, that's exactly what happened 💯  Reverted and fixed by running from the command line.

chuckwondo:
Much better :-)

I think I found a solution to the yamlfmt issue, if you want to give it a try. It works for me, but that's because I'm running macOS, so if you want to try the following on your Windows machine, that would be great.

If it works, I'd like you to add it to this PR so that it is a complete fix for #1143.

Please do the following:

First, create a file named `scripts/yamlfmt.py` (the `scripts` directory already exists in the repo), with the following contents:

```python
"""Helper script that automatically skips the yamlfmt pre-commit hook on Windows.

This is not intended to be called directly.  It is configured to be called by
pre-commit via the configuration in .pre-commit-config.yml for the yamlfmt hook.

Due to https://github.com/google/yamlfmt/issues/263, comments in yaml files are
not handled properly under Windows, so any Windows user that causes yamlfmt to
format yaml files will cause CI build failure because we build under Linux, and
the resulting format will differ, causing file changes during the pre-commit
step, causing build failure.

Therefore, this script avoids calling yamlfmt (via pre-commit) when running on
Windows.
"""
import platform
import subprocess
import sys

if platform.system() != "Windows":
    sys.exit(subprocess.call(["yamlfmt", *sys.argv[1:]]))
```

Then, modify the `yamlfmt` hook configuration in `.pre-commit-config.yml` to look like so:

```yaml
  - repo: https://github.com/google/yamlfmt
    rev: v0.20.0
    hooks:
      - id: yamlfmt
        entry: python scripts/yamlfmt.py
        types_or: [yaml]
        exclude: ".*/vcr_cassettes/.*\\.yaml"
```

Notice that I've added `entry` to that section, which will call the script above, rather than calling `yamlfmt` directly.

Once you've added the new script and modified the pre-commit config, try running `pre-commit run -a` again (_without_ setting `SKIP`) to see if it leaves the yaml files unchanged. If so, then please commit the new changes to this PR.

ana-sher:
It is a good way to solve it until https://github.com/google/yamlfmt/issues/263 will be fixed! Worked as intended on my Windows machine, thank you @chuckwondo , added script to the PR.

chuckwondo:
> It is a good way to solve it until [google/yamlfmt#263](https://github.com/google/yamlfmt/issues/263) will be fixed! Worked as intended on my Windows machine, thank you @chuckwondo , added script to the PR.

Fantastic! I'm glad that works.

I think the only thing left to do is add an entry to CHANGLOG.md indicating your contribution!

Under the "Unreleased" heading (near the top of CHANGELOG.md), there is a "Fixed" subheading. Please add the following item for your fix:

```plain
- Fix undesirable pre-commit changes when running on Windows
  ([#1143](https://github.com/nsidc/earthaccess/issues/1143)) (@ana-sher)
```

cc: @danielfromearth

ana-sher:
Done :white_check_mark: Thanks a lot for the mentoring on my first contribution, @chuckwondo!

danielfromearth:
I think this looks great. The entry script is a clever way handle the issue!

If/when `yamlfmt` is updated so there are no longer differences between Operating Systems, this `scripts/yamlfmt.py` script might linger on yet then be unnecessary.  I'm not sure if there is a way, but can we set something up so we would be alerted if we no longer need this script?

chuckwondo:
> I think this looks great. The entry script is a clever way handle the issue!
> 
> If/when `yamlfmt` is updated so there are no longer differences between Operating Systems, this `scripts/yamlfmt.py` script might linger on yet then be unnecessary. I'm not sure if there is a way, but can we set something up so we would be alerted if we no longer need this script?

I've subscribed to the yamlfmt issue, so I'll get notified of any future activity on the thread. If you want to do the same, then at least a few of us will see if/when it gets resolved.

If/when the yamlfmt issue if fixed, we can then open an issue to remove this newly added script along with the `entry` config added to `.pre-commit-config.yml`. It should be fairly obvious that if the `entry` is removed for the hook, that the script it refers to should also be removed.

danielfromearth:
Sounds good to me. I just now subscribed to that issue thread too 👀

PR #1166, opened 2025-12-10
Eula error handling

### LINKED ISSUE #36: Error messages when user hasn't accepted dataset EULA can be confusing

ana-sher:
Resolves #36
Contains:
- @betolink unit test that is reproducing EULA error (comes in `response.text`, with html that includes "eula" error inside)
- fix in `_download_file`, because if catching error further it won't contain `text` or `content`, will only raise error for given status code.

- [x] Please review our
      before getting started.
      example `closes #1`. See
- [x] Update the documentation and/or the `README.md` with details of changes to the
      decorators, etc.

GitHub at this point!

</details>

will help you out!

- [x] Add unit tests for any new features.

</details>

----
📚 Documentation preview 📚: https://earthaccess--1166.org.readthedocs.build/en/1166/

### DIFF
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -11,7 +11,7 @@ import responses
 import s3fs
 from earthaccess import Auth, Store
 from earthaccess.auth import SessionWithHeaderRedirection
-from earthaccess.exceptions import EulaNotAccepted
+from earthaccess.exceptions import DownloadFailure, EulaNotAccepted
 from earthaccess.store import EarthAccessFile, _open_files
 from pqdm.threads import pqdm
 
@@ -62,7 +62,31 @@ class TestEula(unittest.TestCase):
             status=401,
         )
         store = Store(self.auth)
-        with self.assertRaises(EulaNotAccepted):
+        with self.assertRaisesRegex(
+            EulaNotAccepted, f"Eula Acceptance Failure for {mocked_url}"
+        ):
+            store.get([mocked_url], "/tmp")
+
+    @responses.activate
+    def test_detects_non_eula_errors(self):
+        response = " blah blah error!"
+        mocked_url = "https://example.com/protected_file.nc"
+        responses.add(
+            responses.GET,
+            "https://urs.earthdata.nasa.gov/profile",
+            json={},
+            status=200,
+        )
+        responses.add(
+            responses.GET,
+            url=mocked_url,
+            body=response,
+            status=401,
+        )
+        store = Store(self.auth)
+        with self.assertRaisesRegex(
+            DownloadFailure, f"Download failed for {mocked_url}. Status code: 401"
+        ):
             store.get([mocked_url], "/tmp")
 
     def tearDown(self):


earthaccess/exceptions.py
```diff
@@ -45,3 +45,13 @@ class ServiceOutage(Exception):
     """
 
     pass
+
+
+class EulaException(Exception):
```
chuckwondo: @betolink, what do you think about making this a subclass of `DownloadFailure`?

earthaccess/store.py
```diff
@@ -850,6 +851,13 @@ def _download_file(self, url: str, directory: Path) -> Path:
...
             session = self.thread_locals.local_thread_session
             with session.get(url, stream=True, allow_redirects=True) as r:
+                if r.status_code in [401, 403]:
+                    text = (r.text or "").lower()
+                    if "eula" in text:
+                        raise EulaException(
+                            "You must accept the EULA to download this data."
+                        )
+
                 r.raise_for_status()
```
chuckwondo: If the response text contains "eula" (ignoring case), we should use that text as the error message, rather than making up our own message.

Also, if the response is not successful, but the text does not include "eula", we want to raise a general DownloadFailure (which seems to have been accidentally left out as part of the previous PR that added the DownloadFailure exception class -- i.e., the class was added, but wasn't used here, as it should have been).

```suggestion
                if not r:
                    if "eula" in (text := r.text or "").lower()
                        raise EulaException(text)
                    raise DownloadFailure(text)
```

tests/unit/test_store.py
```diff
@@ -11,10 +11,79 @@
...
+        )
+        responses.add(
+            responses.GET,
+            url=mocked_url,
+            body=response,
+            status=401,
+        )
+        store = Store(self.auth)
+        # test
+        with self.assertRaises(EulaException):
```
chuckwondo:
```suggestion
        with self.assertRaises(EulaException, msg=response):
```

tests/unit/test_store.py
```diff
@@ -11,10 +11,79 @@
...
+        responses.add(
+            responses.POST,
+            "https://urs.earthdata.nasa.gov/api/users/find_or_create_token",
+            json=json_response,
+            status=200,
+        )
+
+        self.auth = Auth()
+        self.auth.login(strategy="environment")
+        self.assertEqual(self.auth.authenticated, True)
```
chuckwondo: We can use the simpler `assert` syntax:

```suggestion
        assert self.auth.authenticated
```

tests/unit/test_store.py
```diff
@@ -11,10 +11,79 @@
...
+    @responses.activate
+    def test_store_can_create_https_fsspec_session(self):
+        responses.add(
+            responses.GET,
+            "https://urs.earthdata.nasa.gov/profile",
+            json={},
+            status=200,
+        )
+        store = Store(self.auth)
+        self.assertTrue(isinstance(store.auth, Auth))
```
chuckwondo:
```suggestion
        assert isinstance(store.auth, Auth)
```

tests/unit/test_store.py
```diff
@@ -11,10 +11,79 @@
...
+        https_fs = store.get_fsspec_session()
+        self.assertEqual(type(https_fs), type(fsspec.filesystem("https")))
```
chuckwondo:
```suggestion
        assert type(https_fs) == type(fsspec.filesystem("https"))
```

tests/unit/test_store.py
```diff
@@ -11,10 +11,79 @@
...
+            status=200,
+        )
+        responses.add(
+            responses.GET,
+            url=mocked_url,
+            body=response,
+            status=401,
+        )
+        store = Store(self.auth)
+        # test
```
chuckwondo:
```suggestion
```

earthaccess/store.py
ana-sher: We want to include full html page that was returned in `text` to exception message? I could strip it down to the closest ">" message "<" if we are sure it's always HTML.

chuckwondo: Oh, right. I forgot it's returning HTML.

I don't think it's worth trying to parse out the message at runtime. Can you reproduce the response and just manually copy the text of the message and then paste it into the code?

chuckwondo:
@ana-sher, looks like pre-commit failed on one of my suggestions: https://results.pre-commit.ci/run/github/399867529/1765488551.Ty6xULlSTtagswoCnbUJyQ

earthaccess/exceptions.py
betolink: Maybe? one thing I forgot is that this error will also arise when we open the file with `.open()` would that be semactically download error too? @chuckwondo

earthaccess/store.py
betolink: what about logging the content and telling the users that the full response is in the error logs?

chuckwondo: I'm leaning towards _not_ logging the full response, because it is a bunch of HTML, and would likely be annoying and probably too verbose and messy for a user to bother to read to find the error message.

If we can reproduce the HTML response without much effort, I'd say we should do that (like you did during our hack session the other day), find the error message in the HTML response, and copy the error message to put into the code. We may want to slightly adjust the wording, and include the data URL in the message too, so the user knows exactly what failed to be downloaded.

earthaccess/exceptions.py
chuckwondo: I'd say that it would be semantically the same. When opening, we still end up attempting to "download" data. The only difference is _how much_ data is being downloaded (either all of it, or some of it).

In the case of `open`, a user likely won't care or notice whether or not EulaException is a DownloadFailure, but in the case of `download`, it might be nice to be able to catch DownloadFailure and have that also catch EulaException.

Also, now that I think about it, I suggest we rename `EulaException` to `EulaNotAccepted` to make it a bit more clear (note also the intentional _lack_ of the suffix Error or Exception).

ana-sher: If we are having same issue on `.open()`, will try to reproduce and add test on that as well, thank you @betolink.
Might need to intercept and raise in `_open_urls_https` or `_open_files`.

earthaccess/store.py
ana-sher: Was reproducing this issue again and got 403 instead of 401 with `json`, not sure why. Was testing for `S5P_L2__CH4____HiR` (same collection as for hackdays) and same credentials. Also tested for `sentinel-1c_raw` and `ALOS2_L1_PSR2` - same json with eula error was returned. For `ALOS_PSR_RTC_LOW` and `ERS-1_L1` got 401 with text `HTTP Basic: Access denied`, no EULA mentioned...
image
image

I can check content-type in headers and output `error_description` + `resolution_url` + `data url` if `json`. If not - fallback to `"Eula Acceptance Failure"` + `data url`?

earthaccess/exceptions.py
ana-sher: testing:
when calling `earthaccess.open` for 
1. `"S5P_L2__CH4____HiR"` file-like objects were created (maybe further xarray processing would give error).
2. `"sentinel-1c_raw"` 401 returned that was wrapped in `FileNotFoundError` for the file by `fsspec`. By opening that file url in browser - automatically redirected me to accept eula, so giving data url if fsspec returns `FileNotFoundError` might be a good idea. But even fsspec returns it:
```
raise FileNotFoundError(url) from exc
E                   FileNotFoundError: https://datapool.asf.alaska.edu/RAW/SC/S1C_IW_RAW__0SDV_20250328T182304_20250328T182337_001645_002A8D_2E72.zip

C:\Users\dmitr\miniconda3\envs\earthaccess\Lib\site-packages\fsspec\implementations\http.py:451: FileNotFoundError
```
3. `"ALOS2_L1_PSR2"` (also has eula 403 on `download`) file-like objects were created without error.
4. `"ERS-1_L1"` - `FileNotFoundError`
5. `"ALOS_PSR_RTC_LOW"` - `FileNotFoundError`

ana-sher: Suggesting to move `earthaccess.open` EULA error processing to another PR since we are not sure yet how to infer it from returned `FileNotFoundError`

chuckwondo:
@betolink, @ana-sher and I just went over this again in today's hackday, and I think it looks good. I approved it, but wanted you to take a quick look (there's not much code to review) before merging, especially since we started this with you on a previous hackday a couple months ago.

PR #1183, opened 2026-01-27
Always use EDL user token to fetch s3 creds

### LINKED ISSUE #1181: [BUG] Auth instances can fail to recognize that EARTHDATA_TOKEN environment variable is in play

### Have you checked the status of Earthdata services?

- [x] I've executed `earthaccess.status()` and both CMR and EDL returned `'OK'`.

### Current Behavior

During hackternoon investigation of #1136, we noticed an issue created by the introduction of the token environment variable. Specifically, `Auth.get_s3_credentials` may assume that a username and password have been provided when they have not:

https://github.com/nsidc/earthaccess/blob/a4fcc339ff9f188e07d5fe7364b064b0e5f7049b/earthaccess/auth.py#L190-L193

@ana-sher Offered to write a test, maybe a parameterized test, to catch any other instances where use of a token instead of username/passwd is not handled.

@chuckwondo Suggests refactoring the whole auth module to only use token (with one call at the top to get a token via username/passwd if not provided).

### Expected Behavior

- We should have a test that catches this broken path.
- The Auth module should use the `token` property when able.

### Steps To Reproduce

Here's a code snippet that demonstrates the resulting failure and invalid warning.

```python
import os
import earthaccess

os.environ["EARTHDATA_TOKEN"] = # put in a valid token
auth = earthaccess.login(strategy="environment")

    short_name="NLDAS_NOAH0125_H",
    count=1,
)
earthaccess.download(results)
```

### Environment

```markdown
- JUPYTER_IMAGE=openscapes/python:65d6916
```

### Additional Context

_No response_

### LINKED ISSUE #1182: Always use bearer token auth, not basic auth
There is at least one location where an http request is made using the user's basic auth creds (EDL username and password) rather than the user's EDL token: `Auth.get_s3_credentials`. There may be other locations, and if so, they should also be addressed with this issue.

Using basic auth with EDL creds creates a latency penalty due to multiple redirects that it incurs. Using the user's EDL token avoids the redirects and avoids this penalty, which can be critical in applications that require as little latency as possible.

The EDL token should be used everywhere it is applicable to avoid unnecessary latency.

chuckwondo:
This doesn't change any functionality, but speeds up fetching of s3 credentials by about 4x based on very primitive benchmarking, due to always using an EDL (bearer) token for fetching temp creds, rather than using EDL username/password (basic auth):

```python
from timeit import timeit
import earthaccess

auth = earthaccess.login()

# These 2 lines are necessary only when running against
# earthaccess code that does not include the code in this PR,
# due to issue 1181, which this PR also fixes.
setattr(auth, "username", None)
setattr(auth, "password", None)

endpoint = "https://data.nsidc.earthdatacloud.nasa.gov/s3credentials"

print(timeit(lambda: auth.get_s3_credentials(endpoint=endpoint), number=10))
```

Benchmark results running on my pre-M1 MacBook:

- Without this PR: 36.76829030702356 (avg. of ~3.7s per s3 creds fetch)
- With this PR: 8.482286745973397 (avg. of ~0.85s per s3 creds fetch)
- Speedup: 36.76829030702356 / 8.482286745973397 ~= 4.33

The speedup is due to a reduction in the number of requests. In particular, using basic auth ends up making 4 times the number of requests than using a bearer token does, so the 4x speedup jives with reducing the number of requests by a factor of 4.

Fixes #1182
Fixes #1181

- [ ] Please review our
      before getting started.
      example `closes #1`. See
      decorators, etc.

GitHub at this point!

</details>

will help you out!

</details>

----
📚 Documentation preview 📚: https://earthaccess--1183.org.readthedocs.build/en/1183/

### DIFF
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -321,9 +321,7 @@ class Auth(object):
             logger.info("You're now authenticated with NASA Earthdata Login")
 
             token = token_resp.json()
-            logger.debug(
-                "Using token with expiration date: %s", token["expiration_date"]
-            )
+            logger.info("Using token with expiration date %s", token["expiration_date"])
             self.token = token
             self.authenticated = True
 


earthaccess/auth.py
```diff
@@ -99,7 +99,9 @@ class Auth(object):
     def __init__(self) -> None:
         # Maybe all these predefined URLs should be in a constants.py file
         self.authenticated = False
-        self.token: Optional[Mapping[str, str]] = None
+        self.token: Mapping[str, str] | None = None
+        self.username: str | None = None
+        self.password: str | None = None
```
chuckwondo, line 104: This is the crux of the fix for #1181, which also helps with the rest of the changes for #1182.

earthaccess/search.py
```diff
@@ -260,7 +259,9 @@ def parameters(self, **kwargs: Any) -> Self:
     def print_help(self, method: str = "fields") -> None:
         """Prints the help information for a given method."""
         print("Class components: \n")  # noqa: T201
-        print([method for method in dir(self) if method.startswith("_") is False])  # noqa: T201
+        print(  # noqa: T201
+            [method for method in dir(self) if method.startswith("_") is False]
+        )
```
chuckwondo, line 264: Ruff ended up reformatting, which then put the noqa comment on the wrong line, which then made ruff complain about the print call, so this is just to satisfy ruff.

earthaccess/auth.py
```diff
@@ -328,30 +309,28 @@ def _get_credentials(
...
                 msg = f"Authentication with Earthdata Login failed with:\n{token_resp.text}"
                 logger.error(msg)
                 raise LoginAttemptFailure(msg)
 
             logger.info("You're now authenticated with NASA Earthdata Login")
-            self.username = username
-            self.password = password
 
             token = token_resp.json()
             logger.debug(
```
betolink: Maybe we should "escalate" this to info or warning? I think we should always see this when we activate the logging.

earthaccess/auth.py
```diff
@@ -328,30 +309,28 @@ def _get_credentials(
...
+                "Using token with expiration date: %s", token["expiration_date"]
             )
             self.token = token
             self.authenticated = True
 
         return self.authenticated
 
-    def _find_or_create_token(self, username: str, password: str) -> requests.Response:
-        with SessionWithHeaderRedirection(
-            self.system.edl_hostname, (username, password)
```
betolink, line 353: what happens if we are using a different environment like UAT?  is `self.system` no longer necessary here?

chuckwondo: Yep, it's still necessary. The same line appears in the `get_session` method: https://github.com/nsidc/earthaccess/blob/abda5ca53fee36ccd96876f62292d3b28b09fb05/earthaccess/auth.py#L230, so this change simply avoids duplicate code.

earthaccess/auth.py
chuckwondo: Sure, I can bump this to info...

PR #1198, opened 2026-02-06
Download files with temporary names

### LINKED ISSUE #1131: Earthaccess should allow you to re-download and replace a file
When we download an "on-prem" file or a cloud file, we skip downloading any file that already exists on disk:
* on-prem: https://github.com/nsidc/earthaccess/blob/main/earthaccess/store.py#L859-L860
* cloud: https://github.com/nsidc/earthaccess/blob/main/earthaccess/store.py#L702-L703

(logging behavior is also different between the two 🫤 ) 

However, there are cases where a user may want to download an replace a local file (e.g., manipulated the file, partial downloads).

WeatherGod:
* Also add a force kwarg in many places
* Closes #1131

- [x] Please review our
      before getting started.
      example `closes #1`. See
      decorators, etc.

GitHub at this point!

</details>

will help you out!

- [x] Add unit tests for any new features.

</details>

----
📚 Documentation preview 📚: https://earthaccess--1198.org.readthedocs.build/en/1198/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Changed
 
-- Local files are saved with ".part" suffixes while downloading.
+- While downloading, files will have temporary names until download completes.
 
 ### Added
 
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -3,6 +3,7 @@ import logging
 import tempfile
 import threading
 import traceback
+from collections.abc import Generator
 from contextlib import contextmanager
 from functools import lru_cache
 from itertools import chain
@@ -183,7 +184,7 @@ def _get_url_granule_mapping(
 
 
 @contextmanager
-def _sibling_tempfile(sibling: Path) -> Path:
+def _sibling_tempfile(sibling: Path) -> Generator[Path, None, None]:
     """Save the file to a temporary name in case of incomplete download.
     Temp file is in the same directory to guarantee atomic replacement,
     and is automatically removed upon exit of the context manager if the
@@ -196,8 +197,8 @@ def _sibling_tempfile(sibling: Path) -> Path:
 
     temp_fh = tempfile.NamedTemporaryFile(
         dir=sibling.parent,
-        prefix="delete_me_",  # In case auto-delete fails, make it obvious to users
-        delete_on_close=False,
+        prefix="partial_",  # In case auto-delete fails, make it obvious to users
+        delete=False,
     )
     temp_path = Path(temp_fh.name)
 
@@ -920,12 +921,14 @@ class Store(object):
                         f"Download failed for {url}. Status code: {r.status_code}"
                     )
 
-                with _sibling_tempfile(path) as temp_path:
-                    with open(temp_path, "wb") as f:
-                        # Cap memory usage for large files at 1MB per write to disk per thread
-                        # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
-                        for chunk in r.iter_content(chunk_size=1024 * 1024):
-                            f.write(chunk)
+                with (
+                    _sibling_tempfile(path) as temp_path,
+                    open(temp_path, "wb") as f,
+                ):
+                    # Cap memory usage for large files at 1MB per write to disk per thread
+                    # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
+                    for chunk in r.iter_content(chunk_size=1024 * 1024):
+                        f.write(chunk)
         else:
             logger.info(f"File {local_filename} already downloaded")
         return path
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -1,5 +1,7 @@
 import logging
+import operator as op
 import os
+from collections.abc import Callable
 from pathlib import Path
 from unittest.mock import patch
 
@@ -189,8 +191,16 @@ def test_download(tmp_path, selection, use_url):
     assert all(Path(f).exists() for f in files)
 
 
-@pytest.mark.parametrize("force", [True, False])
-def test_force_download(tmp_path, force):
+@pytest.mark.parametrize(
+    ("force", "cmp"),
+    [
+        # Redownloading should update all of the mtimes, so first list of mtimes will be less than second list of mtimes
+        (True, op.lt),
+        # No forced downloading, so first list of mtimes will be the same as the second list of mtimes
+        (False, op.eq),
+    ],
+)
+def test_force_download(tmp_path, force: bool, cmp: Callable[[float, float], bool]):
     results = earthaccess.search_data(
         count=2,
         short_name="ATL08",
@@ -201,20 +211,14 @@ def test_force_download(tmp_path, force):
     assert isinstance(files, list)
     assert all(Path(f).exists() for f in files)
 
-    # Verify force behavior
+    # Make sure no temp files are left behind
+    assert len(os.listdir(tmp_path)) == len(files)
+
+    # Verify force behavior by calling download again and checking mtimes
     first_mtimes = [f.stat().st_mtime for f in files]
     second_files = earthaccess.download(results, str(tmp_path), force=force)
     second_mtimes = [f.stat().st_mtime for f in second_files]
-    if force:
-        # Redownloading should update all of the mtimes
-        assert all(
-            mtime1 < mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
-        )
-    else:
-        # No forced downloading, so no change in any mtimes.
-        assert all(
-            mtime1 == mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
-        )
+    assert all(cmp(*mtime_pair) for mtime_pair in zip(first_mtimes, second_mtimes))
 
 
 def fail_to_download_file(*args, **kwargs):
--- a/tests/integration/test_cloud_download.py
+++ b/tests/integration/test_cloud_download.py
@@ -1,5 +1,8 @@
 import logging
+import operator as op
+import os
 import shutil
+from collections.abc import Callable
 from pathlib import Path
 
 import earthaccess
@@ -117,8 +120,16 @@ def test_earthaccess_can_download_cloud_collection_granules(tmp_path, daac):
             )
 
 
-@pytest.mark.parametrize("force", [True, False])
-def test_multi_file_granule(tmp_path, force):
+@pytest.mark.parametrize(
+    ("force", "cmp"),
+    [
+        # Redownloading should update all of the mtimes, so first list of mtimes will be less than second list of mtimes
+        (True, op.lt),
+        # No forced downloading, so first list of mtimes will be the same as the second list of mtimes
+        (False, op.eq),
+    ],
+)
+def test_multi_file_granule(tmp_path, force: bool, cmp: Callable[[float, float], bool]):
     # Ensure granules that contain multiple files are handled correctly
     granules = earthaccess.search_data(short_name="HLSL30", count=1)
     assert len(granules) == 1
@@ -127,17 +138,11 @@ def test_multi_file_granule(tmp_path, force):
     files = earthaccess.download(granules, str(tmp_path))
     assert {Path(f).name for f in urls} == {Path(f).name for f in files}
 
-    # Verify force behavior
+    # Make sure no temp files are left behind
+    assert len(os.listdir(tmp_path)) == len(files)
+
+    # Verify force behavior by calling download again and checking mtimes
     first_mtimes = [f.stat().st_mtime for f in files]
     second_files = earthaccess.download(granules, str(tmp_path), force=force)
     second_mtimes = [f.stat().st_mtime for f in second_files]
-    if force:
-        # Redownloading should update all of the mtimes
-        assert all(
-            mtime1 < mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
-        )
-    else:
-        # No forced downloading, so no change in any mtimes.
-        assert all(
-            mtime1 == mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
-        )
+    assert all(cmp(*mtime_pair) for mtime_pair in zip(first_mtimes, second_mtimes))


WeatherGod:
Haven't yet tried to wrap my head around the unit tests to figure out how we would verify both forced and unforced behavior.

jhkennedy:
I'll try to make time to review this soon, but I will re-emphasise I'd prefer that the `.part` temporary files be downloaded to a temporary location and not the download location, or to have entirely different names. Many down-stream processing tools process products based on the files in the directory, parsing the file names, and not all of them are smart enough to ignore product files appended with `.part`, so they would attempt to process partial files. 

We can eliminate that entirely by using the standard `tempfile` module and let normal OS processes clean up the partial files.

@betolink @chuckwondo @mfisher87 etc. might all have a different opinion on that though.

WeatherGod:
You raise an interesting point about external processes operating on contents of a directory, however under the current behavior, doing that could have resulted in random failures too if one started operating on a file that was in the process of downloading.

There is a fundamental operating system reason for keeping the in-progress downloads in the same directory as the final files: atomic renames. If the in-progress downloads are in a different location, then there is no guarantee that the rename will be atomic and may result in a copy, thereby leading to the exact same possible problems as before (as well as introducing another possible moment of running out of storage space).

As for doing NamedTemporaryFile or not, I have two concerns. First, I don't think automatic cleanup is all that great. Not having partial downloads make it a bit harder to know a failure happened, and a bit harder to debug. Second, I have had issues in the past with NamedTemporaryFile on Windows systems. IIRC, on Windows, I needed to close the file object in order for another function to be able to open it, which negates the auto-cleanup ability of NamedTemporaryFile. I've also had issues with permissions as well.

WeatherGod:
I've added a few integration tests, but not any unit tests, which given the amount of mocking I'm seeing in there, I might have to make my own unit tests rather than hitching on existing tests. Not sure, though.

WeatherGod:
And, to be clear, I'm not against some of the other ideas. I think this is a good starting point and having something concrete to play with lets us figure out how much more complicated we need to make this.

earthaccess/store.py
```diff
@@ -692,13 +716,22 @@ def _get(
...
-        self, s3_fs: fsspec.AbstractFileSystem, file: str, local_path: Path
+        self,
+        s3_fs: fsspec.AbstractFileSystem,
+        file: str,
+        local_path: Path,
+        force: bool = False,
     ) -> Path:
         file_name = local_path / Path(file).name
-        if file_name.exists():
+        if file_name.exists() or force:
```
```suggestion
        if file_name.exists() and not force:
```

chuckwondo:
@WeatherGod, thanks for taking a stab at this.

Your concerns regarding temp files and atomic replace are valid, but @jhkennedy raises a valid concern as well about potentially unwanted process triggers due to writing the temporary file to the same directory.

I will review your PR and provide a solution to allow for atomic replace using a temp file in a _different_ directory.

WeatherGod:
Python devs worked hard on the file renaming issue almost 15 years ago (man, I'm old...): https://github.com/python/cpython/issues/53074

There are a lot of edge cases we would have to watch out for. There are only two guaranteed approaches: rename within the same directory, and a rename from a subdirectory that we created so we know it isn't from another volume. There are some packages like https://python-atomicwrites.readthedocs.io/en/latest/. However, they all note problems with Windows support, or don't even bother. I haven't found one yet that says otherwise.

chuckwondo:
True, there's no absolute way to guarantee we can obtain a temp dir on the same device as the target dir, other than using the target dir itself, so perhaps we can use a temp filename within the target directory to eliminate (or at least significantly reduce) the possibility of the temp file triggering any sort of downstream process that @jhkennedy has concerns about. Since such processes typically look for specific filename patterns, using a randomly generated temporary name might be enough assurance to sidestep that concern. What do you think about that @jhkennedy?

(@WeatherGod, I'll still provide feedback on your PR, as I have some suggestions either way.)

jhkennedy:
> so perhaps we can use a temp filename within the target directory...
> using a randomly generated temporary name might be enough assurance to sidestep that concern

@chuckwondo that sounds reasonable to me. I also suggested different file names as an alternative:
> or to have entirely different names.

But I'm happy to let y'all work something out as either way it will be an improvement.

earthaccess/api.py
```diff
@@ -453,6 +456,7 @@ def open(
     show_progress: Optional[bool] = None,
     pqdm_kwargs: Optional[Mapping[str, Any]] = None,
     open_kwargs: Optional[Dict[str, Any]] = None,
+    force: bool = False,
```
chuckwondo: Why did you add this to `open`?

earthaccess/store.py
```diff
@@ -692,13 +704,22 @@ def _get(
...
             return file_name  # Skip if already exists
 
-        s3_fs.get([file], str(local_path), recursive=False)
+        # Save the file with a .part suffix in case of incomplete downloads
+        temp_name = file_name.with_name(file_name.name + ".part")
+        s3_fs.get([file], str(temp_name), recursive=False)
+        # Rename successful downloads to the finalized name.
+        # TODO: Do we want rename() or replace()?
+        temp_name.replace(file_name)
```
chuckwondo: I suggest a bit of a tweak to ensure we guarantee tidying up:

```python
# Save the file to a temporary name in case of incomplete download.
# Temp file is in the same directory to guarantee atomic replacement,
# and is automatically removed upon exit of the context manager if the
# download raises an exception.
with _sibling_tempfile(file_name) as temp:
    temp.close()  # We only want the name, not the file handle.
    s3_fs.get([file], temp.name, recursive=False)
    # Atomically rename successful download to the finalized name.
    Path(temp.name).replace(file_name)
```

where `sibling_tempfile` is defined as follows, and can also be used in the other download function as well:

```python
def _sibling_tempfile(sibling: Path) -> tempfile._TemporaryFileWrapper[bytes]:
    # Make sure the sibling file's parent directory exists.  This is thread-
    # safe, as mkdir is atomic, and only one thread will succeed in making the
    # directory if it does not already exist.  Others succeed due to exist_ok.
    sibling.parent.mkdir(parents=True, exist_ok=True)

    return tempfile.NamedTemporaryFile(
        dir=sibling.parent,
        prefix="delete_me_",  # In case auto-delete fails
        delete_on_close=False,
    )
```

WeatherGod:
Would it make sense to have some human-readable aspect to the temporary filenames, like the dataset shortname? Just thinking out loud.

chuckwondo:
> Would it make sense to have some human-readable aspect to the temporary filenames, like the dataset shortname? Just thinking out loud.

Fair question, but we may have nothing other than the URL in hand. However, even if we always had more information, I would suggest that using any "meaningful" name for the temp file wouldn't matter because the user should never see it since it should always get cleaned up.

Even if something went so horribly awry that the temp file weren't removed, there would be nothing the user could safely do with it anyway, as there would be no guarantee that the dangling temp file would be complete (most likely it would not be).

Therefore, now that I've been rambling, I believe using a "meaningful" name might actually be more harmful because if for some reason the file weren't removed automatically, the user might try to do something with it, which would be no good if the file is incomplete. With a random name, the user would have no idea what it's related to, and thus likely just delete it without attempting to do anything with it. No harm done.

chuckwondo:
We might actually want to use a temp filename prefix like "delete_me_" so that if the temp file were not automatically removed, we're telling the user that they should delete it.

earthaccess/store.py
```diff
@@ -855,11 +885,16 @@ def _download_file(self, url: str, directory: Path) -> Path:
                         f"Download failed for {url}. Status code: {r.status_code}"
                     )
 
-                with open(path, "wb") as f:
+                # Save the file with a .part suffix so that incomplete downloads won't
+                # look like complete downloads, and can still be inspected.
+                temp_path = path.with_name(path.name + ".part")
+                with open(temp_path, "wb") as f:
```
chuckwondo: Make use of `_sibling_tempfile` function I wrote in another commend:

```python
# Save the file to a temporary name in case of incomplete download.
# Temp file is in the same directory to guarantee atomic replacement,
# and is automatically removed upon exit of the context manager if the
# download raises an exception.
with _sibling_tempfile(path) as f:
```

earthaccess/store.py
```diff
@@ -855,11 +885,16 @@ def _download_file(self, url: str, directory: Path) -> Path:
...
                     # Cap memory usage for large files at 1MB per write to disk per thread
                     # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
                     for chunk in r.iter_content(chunk_size=1024 * 1024):
                         f.write(chunk)
+                # Rename successful downloads to the final name.
+                temp_path.replace(path)
```
chuckwondo: We need to indent these one level to make sure they're within the context manager above, otherwise the temp file will be removed (upon exit of the context manager) before we can rename it:

```suggestion
                    # Close the temp file and rename it to the final name.
                    f.close()
                    Path(f.name).replace(path)
```

tests/integration/test_api.py
```diff
@@ -189,6 +189,34 @@ def test_download(tmp_path, selection, use_url):
...
+    if force:
+        # Redownloading should update all of the mtimes
+        assert all(
+            mtime1 < mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
+        )
+    else:
+        # No forced downloading, so no change in any mtimes.
+        assert all(
+            mtime1 == mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
+        )
```
```suggestion
# Move these imports to the top of the file
import operator as op
from collections.abc import Callable

@pytest.mark.parametrize(("force", "cmp"), [(True, op.lt), (False, op.eq)])
def test_force_download(tmp_path, force: bool, cmp: Callable[[float, float], bool]):
    results = earthaccess.search_data(
        count=2,
        short_name="ATL08",
        cloud_hosted=True,
        bounding_box=(-92.86, 16.26, -91.58, 16.97),
    )
    files = earthaccess.download(results, str(tmp_path))#, force=force)
    assert isinstance(files, list)
    assert all(Path(f).exists() for f in files)

    # Make sure no temp files are left behind
    assert len(os.listdir(tmp_path)) == len(files)

    # Verify force behavior
    first_mtimes = [f.stat().st_mtime for f in files]
    second_files = earthaccess.download(results, tmp_path, force=force)
    second_mtimes = [f.stat().st_mtime for f in second_files]
    assert all(cmp(*mtime_pair) for mtime_pair in zip(first_mtimes, second_mtimes))
```

tests/integration/test_cloud_download.py
```diff
@@ -117,11 +117,27 @@ def test_earthaccess_can_download_cloud_collection_granules(tmp_path, daac):
             )
 
 
-def test_multi_file_granule(tmp_path):
+@pytest.mark.parametrize("force", [True, False])
+def test_multi_file_granule(tmp_path, force):
```
chuckwondo:
```suggestion
@pytest.mark.parametrize(("force", "cmp"), [(True, op.lt), (False, op.eq)])
def test_multi_file_granule(tmp_path, force: bool, cmp: Callable[[float, float], bool]):
```

tests/integration/test_cloud_download.py
```diff
@@ -117,11 +117,27 @@ def test_earthaccess_can_download_cloud_collection_granules(tmp_path, daac):
...
+    if force:
+        # Redownloading should update all of the mtimes
+        assert all(
+            mtime1 < mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
+        )
+    else:
+        # No forced downloading, so no change in any mtimes.
+        assert all(
+            mtime1 == mtime2 for mtime1, mtime2 in zip(first_mtimes, second_mtimes)
+        )
```
chuckwondo:
```suggestion
    assert all(cmp(*mtime_pair) for mtime_pair in zip(first_mtimes, second_mtimes))
```

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
...
 ## [Unreleased]
 
+### Changed
+
+- Local files are saved with ".part" suffixes while downloading.
+
+### Added
+
+- Added `force` kwarg to `download()` to force redownloads.
+
```
chuckwondo, line 17: I've removed the "changed" bit because that's an implementation detail that simply makes the download process more robust.

```suggestion
### Added

- Add `force` kwarg to `download()` (`False` by default) to allow forcing re-downloads to same destination
  ([#1131](https://github.com/nsidc/earthaccess/issues/1131)) (@WeatherGod)
```

earthaccess/api.py
WeatherGod: My original reading of `store.py` was that the open functions were also using the download code, so I wired it up through `open()` for completeness. I then realized that I misread it and started taking it back out. I forgot to include that for `api.py` in the commit.

CHANGELOG.md
WeatherGod: While it is an implementation detail, it is a behavior change intended for improved user experience, which is worthwhile showcasing somehow, maybe even specifically referencing #1131 by saying "Can now redownload incomplete downloads without manual intervention".

tests/integration/test_api.py
WeatherGod: Not a fan of this suggestion. I've always believed that clarity and simplicity are the most important features of a test. Not even sure this actually saves us any lines of code.

earthaccess/store.py
WeatherGod: If we `yield`, then we can do a `try...except` and clean up on exceptions, too. A little worried that "delete_me_" suffix might be taken too eagerly (by both humans and AI agents).

WeatherGod: Also, do we want to put this at the top of `store.py`, or would this make sense in a new file like `utils/_sibling.py` or something like that? I also have some unit tests from one of my own projects that can help exercise this function.

WeatherGod: I pushed up a contextmanager-based approach similar to one I've successfully used in other projects over the years. Should be a lot cleaner. I put it in `store.py` for now, but we can move it elsewhere if needed. I'm still -0 on the "delete_me_" prefix.

WeatherGod:
I do not know what ruff is complaining about. I ran `ruff format` on the codebase and I'm not seeing anything changed.

WeatherGod:
Ah, needed to run `ruff check` in addition to `ruff format`.

earthaccess/store.py
```diff
@@ -692,13 +743,18 @@ def _get(
...
     ) -> Path:
         file_name = local_path / Path(file).name
-        if file_name.exists():
-            return file_name  # Skip if already exists
+        if file_name.exists() and not force:
+            return file_name  # Skip if already exists and not forcing re-download
 
-        s3_fs.get([file], str(local_path), recursive=False)
+        with _sibling_tempfile(file_name) as temp_name:
+            s3_fs.get([file], str(temp_name), recursive=False)
```
WeatherGod, line 758: I should point out the subtle change here, specifying the destination file rather than the destination path. Pretty sure this should work the same way as before.

WeatherGod:
I think GitHub is acting up today, doubling up on some of my replies. Sorry if it is spamming.

earthaccess/store.py
chuckwondo: I like your enhancement to `_sibling_tempfile` to also do the replacement. 👍

earthaccess/store.py
```diff
@@ -855,11 +920,12 @@ def _download_file(self, url: str, directory: Path) -> Path:
...
-                    # Cap memory usage for large files at 1MB per write to disk per thread
-                    # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
-                    for chunk in r.iter_content(chunk_size=1024 * 1024):
-                        f.write(chunk)
+                with _sibling_tempfile(path) as temp_path:
+                    with open(temp_path, "wb") as f:
+                        # Cap memory usage for large files at 1MB per write to disk per thread
+                        # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
+                        for chunk in r.iter_content(chunk_size=1024 * 1024):
+                            f.write(chunk)
```
chuckwondo: Feel free to ignore this suggestion, but if you want to remove a level of indentation, this does the trick:

```python
                with _sibling_tempfile(path) as temp_path, open(temp_path, "wb") as f:
                    # Cap memory usage for large files at 1MB per write to disk per thread
                    # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
                    for chunk in r.iter_content(chunk_size=1024 * 1024):
                        f.write(chunk)
```

Alternatively, if you want to stack things:

```python
                with (
                    _sibling_tempfile(path) as temp_path,
                    open(temp_path, "wb") as f,
                ):
                    # Cap memory usage for large files at 1MB per write to disk per thread
                    # https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content
                    for chunk in r.iter_content(chunk_size=1024 * 1024):
                        f.write(chunk)
```

tests/integration/test_api.py
chuckwondo: Yes, my suggestion is arguably a bit less readable, and is perhaps on the pedantic end of the scale.

I'm not suggesting it as a means to reduce lines of code, but rather as a means to create a more cohesive approach to the parametrization.

As you've written, you have effectively only half parametrized the function because both the `force` flag and the comparison function constitute the entirety of the parametrization. By parameterizing only the `force` flag, and then using the flag to explicitly determine test logic, you've "split" the parameter.

Again, my suggestion is leaning towards the pedantic end of the scale, but that's how I would write it.

If you feel very strongly against this suggestion, I'm not fussed if you choose not to incorporate it.

chuckwondo: If you choose not to incorporate this suggestion, please at least add the type annotation to the `force` parameter. I'm encouraging inclusion of good typing in our test functions, at least as we add new ones.

CHANGELOG.md
chuckwondo: I don't totally agree, but not fussed if you choose to include it. If you keep it, please update the description appropriately, since we have changed the logic to do something different than using a `.part` suffix. Perhaps just use more generic language about using a temp file.

chuckwondo (changes requested):
This is looking good!

I have a few more comments, which you may take or leave, at your discretion.

I essentially approve this PR, but for the moment, marking as requesting changes, as I would like at least one other maintainer to take a look too. Perhaps @jhkennedy, @betolink, or @mfisher87.

Do you want to take this out of Draft mode?

I'm going to rerun the blocked integration tests to see if they pass (I imagine they will).

earthaccess/store.py
```diff
@@ -180,6 +182,43 @@ def _get_url_granule_mapping(
...
+    )
+    temp_path = Path(temp_fh.name)
+
+    try:
+        # On some systems, a process can't have two handles open for writing to the same file,
+        # so we need to close the file (hence why we don't "delete_on_close" above).
+        temp_fh.close()
+        yield temp_path
+        # Atomically rename successful uses of the temporary file
+        temp_path.replace(sibling)
```
chuckwondo, line 211: Ugh! The `delete_on_close` parameter was added to python 3.12 (and it defaults to `True`), but we still support 3.11, so I think we'll need to tweak this (int. tests for 3.11 fail since `delete_on_close` is not a kw param).

I think this might do the trick:

```suggestion
    temp_fh = tempfile.NamedTemporaryFile(
        dir=sibling.parent,
        prefix="delete_me_",  # In case auto-delete fails, make it obvious to users
    )
    temp_path = Path(temp_fh.name)

    try:
        # On some systems, a process can't have two handles open for writing to the same
        # file, so we need to close the file, but we need to do this directly on the
        # wrapped file. If we call temp_fh.close(), the temp file will be immediately
        # deleted on python 3.12 or later. This is because in 3.12, a `delete_on_close`
        # keyword parameter was added to NamedTemporaryFile, and it defaults to True.
        # However, we still support python 3.11, so we cannot specify the keyword
        # parameter to set it to False.  Closing the wrapped file directly avoid this
        # problem.
        temp_fh.file.close()
        yield temp_path
        # Atomically rename successful uses of the temporary file
        temp_path.replace(sibling)
```

chuckwondo:
@WeatherGod, I added a review comment about needing to adjust `_sibling_tempfile` to accommodate python 3.11, but there are other issues causing int. tests to fail. I'll take a look to see what's going on, but would appreciate other eyes on the failures to help debug: https://github.com/nsidc/earthaccess/actions/runs/21843450130/job/63442124893

cc: @jhkennedy, @betolink, @mfisher87

chuckwondo:
> @WeatherGod, I added a review comment about needing to adjust `_sibling_tempfile` to accommodate python 3.11, but there are other issues causing int. tests to fail. I'll take a look to see what's going on, but would appreciate other eyes on the failures to help debug: https://github.com/nsidc/earthaccess/actions/runs/21843450130/job/63442124893
> 
> cc: @jhkennedy, @betolink, @mfisher87

Ah, actually, I think the additional issue is that there are some "on prem" collections that may no longer be "on prem," so we may simply need to remove/change them in the int. tests. I'll dig a big deeper on this.

Therefore, I believe the only issue with the code in this PR is what I described in the code comment on `_sibling_tempfile`.

earthaccess/store.py
WeatherGod: We don't need that. Just set `delete=False`. I actually didn't notice the `delete_on_close` parameter, confusing it for the `delete` parameter I've used since python 2.7. We are managing the delete ourselves with the sibling context manager.

WeatherGod:
If you don't mind, I would also like to change the prefix from "delete_me_" to "partial_" or something. I'm really honestly worried about an AI agent scanning a directory and seeing the filenames as instructions.

earthaccess/store.py
WeatherGod: Heh, was so focused on bolting this feature on that I didn't see the obvious joining of the `with`'s.

tests/integration/test_api.py
WeatherGod: I'll meet you halfway here and add comments to the test cases to make it clear why the comparison should be that way.

earthaccess/store.py
chuckwondo: Ah, yes, that should do the trick more simply. 👍

chuckwondo:
> If you don't mind, I would also like to change the prefix from "delete_me_" to "partial_" or something. I'm really honestly worried about an AI agent scanning a directory and seeing the filenames as instructions.

Sure, no problem, or even just drop the prefix altogether, but I'm fine with avoiding "delete_me_" however you see fit.

chuckwondo (changes requested):
Looks great!

For the moment, I'm leaving marked as "request changes" while I look more into the failing int. tests. However, with the delete_on_close removal, I don't think any remaining failures will be related to the changes in the PR.

chuckwondo:
Looks like some recent merges occurred on the main branch. Would you mind bringing your branch up to date?

chuckwondo:
@WeatherGod, I fixed the broken int. tests and merged to the main branch. Please update your branch with the latest changes on the main branch, then I'll re-trigger the int. tests for you.

chuckwondo:
Everything passed!

@jhkennedy, @betolink, and @mfisher87, final chance to look over this PR before I merge it.

PR #1212, opened 2026-02-15
Re-organize docs into top-level user/contributor sections

mfisher87:
@saberbrasher I think this doesn't completely resolve any of our open issues, but lays the groundwork. If anything, it mostly addresses #1162. What do you think?

* Enable tab navigation to keep the sidebar a bit less overwhelming
* Re-group everything in to two top-level categories. Do you want to know information about the software (overview), how to use the software (user guide), or how to contribute to the software (contributor guide)?
* Keep the rest of the structure the same. This feels like the first step to a larger re-org, and I want this PR to be easy to review!
* Remove unused docs page `results.md`
* Add decision records to docs using a dynamic listing plugin `awesome-nav`. This separates the navigation from the main mkdocs configuration in `.nav.yml` files. It's capable of dynamically generating the nav with glob patterns, which I use to detect and register all decision record documents. This means #1047 will be mergeable and the doc will show up without updating any nav files. This is important to me. I (and apparently "we" as well, since there were more than 1 unlisted page) consistently forget to update the nav when I update the docs.

<details><summary>Pull Request (PR) draft checklist - click to expand</summary>

- [ ] Please review our
      [contributing documentation](https://earthaccess.readthedocs.io/en/latest/contributing/)
      before getting started.
- [ ] Populate a descriptive title. For example, instead of "Updated README.md", use a
      title such as "Add testing details to the contributor section of the README".
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [ ] Populate the body of the pull request with:
    - A clear description of the change you are proposing.
    - Links to any issues resolved by this PR with text in the PR description, for
      example `closes #1`. See
      [GitHub docs - Linking a pull request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
- [ ] Update `CHANGELOG.md` with details about your change in a section titled
      `## Unreleased`. If such a section does not exist, please create one. Follow
      [Common Changelog](https://common-changelog.org/) for your additions.
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [ ] Update the documentation and/or the `README.md` with details of changes to the
      earthaccess interface, if any. Consider new environment variables, function names,
      decorators, etc.

Click the "Ready for review" button at the bottom of the "Conversation" tab in GitHub
once these requirements are fulfilled. Don't worry if you see any test failures in
GitHub at this point!

</details>

[...]

PR #1224, opened 2026-02-20
Fix broken links

mfisher87:
## Description

There are some broken links since moving some files around in #1212 . These weren't caught because the "latest" version of the docs only changed post-merge. This PR fixes those by pointing to the new location or a using a versioned link as appropriate.

---

- [x] Open PR as draft

#### Merge checklist

- [x] PR title is descriptive
- [x] If needed, unit tests added
- [x] All checks passing (comment `pre-commit.ci autofix` if pre-commit is failing)
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1224.org.readthedocs.build/en/1224/

### DIFF
--- a/docs/user/tutorials/emit-earthaccess.ipynb
+++ b/docs/user/tutorials/emit-earthaccess.ipynb
@@ -46,7 +46,7 @@
    "source": [
     "### Searching for the dataset with `.search_datasets()`\n",
     "\n",
-    "> Note: API docs can be found at [earthaccess](https://earthaccess.readthedocs.io/en/stable/api/)"
+    "> Note: See [our API docs](/api/index.md) for details"
    ]
   },
   {


mfisher87:
Stuck on "network unreachable" -- presumably not a problem with the links, maybe with GitHub infrastructure? :\

saberbrasher:
I am still a bit confused about how "stable" and "latest" impact all our links (and keep them running)... TBH, I'm not sure how much I like our default user docs being linked to major version releases (stable) instead of being more "live" (defaulting to latest). Could one of you folks who have been working on this longer than me explain the logic please?

mfisher87:
It's all a tradeoff, there's no one best approach... :(

* stable: 
  * Pro: Our changes to the application in development don't impact the docs people see. So they can be confident that API documentation is accurate to the version they're using, not the version we're working on.
  * Con: Our new how-tos, tutorials, and organizational improvements since the last release won't be visible by default
* latest:
  * Pros/cons: Basically the inverse of previous
* vX.Y.Z:
  * Pro: Permalink to a specific version of docs that you know will never change
  * Con: May need to update the link as older versions become obsolete

Stable is pretty conventional as a default. If you check out the numpy and pandas docs we were using as guidance last week they both use stable. So does Xarray. JupyterLab, however, uses latest. I don't know what rationale each of these projects is using to make that decision, they're just the first projects that come to mind.

saberbrasher:
Thanks for the context, @mfisher87! 

To maybe counter the stable pro:
If we are working on a new release, where we know the API will change in some way, why would we publish those specific changes to the docs before that release anyways? Wouldn't it be better to just leave those docs changes as a PR until the actual API has changed, and then publish them? And then, if we do make changes to the docs between releases, we make sure we are only making changes that don't change actual understanding/ functionality of the library... so new how-tos, new content pages, fixing typos, changing layouts, stuff like that. 

If we were working like the above, then using latest as the default should always be fine and not break anything... I think. Right?

mfisher87:
>If we are working on a new release, where we know the API will change in some way, why would we publish those specific changes to the docs before that release anyways? Wouldn't it be better to just leave those docs changes as a PR until the actual API has changed, and then publish them?

Much of our API documentation is auto-generated from code. So as we implement the API change in the code, the "latest" docs will change automatically as the code changes. It's also beneficial for contributors to have access to documentation that reflects the current state of development in the repo, including manually written docs like how-tos that won't be used by users until the next release.

There's a whole other way of managing our docs @jhkennedy has suggested: decoupling our docs into a different repo. But I'm not sure I'm bought in to that as a way forward. That's a big change with its own pros and cons that should probably have a full discussion in another issue, and consensus-building process for making that decision.

mfisher87:
@saberbrasher do you have any changes you'd request to this PR? I'd like to move this PR forward together and continue this discussion in another issue / Zulip! I think it's important to get this merged so the links in our README work again :)

README.md
asteiker, line 1: @mfisher87 Going off of our default version pointing to stable, I'm wondering whether our ReadMe should also point to stable?

Issue #1244, opened 2026-03-04
binder, repo2docker builds failing due to python version and missing source


itcarroll:
We include a simple `environment.yml` for binder and repo2docker users. It has no pinning and only installs the latest version of `earthaccess`. Until repo2docker gains some kind of lockfile support, we'll need to add a minimum Python version to match the minimum version `earthaccess` supports (currently 3.11).

itcarroll:
Well, this is surprising.

1. repo2docker (and thus binder) don't appear to support minimum version pinning. so we need to pin with `python==3.11`
2. repo2docker copies the project to the build location AFTER installing environment.yml, which obviously fails

chuckwondo:
To avoid confusion, we likely want to add the necessary files to a `.binder/` directory rather than at the root of the repo.

itcarroll:
We reference `environment.yml` in our contributing guide, so is the suggestion to leave that alone and create `.binder` folder that works for binder/repo2docker? Makes sense to me, too bad about the clutter.

mfisher87:
> 1. repo2docker (and thus binder) don't appear to support minimum version pinning

Wat? Doesn't it just use mamba under the hood?

chuckwondo:
> We reference `environment.yml` in our contributing guide, so is the suggestion to leave that alone and create `.binder` folder that works for binder/repo2docker? Makes sense to me, too bad about the clutter.

We would not be able to leave `environment.yml` in the root of the repo, as repo2docker would see it there and use that one and ignore the `.binder/` folder. If we want to keep one in the root for contributors to use, we would have to rename it so repo2docker does not recognize it. Frankly, I may be outnumbered, but I suggest eliminating the one at the root of the repo as an option for contributors. To simplify things, there should really be only 1 way to create a dev env for a contributor, but that's a bit tangential.

mfisher87:
> I suggest eliminating the one at the root of the repo as an option for contributors.

:100: :100: :100:

I have also seen pushback to this, but I think this is the right way to go. And, personally, I think we should be pushing `uv` for developers. They will be happier that way (simpler UX, eliminates many pitfalls), and if they have a reason to reject `uv` they probably know how to use `pip` anyway.

itcarroll:
> Wat? Doesn't it just use mamba under the hood?

It does when it comes to updating the environment. But however repo2docker chooses which CondaBuildPack to use ignores the ">=" requirement.

[...]

PR #1245, opened 2026-03-04
add a ".binder" folder to support mybinder.org and repo2docker builds

itcarroll:
## Description

I don't know for how long #1244 has been a problem, but I don't see how the current repository worked with repo2docker and thus with mybinder.org. The docker builds would find the "environment.yml" file, but would not have the source tree and could not use `pip install -e`.

This PR adds a .binder folder that includes
1. a "environment.yml" file to constrain the python version. Unfortunately, a minimum version is not recognized by repo2docker, only an explicit version.
2. a "postBuild" to install the package with pip as advised in jupyterhub/repo2docker#1265

- [x] Open PR as draft

#### Merge checklist

- [x] PR title is descriptive
~~- [ ] If needed, `CHANGELOG.md` updated~~
~~- [ ] If needed, docs and/or `README.md` updated~~
~~- [ ] If needed, unit tests added~~
- [ ] All checks passing (comment `pre-commit.ci autofix` if pre-commit is failing)
- [x] At least one approval

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -2,10 +2,8 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
-  # This environment bootstraps pip, the actual dev environment
-  # is installed and managed with pip
   - python==3.11
-  - pip
+  - pip>=25.1
 variables:
   # Allow pip installs when conda environment is active
   PIP_REQUIRE_VENV: 0
--- a/.binder/postBuild
+++ b/.binder/postBuild
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 set -eux
-python -m pip install --editable ".[dev,test,docs]"
+pip install --group dev --group test --group docs --editable .


itcarroll:
I confirmed that mybinder.org launch is okay.

.binder/environment.yml
mfisher87, line 1: Can we move the `environment.yml` file from the repo root to here?

itcarroll: The same file does not work for local development setup and for binder.

mfisher87: Ideally one file can be used for both development and binder?

itcarroll: Ideally, but that would require a change to how repo2docker works.

mfisher87: >The same file does not work for local development setup and for binder.

>Ideally, but that would require a change to how repo2docker works.

Could you elaborate?

itcarroll: If you look at `repo2docker --debug --no-build .` you'll see that the build copies `src/environment.yml` to the working dir and runs `mamba env update environment.yml`. Then it copies all of `src/` to the working directory. That's why it is suggested to put `pip install -e` in a postBuild script.

mfisher87: Thanks! Apologies in advance if I ask you to re-explain this to me later ;)

To rephrase and make sure I understand: At the time it installs from `environment.yml`, it does not yet have the source code in place that's needed to `pip install --editable .`

.binder/postBuild
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+set -eux
+python -m pip install --editable ".[dev,test,docs]"
```
mfisher87: Do we expect binder users to run tests or build docs or do any sort of development? Perhaps this could be:


```suggestion
python -m pip install .
```

.binder/environment.yml
itcarroll: No worries, I appreciate being checked. Your understanding matches mine.

.binder/postBuild
itcarroll: My assumption was that the existence of the binder badge workflow indicates someone might be using binder for development.

mfisher87: That's surprising to me! I assumed it was just for quickly trying it out without needing setup on the user's end. E.g. I would be uncomfortable using my Git credentials in such an environment, but that's just me.

I think we can safely remove the docs group, though, because building the docs would require an HTTP server to host them and we wouldn't be able to run that in the binder environment.

mfisher87: Either way, we should probably drop the extra syntax and use PEP735 group syntax. Maybe it should be:

```suggestion
python -m pip install --group dev --group test --editable .
```

itcarroll: Can you confirm that **building** the docs requires an HTTP server?

One could use binder to run tests and confirm that docs build without having to mess with your local clone/environment (during work on another PR, for instance). I'm inclined to support all dev cases.

mfisher87: No, building doesn't, but viewing does (in the cloud at least; locally I can just open the file with my browser). My assertion is there's no reason to build unless you can view, except to see if there are build failures, and PRs already do that!

mfisher87: > I'm inclined to support all dev cases.

I think this is fine, no need to overoptimize at this stage :)

mfisher87: But let's still use the `--group` syntax please! It looks like extras syntax works (unexpected), but I don't like that that's overloaded and I think it's confusing :)

itcarroll: Great, agree in principle! I hadn't actually noticed we migrated to "dependency-groups" 👍.

Only now (lolz) I'm getting an error with the `--group` syntax in the ~~"environment.yml"~~ postBuild file.

mfisher87: What error are you getting?

itcarroll:
Note that I tried to switch the root level "environment.yml" to use `--group` as well, and it does not work.

```
Installing pip packages: --group dev, --editable .
Usage: __main__.py [options]

ERROR: Invalid requirement: --group dev
__main__.py: error: no such option: --group
```

That was with pip 26.0.1.

One more review request coming.

.binder/postBuild
itcarroll: Solved. It was the pip version, at least for the "postBuild" usage.

Issue #1250, opened 2026-03-06
Consider a more object-oriented API


mfisher87:
I've been thinking about the current architecture and statefulness of earthaccess, and the sustainability of those choices.
I'll admit I haven't thought deeply about this but I want to put it out there and see how it resonates or doesn't resonate.

## TLDR

Based on our quickstart:

```python
import earthaccess

ea_uat = earthaccess.client(system=earthaccess.UAT)  # <<< New
ea_uat.login(strategy="netrc")
results = ea_uat.search_data(
    short_name='ATL06',
    bounding_box=(-10, 20, 10, 50),
    temporal=("1999-02", "2019-03"),
    count=10,
)
ea_uat.download(results, "./local_folder")
```

Other methods on the client, e.g. `client.system_status()`, `client.search_services()` respect choices made at client initialization.

Or something.

## Benefits

* Testability for us: Less statefulness to think about by eliminating the current global `_auth` interface.
* Testability for users: Can create fixtures that test their code with UAT  or PROD and be more confident about state
* Users can "save" some of their choices in the client, e.g. which system it's using, and be more confident that they haven't "missed a spot" somewhere in the code.
* Ability to make some additional breaking changes at the same time, e.g. an enum for auth stategy? What else? The results class comes to mind but IIRC we intended to make that class iterable to avoid a breaking change.
* Exercising our policies for breaking changes
* Setting expectations for users when we make breaking changes
* Being our first major breaking change, establishing confidence that we can successfully navigate breaking changes as a community

## Costs?

* Instead of "NASA data in 3 lines of code", "NASA data in 4 lines of code"
* Users deal with a breaking change to an interface they know and love
* Labor -- programming the feature, updating our tests, docs, outreach
* Other? Edit me!

These costs are not insignificant. That said, it's sometimes worth the cost for sustainability and maintainability. Is this one of those cases? 🤷 

### Cost mitigation

How much of those costs can be mitigated by changing the architecture and building convenience methods on top which maintain the current API (i.e. user does `earthaccess.login()` and we create a global singleton `EarthaccessClient` instance under the hood)? Is that a worthwhile compromise? What costs will the compromise itself incur? If we do the compromise, should it be temporary and come with `DeprecationWarning`s?

## What am I missing?

Y'all know better than me, please tell me 😁

mfisher87:
[...]

PR #1259, opened 2026-03-12
Set timeout for all requests in set_requests_session

tylanderson:
## Description

Adds timeouts on requests during `set_requests_session`. I had encountered an issue where the first request would run much longer than expected before timing out, this will surface that error faster. Set to 10s for now.

---

- [x] Open PR as draft

#### Merge checklist

- [ ] PR title is descriptive
- [ ] PR body contains links to related and resolved issues (e.g. `closes #1`)
- [ ] If needed, `CHANGELOG.md` updated
- [ ] If needed, unit tests added
- [ ] All checks passing (comment `pre-commit.ci autofix` if pre-commit is failing)
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1259.org.readthedocs.build/en/1259/


mfisher87:
Nice! Thank you! Totally agree a requests session should always have a timeout.

Ideally the user would have some control over this, but the current architecture of earthaccess makes that a bit challenging. A more object oriented approach (client/session interface) for earthaccess (#1250) would make this easier for the user to control. For now, maybe an environment variable would enable that.

I worry that a 10 second timeout might negatively impact folks with less reliable internet connections (e.g. working in remote areas, mobile hotspot, satellite, ??), so I'd suggest we implement some mechanism for the user to override before merging this PR.

Thoughts?

tylanderson:
Yeah agree that configuration currently is a bit difficult as things stand now. I ran a few tests with my network simulating lower performing networks (e.g. 3G) which did work, that being said I think having this be configurable via env is a good option for now, I'll add that in

mfisher87:
Thanks Tyler! :heart:

earthaccess/store.py
```diff
@@ -35,6 +36,27 @@
...
+    Returns:
+        int: The request timeout in seconds.
+    """
+    timeout_str = os.getenv("EARTHACCESS_REQUEST_TIMEOUT")
+    timeout = (
+        int(timeout_str)
+        if timeout_str and timeout_str.isdigit()
+        else DEFAULT_REQUEST_TIMEOUT
+    )
+    return timeout
```
mfisher87, line 58: A bit nitpicky, hope you don't mind!


```suggestion
    timeout_str = os.getenv("EARTHACCESS_REQUEST_TIMEOUT", default="")

    if timeout_str.isdigit():
        return int(timeout_str)
    else:
        return DEFAULT_REQUEST_TIMEOUT_SECONDS
```

earthaccess/store.py
```diff
@@ -35,6 +36,27 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_REQUEST_TIMEOUT = 10  # seconds
```
mfisher87, line 39:
```suggestion
DEFAULT_REQUEST_TIMEOUT_SECONDS = 10
```

What do you think?

tests/unit/test_store.py
```diff
@@ -393,3 +395,87 @@ def test_sibling_tempfile_error(tmp_path):
...
+    )
+
+    auth = Auth()
+    auth.login(strategy="environment")
+
+    return Store(auth)
+
+
+@pytest.mark.parametrize(
+    "env_timeout, first_status, second_status, expected_timeout, expected_calls",
```
mfisher87, line 427:
```suggestion
    ("env_timeout", "first_status", "second_status", "expected_timeout", "expected_calls"),
```

mfisher87 (changes requested):
I think we need a bit more work before this is ready! See comments above.

I've also added some other reviewers who have touched this code. The main question in my mind now is should we set the timeout for every request executed under the session? Why just `set_requests_session`?

Maybe we should modify our [requests.Session subclass](https://github.com/earthaccess-dev/earthaccess/blob/35fa670493ce7324e6e75231e1c1f6994f92ae3c/earthaccess/auth.py#L82) instead. Then, instead of the `try` block we can also make the warning a default behavior.

E.g. (pseudocode)

```
class SessionWithTimeoutAndHeaderRedirection(requests.Session):
    """Requests removes auth headers if the redirect happens outside the
    original req domain.
    """

    def __init__(
        self,
        edl_hostname: str, auth: tuple[str, str] | None = None
    ) -> None:
        super().__init__()
        self.headers.update({"User-Agent": user_agent})

        if auth:
            hook = BasicAuthResponseHook(edl_hostname, auth)
            self.hooks["response"].append(hook)

    @cached_property
    def timeout(self) -> tuple[int, int] | tuple | None:
         # Get it from the envvar / default value

    def request(self, method, url, **kwargs):
        kwargs.setdefault('timeout', self.timeout)
        try: 
            return super().request(method, url, **kwargs)
        except requests.Timeout as e:
            # print warning
            raise e
```

It's been a while since I thought about this, but setting a default timeout on a session was a big point of contention with the Requests community like a decade ago or more. I think the recommended way to do this is actually to override an Adapter class. So I wouldn't consider the above ☝️ to be correct without more research 😁 

Another option we have is to just pass a timeout to every single requests call, and Ruff can help us do that: https://docs.astral.sh/ruff/rules/request-without-timeout/

tylanderson:
> The main question in my mind now is should we set the timeout for every request executed under the session? Why just set_requests_session?

I think that this could be something that I can take a look at and see what kind of lift this would take, although I am unfamiliar with how the sessions within earthaccess are handled and where/if they are shared

Sherwin-14:
Hey @tylanderson ! Are you still interested in working on this PR, or would you mind if someone else picks it up?

Sherwin-14:
@mfisher87 Happy to pick this up if @tylanderson isn't actively working on it this is a small but valuable fix.

On the approach, my instinct is the simpler route: a tiny helper that reads `EARTHACCESS_REQUEST_TIMEOUT` (with a sensible default), then pass timeout = to each request call site. It lets us enable Ruff's S113 so future requests can't slip through without a timeout.

I'm not married to that though  if we'd rather centralize it in the session, I'm happy to go that direction instead. Just wanted to check where you'd like this to land before I start, so I don't go down the wrong path.

chuckwondo:
> @mfisher87 Happy to pick this up if @tylanderson isn't actively working on it this is a small but valuable fix.
> 
> On the approach, my instinct is the simpler route: a tiny helper that reads `EARTHACCESS_REQUEST_TIMEOUT` (with a sensible default), then pass timeout = to each request call site. It lets us enable Ruff's S113 so future requests can't slip through without a timeout.
> 
> I'm not married to that though if we'd rather centralize it in the session, I'm happy to go that direction instead. Just wanted to check where you'd like this to land before I start, so I don't go down the wrong path.

@mfisher87, since you're already engaged in this thread, I'll defer to you, but @Sherwin-14's suggestion seems reasonable to me. It appears you may be on holiday, so we'll wait for you to weigh in.

PR #1265, opened 2026-03-22
Add .env file support for environment login strategy

### LINKED ISSUE #748: Enable getting envvars from `.env` file
Currently we use `os.environ` to access the envvars directly, which doesn't care about a `.env` file. If we use either `dotenv` library or `pydantic-settings` (my personal preference!) we can automatically load envvars from a `.env` file. `pydantic-settings` offers additional encapsulation of valdation and derived values.

Sherwin-14:
This PR allows the `environment` login strategy to load environment variables from `.env` file.  Resolves #748.

----
📚 Documentation preview 📚: https://earthaccess--1265.org.readthedocs.build/en/1265/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Added
 
+- Added .env file support for environment login strategy.
+  ([#748](https://github.com/earthaccess-dev/earthaccess/issues/748)) (@Sherwin-14)
+
 - Added `force` kwarg to `download()` to force redownloads.
 
 ## [v0.16.0] - 2026-01-30
--- a/docs/user/howto/authenticate.md
+++ b/docs/user/howto/authenticate.md
@@ -24,8 +24,8 @@ auth = earthaccess.login(strategy="netrc")
 ```
 
 If your Earthdata Login credentials are set as the environment variables
-`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`, you can explicitly specify their
-use:
+`EARTHDATA_USERNAME` and `EARTHDATA_PASSWORD`, either in a `.env` file
+or your local environment, you can explicitly specify their use:
 
 ```py
 auth = earthaccess.login(strategy="environment")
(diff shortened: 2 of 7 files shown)


jhkennedy:
I feel like I've missed some conversations here, but I'm not sure we should add `load_dotenv` to the earthaccess init.

It looks like this is only being used for earthdata login credentials, and a netrc file is a more standard place to serialize and store the credentials, which we'll also write to disk for users. I'm not sure what benefit supporting `.dotenv` adds and makes reasoning about credentials more complicated?

Similarly, I've typically seen .env use more for testing than for end-users, so I would have expected to see this in a `conftest.py` or similar.

Sherwin-14:
I agree, but to be honest, `.env` is more popular.

I think that in itself is good reason to support it.

`.netrc` isn't as well known as `.env`, and using `.env` for managing
secrets has pretty much been standardized.

`.netrc` is also user-local rather than project local.

Though an archaic usecase, a developer could use different accounts for
different projects if `.env` was added. Which is possible with `.netrc`,
but requires you to fiddle around.

Though, I suppose we could remove load_dotenv and only mention it in the
documentation, that's fine too.

I think we should wait to hear what Matt has to say, since he was the one
who created the issue.

chuckwondo:
I agree with @jhkennedy.

Although dotenv is a small package, our dependency tree is already quite bloated (we should define "extras" to help end-users avoid unwanted deps).

Further, this approach does not allow users to configure things differently than the default (i.e., this approach supports _only_ use of a single file named `.env`).

If a user wants to use a "dot env" file, they should do so on their own so they have full control.

chuckwondo:
> I agree with @jhkennedy.
> 
> Although dotenv is a small package, our dependency tree is already quite bloated (we should define "extras" to help end-users avoid unwanted deps).
> 
> Further, this approach does not allow users to configure things differently than the default (i.e., this approach supports _only_ use of a single file named `.env`).
> 
> If a user wants to use a "dot env" file, they should do so on their own so they have full control.

I suppose this is more of a downvote for the original issue: #748

Sherwin-14:
True.

I guess I can remove calls and the dependency on dotenv(), and only add
lines to the documentation that mention how you can use .env.

Sherwin-14:
Closing this PR as the issue is closed as not planned.

PR #1292, opened 2026-04-13
Enabled TC (type-checking) ruleset and fixed violations

Sherwin-14:
Fixed errors based on TC rulesets, as suggested in the [issue](https://github.com/earthaccess-dev/earthaccess/issues/383) description.

----
📚 Documentation preview 📚: https://earthaccess--1292.org.readthedocs.build/en/1292/


chuckwondo (approved):
Thanks @Sherwin-14

PR #1298, opened 2026-04-14
Return a `Results` class from search and query methods

mfisher87:
## Description

Left to do:

- [ ] Make the `Results` class iterable
- [ ] Remove demo notebooks
- [ ] Convert `TODO` comments to issues, clean up those comments
- [ ] Add a unit test (no actual query) for `to_gdf`

---

- [x] Please review our [Pull Request Guide](https://earthaccess.readthedocs.io/en/latest/contributor/pr-guide/)

#### Merge checklist

- [ ] PR title is descriptive
- [ ] PR body contains links to related and resolved issues (e.g. `closes #1`)
- [ ] If needed, `CHANGELOG.md` updated
- [ ] If needed, unit tests added *(unsure how? see below!)*
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1298.org.readthedocs.build/en/1298/


trey-stafford:
Discussed during earthaccess community call today:

* Subclassing list isn't a great idea. It provides opportunities for mutating the granule/collections in the list without updates to the corresponding query class.
* We agreed that we should create a base `Results` class that `GranuleResults` and `CollectionResults` can subclass. Making two separate containers for the different results types allows for e.g., `isinstance` checks against specific collection types.
* The results should be immutable
* We would like to support lazy loading of results, but this is also in conflict with our desire to provide e.g., `to_gdf`, which would require materializing the results. Materializing the results on one iteration can "wipe out" the results, preventing subsequent iterations. How do we present results to users in a way that they can easily iterate over multiple times lazily? Should we support methods that force materialization?
* We discussed how `to_gdf` should work. Ideally we don't have column names that are the full CMR representation, but instead show readable column names.

jhkennedy:
Sorry I couldn't make it to this breakout discussion today -- got totally engrossed in the plugin discussion.

> We would like to support lazy loading of results, but this is also in conflict with our desire to provide e.g., `to_gdf`, which would require materializing the results. Materializing the results on one iteration can "wipe out" the results, preventing subsequent iterations. How do we present results to users in a way that they can easily iterate over multiple times lazily? Should we support methods that force materialization?

We started this with the intention of moving towards a more object-oriented API (https://github.com/earthaccess-dev/earthaccess/issues/1250 and described in the "what's next" section of `demo_new.ipynb`). So, I don't think the results class should be "lazy" or directly invoke any searches -- that should be the responsibility of the search client. The client can yield results lazily or return the full set of results, but each materialization should be different results objects. That does imply some additional methods (e.g., equality, combining, etc.) that come with the list class.

Note: we included the query parameters and options for provenance, but the expectation is that they'd be used by the client when doing searches. That part, I think, is a bit clunky still; we kicked around a full provenance type object but didn't settle on anything. 

> Subclassing list isn't a great idea. It provides opportunities for mutating the granule/collections in the list without updates to the corresponding query class.

This is addressed by my comments above. The benefit of sub-classing list is that it's the least breaking change for eathaccess (it already just returns a list).

> The results should be immutable

I could be convinced of this, but I'm not sure it's necessary. Generally, I want user's to be able to use the results -- inculding combining multiple searches into one, etc. That all can be done with an imutable object, so I don't think they are contradictory, but I'd need to think more about the tradeoffs between the two.

> We agreed that we should create a base `Results` class that `GranuleResults` and `CollectionResults` can subclass

makes sense to me.

> We discussed how `to_gdf` should work. Ideally we don't have column names that are the full CMR representation, but instead show readable column names.

Yes, human-readable column names would be ideal, but maybe not necessary for a MVP. 

---

@mfisher87 what do you think?

PR #1302, opened 2026-04-16
ci(dependabot): scan composite action in .github/actions/install-pkg

### LINKED ISSUE #1300: Our install-pkg action is not being updated
We use a custom GitHub Actions action to set up `uv`, report the Python version, and install `earthaccess` and we use dependabot to keep our actions up to date. 

However, dependabot is not updating the actions we use inside our custom [install-pkg action](https://github.com/earthaccess-dev/earthaccess/blob/main/.github/actions/install-pkg/action.yml), so the `setup-uv` action is 3 major versions behind currently.

tmchow:
## Summary

Dependabot's `github-actions` ecosystem does not recurse into composite actions, so the `uses:` references inside `.github/actions/install-pkg/action.yml` were never scanned (`setup-uv` is currently 3 major versions behind). Add a second `github-actions` entry scoped to that action's directory, reusing the existing schedule, grouping, and cooldown so updates land together.

## Test plan

- [ ] After merge, verify dependabot opens an update PR for `.github/actions/install-pkg/action.yml` (e.g. bumping `setup-uv` to the current major version).

Fixes #1300
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

----
📚 Documentation preview 📚: https://earthaccess--1302.org.readthedocs.build/en/1302/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,3 +25,16 @@ updates:
           - "*"
     cooldown:
       default-days: 14
+  # Dependabot's github-actions ecosystem does not recurse into composite
+  # actions under .github/actions/, so each custom action needs its own entry
+  # for the `uses:` references inside action.yml to receive updates.
+  - package-ecosystem: "github-actions"
+    directory: "/.github/actions/install-pkg"
+    schedule:
+      interval: "monthly"
+    groups:
+      gha-dependencies:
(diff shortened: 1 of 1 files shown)


chuckwondo (approved):
Great! Thanks @tmchow!

PR #1312, opened 2026-04-23
Enabled G (logging) ruleset and fixed violations

Sherwin-14:
Fixed errors based on other G ruleset. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1312.org.readthedocs.build/en/1312/


earthaccess/api.py
```diff
@@ -180,7 +180,7 @@ def search_datasets(count: int = -1, **kwargs: Any) -> list[DataCollection]:
     else:
         query = DataCollections().parameters(**kwargs)
     datasets_found = query.hits()
-    logger.info(f"Datasets found: {datasets_found}")
+    logger.info("Datasets found: %s", datasets_found)
```
mfisher87, line 183: I originally had mixed feelings about this rule. Here's the justification: https://docs.astral.sh/ruff/rules/logging-f-string/#why-is-this-bad

I think this is fine!

mfisher87 (approved):
Thanks, Sherwin! Looks great :)

PR #1316, opened 2026-04-27
Enabled SIM (flake8-simplify) ruleset and fixed violations

Sherwin-14:
Fixed errors based on  SIM rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1316.org.readthedocs.build/en/1316/


earthaccess/store.py
```diff
@@ -300,7 +291,7 @@ def _running_in_us_west_2(self) -> bool:
         except Exception:
             return False
 
-        if resp.status_code == _HTTP_OK and resp.content == b"us-west-2":
+        if resp.status_code == 200 and resp.content == b"us-west-2":  # noqa: SIM103
             # On AWS, in region us-west-2
             return True
         return False
```
```suggestion
        return resp.status_code == 200 and resp.content == b"us-west-2"
```

tests/integration/test_api.py
```suggestion
    with (
        patch.object(earthaccess.__store__, "_download_file", fail_to_download_file),
        # With "deferred" exceptions, pqdm catches all exceptions, then at the end
        # raises a single generic Exception, passing the sequence of caught exceptions
        # as arguments to the Exception constructor.
        pytest.raises(Exception) as exc_info,
    ):
        earthaccess.download(
            results,
            tmp_path,
            pqdm_kwargs=dict(exception_behaviour="deferred", disable=True),
        )
```

chuckwondo (dismissed):

earthaccess/api.py
```diff
@@ -420,7 +420,7 @@ def download(
     """
     provider = _normalize_location(str(provider))
 
-    if isinstance(granules, DataGranule):
+    if isinstance(granules, DataGranule):  # noqa: SIM114
         granules = [granules]
     elif isinstance(granules, str):
         granules = [granules]
```
mfisher87, line 428: Why not combine these conditions with an `or`?

earthaccess/search.py
```diff
@@ -740,13 +740,9 @@ def _valid_state(self) -> bool:
...
         spatial_keys = ["point", "polygon", "bounding_box", "line"]
         collection_keys = ["short_name", "entry_title", "concept_id"]
-
-        if any(key in self.params for key in spatial_keys):
-            if not any(key in self.params for key in collection_keys):
-                return False
-
-        # all good then
-        return True
+        return not any(key in self.params for key in spatial_keys) or any(
```
mfisher87, line 743: Nitpick: Maybe these conditions should have variables for readability. E.g.:

```python
param_in_spatial_keys = any(key in self.params for key in spatial_keys)
param_in_collection_keys = any(...)

return not param_in_spatial_keys or param_in_collection_keys
```

It's still a bit confusing 😆  I'm sure we can do better with a bit more thought though.

earthaccess/api.py
chuckwondo: It's a type checking issue.

When done separately, as above, `granules` resolves to either `list[DataGranule]` or `list[str]`.

When combined, `granules` resolves to `list[DataGranule | str]`, which is not what we want, nor does it align with the annotation of the function parameter, which type checkers complain about.

We should either add a comment to this effect, or simply do something like so:

```
    if isinstance(granules, (DataGranule, str)):
        granules = cast("list[str]", [granules])
```

It doesn't matter if `granules` is `list[DataGranule]` rather than `list[str]`. We just want to silence the type checker.

chuckwondo: @mfisher87, with Sherwin's latest updates, everything else looks good to me. Do you have a preference for how we should address this bit of code?

chuckwondo:
@Sherwin-14, I see a merge conflict (which is the cause for the pre-commit failure). Please pull main into your branch and address the conflict.

earthaccess/api.py
mfisher87: Ah, didn't even think about that. I'll take another look.

earthaccess/search.py
mfisher87: Let's not worry about this for this PR if you don't want? Just thinking aloud.

tests/integration/test_api.py
```diff
@@ -253,16 +255,32 @@ def test_download_deferred_failure(tmp_path: Path):
...
         # With "deferred" exceptions, pqdm catches all exceptions, then at the end
         # raises a single generic Exception, passing the sequence of caught exceptions
         # as arguments to the Exception constructor.
         with pytest.raises(Exception) as exc_info:
-            earthaccess.download(
-                results,
-                tmp_path,
-                pqdm_kwargs=dict(exception_behaviour="deferred", disable=True),
+            (
+                (
```
mfisher87: Something really weird happened here. The nested parens I can't explain -- did the formatter do this? The extra call to earthaccess.download (one on line 265, one on 279) seems to be an issue with applying a suggestion? Unclear. This definitely needs some cleanup :)

earthaccess/api.py
mfisher87: I really don't like casts, so I think I fall on the side of comment :)

earthaccess/api.py
```diff
@@ -420,7 +420,7 @@ def download(
     """
     provider = _normalize_location(str(provider))
 
-    if isinstance(granules, DataGranule):
+    if isinstance(granules, DataGranule):  # noqa: SIM114
```
mfisher87, line 425:
```suggestion
    # Separate `isinstance` checks enable the typechecker to resolve `granules` as
    # `list[DataGranule] | list[str]` instead of `list[DataGranule | str]`.
    if isinstance(granules, DataGranule):  # noqa: SIM114
```

mfisher87 (changes requested):

earthaccess/search.py
Sherwin-14: Yeah sure, we can fix this later.

tests/integration/test_api.py
mfisher87: This comment was resolved but looks like in the process of fixing the weird formatting stuff it's no longer resolved!

mfisher87 (changes requested):

mfisher87 (approved):
LGTM, thanks Sherwin!

PR #1332, opened 2026-05-06
Enable C4 (flake8-comprehensions) ruleset and fix violations

Sherwin-14:
Fixed errors based on C (flake8-comprehensions) rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.


pyproject.toml
```diff
@@ -214,11 +214,6 @@ ignore = [
   "B023",
   "B905",
   "BLE001",
-  "C402",
-  "C408",
-  "C416",
-  "C420",
-  "C901",
   "COM812",
```
mfisher87, line 218:
```suggestion
  "C901",
  "COM812",
```

C4 and C9 are different rulesets. Let's keep the mccabe complexity ruleset for a future PR, and in that PR actually refactor reduce the mccabe complexity of the alerted function.

If that doesn't sound like something we'll do in the near future, I'm find with `noqa`'ing for now, because that will prevent us from adding new functions that are overly complex!

earthaccess/store.py
```diff
@@ -353,7 +353,7 @@ def get_s3fs_session(
         """
         return self.get_s3_filesystem(daac, concept_id, provider, endpoint)
 
-    def get_s3_filesystem(
+    def get_s3_filesystem(  # noqa: C901
```
mfisher87:
```suggestion
    def get_s3_filesystem(
```

mfisher87 (changes requested):
Looks great, thank you :) Let me know what you think about my comment.

mfisher87 (approved):
Thank you! 🙇

PR #1334, opened 2026-05-07
Fix FileExistsError during direct in-region downloads

### LINKED ISSUE #1331: FileExistsError when downloading from s3

### Have you checked the status of Earthdata services?

- [x] I've executed `earthaccess.status()` and both CMR and EDL returned `'OK'`.

### Current Behavior

When using the `download` function and `earthaccess` detects that it is executing in-region, the following error occurs:

```plain
FileExistsError: [Errno 17] File exists: '/.../partial_...'
```

### Expected Behavior

The error should not occur and all files should be downloaded from s3 successfully.

### Steps To Reproduce

In an AWS runtime in us-west-2, run the following:

```python
import earthaccess

earthaccess.login()

results = earthaccess.search_data(concept_id="C1442068434-GES_DISC", temporal=("2023-06-01", "2023-06-01"), count=10)
# Force "in-region" in case earthaccess fails to auto-detect it.
earthaccess.__store__.in_region = True
earthaccess.download(results, local_path="./tmp", force=True)
```

This will eventually result in the `FileExistsError` noted above.

### Environment

```markdown
- OS: Ubuntu
- Python: 3.12.11
```

### Additional Context

I debugged the issue. It is due to how `s3fs` implements it's `get` method using async operations under the hood for "bulk" operations.  Although we are not using `get` for bulk downloading (we use it to individually download each file), it still does it's async voodoo under the hood, which causes it to try to create the target files multiple times on multiple threads (as far as I can deduce).

We should use `get_file` (specifically used for single files) instead of `get` (for "bulk").  I was able to make this adjustment manually in the environment where I encountered the error above and it resolved the error.

chuckwondo:
## Description

Fixes #1331

- [ ] Place this Pull Request (PR) in draft until it is ready for review (see below)
- [ ] Please review our [Pull Request Guide](https://earthaccess.readthedocs.io/en/latest/contributor/pr-guide/)
- [ ] Mark "ready for review" after following instructions in the guide

#### Merge checklist

- [x] PR title is descriptive
- [ ] If needed, unit tests added *(unsure how? see below!)*
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1334.org.readthedocs.build/en/1334/

### DIFF
--- a/earthaccess/store.py
+++ b/earthaccess/store.py
@@ -748,7 +748,7 @@ class Store:
         wait=wait_exponential(multiplier=1, min=1, max=10),
         retry=retry_if_exception_type(Exception),
     )
-    def download_cloud_file(
+    def _download_cloud_file(
         self,
         s3_fs: fsspec.AbstractFileSystem,
         file: str,
@@ -760,7 +760,7 @@ class Store:
             return file_name  # Skip if already exists and not forcing re-download
 
         with _sibling_tempfile(file_name) as temp_name:
-            s3_fs.get([file], str(temp_name), recursive=False)
+            s3_fs.get_file(file, str(temp_name))
         logger.info("Downloading: %s", file_name)
         return file_name
 
@@ -799,7 +799,7 @@ class Store:
                 s3_fs = self.get_s3_filesystem(provider=provider)
 
             def _download(file: str) -> Path | None:
-                return self.download_cloud_file(s3_fs, file, local_path, force=force)
+                return self._download_cloud_file(s3_fs, file, local_path, force=force)
 
             results = pqdm(data_links, _download, **(pqdm_kwargs or {}))
             return [r for r in results if r is not None]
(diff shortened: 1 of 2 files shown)


earthaccess/store.py
```diff
@@ -760,7 +760,7 @@ def download_cloud_file(
             return file_name  # Skip if already exists and not forcing re-download
 
         with _sibling_tempfile(file_name) as temp_name:
-            s3_fs.get([file], str(temp_name), recursive=False)
+            s3_fs.get_file(file, str(temp_name))
```
betolink, line 763: Nice, would this be the same behavior for non S3 files? I haven't run into this bug (yet) with our out of region access.

betolink (approved):
Looks good, and since it's internal no need to change anything beyond the changelog. Thanks @chuckwondo !

earthaccess/store.py
chuckwondo: No, this is specific to s3fs and how it handles async under the covers in the case of "bulk" operations.  The `get` method is a "bulk" operation, but `get_file` is not, which is specifically for getting a single file, which is how we're using things, so `get` doesn't make sense for our use case anyway.

Issue #383, opened 2023-11-30
Extend ruff formatting?


jhkennedy:
We might want to consider extending our ruff linting to cover:
- [ ]  "UP": [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up) and [pyupgrade settings](https://docs.astral.sh/ruff/settings/#pyupgrade)
- [ ]  "D":   [pydocstyle](https://docs.astral.sh/ruff/rules/#pydocstyle-d) and [pydocstyle settings](https://docs.astral.sh/ruff/settings/#pydocstyle)
- [ ]  "ANN": [flake8-annotations](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) and [flake8-annotations settings](https://docs.astral.sh/ruff/settings/#flake8-annotations)
- [ ]  "PTH": [flake8-use-pathlib](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth)

suggested here: https://github.com/nsidc/earthaccess/pull/372#discussion_r1410264309 by @jhkennedy and supported by @mfisher87

MattF-NSIDC:
PyUpgrade is one of my favorite rulesets, but we'll need to set it up with `target-version` of 3.8 to avoid upgrading our code out of compatibility with our target Pythons :)

Sherwin-14:
@mfisher87 I added the above rulesets to ruff and I now get 1K+ errors. What can be the next steps forward from here?
```
Found 1306 errors.
[*] 253 fixable with the `--fix` option (280 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

mfisher87:
Do you have [prek](https://github.com/j178/prek) (preferred) or pre-commit installed? When you run the ruff job with prek, it should do 253 fixes automatically. **Don't do the unsafe autofixes** :) 

For the remaining 750+ errors, read through them and make some personal decisions. It's unrealistic to manually fix all of those (and unrealistic to review them) all at once. Are there very common errors? Feel free to add those specific rules (by numeric id) to the ignore list. Is there a whole category of rules from the list above that can be disabled for now and implemented later? There's no reason we need to do all of these in one big batch.

The most important thing to me is that we ensure that all the rules that are autofixed remain enabled in the end. With [this PR](https://github.com/earthaccess-dev/earthaccess/pull/1287) you'll be able to see exactly which rules are being autofixed.

Suggested plan (feel free to adapt if you think of a better one):

[...]

PR #1336, opened 2026-05-09
Enable PERF (performance) and COM (flake8-commas) rulesets and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling PERF and COM ruleset. Related to #383.

----
📚 Documentation preview 📚: https://earthaccess--1336.org.readthedocs.build/en/1336/

### DIFF
--- a/earthaccess/results.py
+++ b/earthaccess/results.py
@@ -62,7 +62,11 @@ class CustomDict(dict):
 
     def _filter_related_links(self, link_type: str) -> list[str]:
         """Filter RelatedUrls from the UMM fields on CMR."""
-        return [link["URL"] for link in self["umm"].get("RelatedUrls",[]) if link.get("Type") == link_type]
+        return [
+            link["URL"]
+            for link in self["umm"].get("RelatedUrls", [])
+            if link.get("Type") == link_type
+        ]
 
 
 class DataCollection(CustomDict):
--- a/earthaccess/results.py
+++ b/earthaccess/results.py
@@ -62,12 +62,7 @@ class CustomDict(dict):
 
     def _filter_related_links(self, link_type: str) -> list[str]:
         """Filter RelatedUrls from the UMM fields on CMR."""
-        matched_links: list = []
-        if "RelatedUrls" in self["umm"]:
-            for link in self["umm"]["RelatedUrls"]:
-                if link["Type"] == link_type:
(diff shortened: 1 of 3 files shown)


earthaccess/results.py
```diff
@@ -66,7 +66,7 @@ def _filter_related_links(self, link_type: str) -> list[str]:
         if "RelatedUrls" in self["umm"]:
             for link in self["umm"]["RelatedUrls"]:
                 if link["Type"] == link_type:
-                    matched_links.append(link["URL"])
+                    matched_links.append(link["URL"])  # noqa: PERF401
```
mfisher87: Based on your previous PRs I think you have a good reason for this. Would you mind using comments (GitHub comments not code comments) going forward to tell the reviewer more about why you made certain choices? I imagine you know me Chuck and I, for example, well enough to know where we might have questions :)

Why not list comprehension?

```python
if not "RelatedUrls" in self["umm"]:
    return []

return [link["URL"] for link inself["umm"]["RelatedUrls" if link["Type"] == link_type]
```
?

earthaccess/search.py
```diff
@@ -508,9 +508,7 @@ def parameters(self, **kwargs: Any) -> Self:
             TypeError: The value of a keyword argument is not an argument or tuple
                 of arguments matching the number and type(s) of the method's parameters.
         """
-        methods = {}
-        for name, func in getmembers(self, predicate=ismethod):
-            methods[name] = func
+        methods = {name: func for name, func in getmembers(self, predicate=ismethod)}
```
mfisher87: This is triggering a C416 in CI. Needs to be:


```suggestion
        methods = dict(getmembers(self, predicate=ismethod))
```

earthaccess/results.py
Sherwin-14: It was added because link["URL"] returns a string, which is an iterable, and so replacing append with extend adds the individual characters of the string instead, which causes tests to start failing.

Good suggestion for the list comprehension. I tunnel visioned on just fixing extend :sweat_smile:

Sherwin-14:
Sorry for the mess, was trying to fix the failing pre-commit.

mfisher87:
pre-commit.ci autofix

mfisher87 (approved):

mfisher87:
Thanks Sherwin :)

PR #1344, opened 2026-05-16
Enable PGH003 (blanket-type-ignore) ruleset and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling PGH ruleset. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1344.org.readthedocs.build/en/1344/


tests/unit/test_store.py
```diff
@@ -192,7 +192,7 @@ def test_store_can_create_s3_fsspec_session(self):
 
         store = Store(self.auth)
         self.assertTrue(isinstance(store.auth, Auth))
-        for daac in [
+        for daac in [  # type: ignore[assignment]
```
mfisher87: Why do we need these new type ignores? If we needed these , wouldn't the `test.yml` workflow be failing?

mfisher87 (commented):
Mostly looks good, I especially like the fact that we could identify type ignores to fully remove. But I expected this PR to only include narrowing or removing type ignores, not adding new ones. Why was that necessary?

tests/unit/test_store.py
Sherwin-14: @mfisher87 I added them when I ran mypy for these files, mypy ended up pointing these additional type issues on top of the existing ones. I can remove this if you want?

mfisher87: Since mypy is passing on `main`, can you identify the cause for these new mypy errors? We should definitely make sure this PR passes typechecking, but I also don't want to use ignores to get it passing.

earthaccess/daac.py
```diff
@@ -1,7 +1,20 @@
...
 
+
+class DAACConfig(TypedDict):
+    short_name: str
+    name: str
+    homepage: str
+    cloud_providers: list[str]
+    on_prem_providers: list[str]
+    s3_credentials: str
+    eulas: list[str]
```
mfisher87: I don't understand how this is helping avoid type errors at the moment. I believe the intent was to type the list below with this, which I agree we should do, but that wasn't done and the type errors seem to be resolved anyway!

What do you think is going on? :thinking:

earthaccess/daac.py
```diff
@@ -1,7 +1,20 @@
...
+
+
 DAACS = [
```
mfisher87:
```suggestion
DAACS: list[DAACConfig] = [
```

earthaccess/daac.py
Sherwin-14: Yes I too had a freeze moment here, apparently all errors that I got in tests were fixed because of typing the list out.

mfisher87: Looks like the typechecker is unhappy :( I'm not sure how to solve the hyphen problem.

mfisher87: Ah there's a functional syntax: [typing.python.org/en/latest/spec/typeddict.html#functional-syntax](https://typing.python.org/en/latest/spec/typeddict.html#functional-syntax)

Sherwin-14: UP013 is blocking us from using functional syntax

mfisher87: The [docs for that rule](https://docs.astral.sh/ruff/rules/convert-typed-dict-functional-to-class/) says it should allow functional syntax when identifiers are not valid python identifiers. When you tried this, did you change the underscores to hyphens? If that doesn't work, we can `# noqa` it?

Sherwin-14: Oops I missed it, sorry for keeping this open. I am fixing it asap.

mfisher87: No rush, thanks Sherwin! :heart:

Sherwin-14: @mfisher87 Unit tests are breaking because `aiohttp` 3.14.0 removed aiohttp.streams.AsyncStreamReaderMixin, breaking `vcrpy` . Should I pin `aiohttp` to a version before 3.14 temporarily?

mfisher87: Yeah, but I think we should make that a separate PR!

mfisher87 (approved):
Looks great as usual, thank you @Sherwin-14 !

PR #1345, opened 2026-05-16
Enable PLW (Warning) and FURB (refurb) rulesets and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling PLW ruleset. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1345.org.readthedocs.build/en/1345/


earthaccess/store.py
```diff
@@ -594,7 +594,7 @@ def _open_urls(
             granules[0].startswith("s3") or granules[0].startswith("http")
         ):
             # TODO: method to derive the DAAC from url?
-            provider = provider
+            pass
```
mfisher87: I think instead of `if ...: pass; else: raise` we should invert the conditional (`if not ...: raise`).

mfisher87 (changes requested):
Nice! Small change request :)

earthaccess/store.py
Sherwin-14: Done! But why is the pre-commit failing here?

mfisher87: Unsure, but it seems to all be autofixable

mfisher87:
pre-commit.ci autofix

mfisher87:
Oh, it's trying to remove `noqa`s for the PLR rule that is not supposed to be ignored (#1347), but the ignore came back!

mfisher87:
I think the problem was an incorrect conflict resolution? The PLR ignore came back, so I removed it in [d460cfb](https://github.com/earthaccess-dev/earthaccess/pull/1345/commits/d460cfbabf6f40c09ecd9a9985a47d112941ae5c)

mfisher87 (approved):

PR #1347, opened 2026-05-16
Enable ARG (flake8-unused-arguments)  and PLR (Refactor) rulesets

Sherwin-14:
Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1347.org.readthedocs.build/en/1347/


mfisher87:
I have mixed feelings about moving the rule from an ignore to `noqa` statements.

There's a positive: It prevents us from adding code that violates the rule.

And there's a negative: The violations are more "invisible".

I think the positive outweighs the negative, so let's merge this and think about it more later!

mfisher87 (approved):

PR #1135, opened 2025-11-11
WIP: Icechunk opener

jbusecke:
This is a first start towards building an icechunk opener for earthaccess (see https://github.com/nsidc/earthaccess/discussions/1132 for context).

This PR depends on #1154 

~~This is still very rough and might change a lot~~

This implements a very minimalist test and opener function in `earthaccess.icechunk._open_icechunk_from_url`. There are a ton of todos and questions, but let me try to point out the most pressing ones. 
- [x] Zarr V3: Icechunk will need zarr v3. Curious about @DeanHenze @betolink assesement how difficult it will be to refactor the codebase to support zarr v3. 
- Store location. Since there is no icechunk store (as far as I know) available on any DAAC so far I have prototyped with the [MUR SST virtual icechunk store](https://github.com/developmentseed/mursst-icechunk-updater/tree/main) that I recently built. This made me think if we should loosen the requirement to have the actual store within DAACS/CMR. I think it is perfectly legitimate as a user to want to open a virtual zarr store *somewhere* which points to EDL accessible data, no?
- [ ] Write docs with examples.  

<details><summary>Pull Request (PR) draft checklist - click to expand</summary>

- [ ] Please review our
      [contributing documentation](https://earthaccess.readthedocs.io/en/latest/contributing/)
      before getting started.
- [ ] Populate a descriptive title. For example, instead of "Updated README.md", use a
      title such as "Add testing details to the contributor section of the README".
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [ ] Populate the body of the pull request with:
    - A clear description of the change you are proposing.
    - Links to any issues resolved by this PR with text in the PR description, for
      example `closes #1`. See
      [GitHub docs - Linking a pull request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
- [ ] Update `CHANGELOG.md` with details about your change in a section titled
      `## Unreleased`. If such a section does not exist, please create one. Follow
      [Common Changelog](https://common-changelog.org/) for your additions.
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [ ] Update the documentation and/or the `README.md` with details of changes to the
      earthaccess interface, if any. Consider new environment variables, function names,
      decorators, etc.

Click the "Ready for review" button at the bottom of the "Conversation" tab in GitHub
once these requirements are fulfilled. Don't worry if you see any test failures in
GitHub at this point!

</details>

[...]

PR #1348, opened 2026-05-18
`open_virtual()` 

betolink:
This PR adds `open_virtual()` a method that simplifies how we open virtual stores behind EDL authentication. 

We can open virtual collections listed at the collection level metadata, for now following the convention adopted by JPL using the following entry in `RelatedUrls`

```json
{
  "Description": "Virtual data set reference ",
  "URLContentType": "DistributionURL",
  "URL": "https://archive.podaac.earthdata.nasa.gov/../MUR-JPL-L4-GLOB-v4.1_combined-ref.json",
  "Type": "GET DATA",
  "Subtype": "VIRTUAL COLLECTION"
}
```

> Note: We may have multiple virtual stores for any given collection so we anticipate this convention may change in the near future, for now the assumption is that each collection has at most one virtual representation.

`open_virtual()` has the following kwargs

* `uri`: this can be a collection level result from collection = earthaccess.search_datasets()[0] a local path, or a publicly accessible reference URL. If we use a collection from earthaccess, we don't need to deal with the cumbersome auth scafolding for the underlaying libraries; in all cases we'll get an xarray dataset instance pointing to our virtual dataset.This can be kerchunk or Icechunk stores
* `access`: this one informs earthaccess if the virtual store points to S3 buckets or HTTP, without inspecting the references we need to be explicit for now. values are direct or external (same as indirect)
* `storage_options`: if we need custom authentication for the virtual store this works the same way xarray uses storage options, not needed for NASA data as it's automatic.
* `force_external`: This is a handy flag to force reference translation, this is if the only available store points to S3 but we are accessing the data from outside AWS us-west-2 setting this to True will translate the URLs on the fly for out of region access.
* `load`: if we should load the indexes in our dataset, if we do we can address the dataset as any other xarray dataset and perform operations like subsetting and aggregations, if not we get a virtualizarr dataset backed by ManifestArray instances, this is only useful if we plan to add more virtual references and update the cube.

The method signature is as follows:

```python
ea.open_virtual(
    uri: 'str | Path | earthaccess.DataCollection',
    *,
    access: 'str' = 'indirect',
    storage_options: 'dict[str, Any] | None' = None,
    force_external: 'bool' = False,
    load: 'bool' = True,
    **kwargs: 'Any',
) -> 'xr.Dataset'
```

Another relevant aspect is that `load=False` is slower than `load=True` for now as an artifact of VirtualiZarr not parsing all the coordinates and variables with embedded values(inlined values). The last release of VirtualiZarr supports it but I ran into some issues. For now `load=False` requires a round trip to a local store for the references, this worked for all stores but adds latency for example with a JSON serialized store opening the references can take up to a couple minutes depending on internet speed. `load=True` is fast as we only load coords and lazy load references (chunk manifests). This will have to change for upcoming releases to improve speed. 

Here is a notebook showing some the the capabilities for this new feature: 
https://notebooksharing.space/view/d2eb24605b6fb6b011710d955052f67de77c38728afdbc48ccc9cc33db34ff57#displayOptions=

Another relevant topic is that Icechunk does not support out of region access for stores with authentication, if something like this PR gets mered we will have the whole set of data virtualization available and no further changes are required in earthaccess. https://github.com/earth-mover/icechunk/pull/2143 
 

## Description

---

- [x] Please review our [Pull Request Guide](https://earthaccess.readthedocs.io/en/latest/contributor/pr-guide/)

#### Merge checklist

- [x] PR title is descriptive
- [x] If needed, unit tests added *(unsure how? see below!)*
- [x] All checks passing *(tip: comment `pre-commit.ci autofix` if pre-commit is failing)*
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1348.org.readthedocs.build/en/1348/


earthaccess/virtual/core.py
```diff
@@ -299,3 +301,495 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+    - **Icechunk** — a versioned Zarr store (``.icechunk`` file/URI).
+    - **VirtualiZarr / kerchunk** — reference-file-backed datasets
+      (``.parquet`` or ``.json`` files).
+
+    When given a ``DataCollection``, the virtual store URL is extracted from
+    its metadata (``GET DATA`` + ``VIRTUAL COLLECTION`` subtype).
+
+    Parameters:
+        uri: A ``DataCollection``, or a path/URI to the virtual store
+            (``.icechunk``, ``.parquet``, or ``.json``).
```
danielfromearth, line 735: For the TEMPO month-long icechunk, it's a directory (see [this example notebook for reading the current, experimental TEMPO icechunk](https://github.com/nasa/ASDC_Data_and_User_Services/blob/main/TEMPO/additional_drafts/read-references-for-TEMPO-from-s3-icechunk.ipynb)), not a `.icechunk` file.  Not sure if that's always the case or not for icechunks.

betolink: yeah good catch, this was just a convention but can be changed, I think another way of making sure it's an icechunk store is the protocol prefix, `icechunk://` that the icechunk store seems to know how to handle. @maxrjones, do you know the canonical way for this?

danielfromearth: I added a `DistributionUrl` to the metadata record for the TEMPO collection experimented with in that notebook (referenced in my last comment). The current code raises an error for that icechunk in S3 because it is a directory, without `.icechunk` suffix:

```python
import earthaccess as ea
ea.login()

collection = ea.search_datasets(short_name="TEMPO_NO2_L3", version="V04")

vds = ea.open_virtual(collection[0])  # raises the error
```

error:
```traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[15], line 11
      7     short_name="TEMPO_NO2_L3",
      8     version="V04",
      9 )
     10 
---> 11 vds = ea.open_virtual(collection[0])
     12 vds

File [/srv/conda/envs/notebook/lib/python3.13/site-packages/earthaccess/virtual/core.py:757](https://openscapes.2i2c.cloud/srv/conda/envs/notebook/lib/python3.13/site-packages/earthaccess/virtual/core.py#line=756), in open_virtual(uri, access, storage_options, force_external, load, **kwargs)
    752     if not _is_icechunk_uri(url) and not _is_kerchunk_uri(url):
    753         msg = (
    754             f"Unrecognised virtual store URL in collection: {url}. "
    755             "Expected a .icechunk, .parquet, or .json file."
    756         )
--> 757         raise ValueError(
    758             msg,
    759         )
    760 else:
    761     url = str(uri)

ValueError: Unrecognised virtual store URL in collection: s3://asdc-prod-public/virtual-reference-docs/TEMPO_L3_V04_icechunk_for_202601. Expected a .icechunk, .parquet, or .json file.
```

danielfromearth (changes requested):

earthaccess/virtual/core.py
betolink: Yeah we need to change how we validate the VDS URL. Should we come up with a convention? e.g. if any in the URL we match icechunk... and we are opening a VDS we assume icechunk. This is to avoid users having to guess the store.

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
         storage_options=storage_options,
     )
+
+
+# ---------------------------------------------------------------------------
+# open_virtual — open existing virtual stores (Icechunk / VirtualiZarr)
+# ---------------------------------------------------------------------------
+
+
+def _is_icechunk_uri(uri: str) -> bool:
```
betolink, line 311: This needs to be refactored, should we just match icechunk in the URI? should we use ic+http protocol notation?

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+        )
+    elif storage_options:
+        storage = icechunk.http_storage(uri, storage_options)
+    else:
+        storage = icechunk.local_filesystem_storage(uri)
+
+    repo = icechunk.Repository.open(storage=storage)
+    session = repo.readonly_session("main")
+    store = session.store
+    return xr.open_zarr(store, **kwargs)
```
betolink, line 356: The reason to return xarray ds instead of the store is to have a unified API with kerchunk. Also what else can we do with a store if it's not opening it with xarray?

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+    store = zarr.storage.FsspecStore(fs, read_only=True)
+    return xr.open_zarr(store, consolidated=False, **kwargs)
+
+
+def _open_icechunk_from_collection(
+    collection: earthaccess.DataCollection,
+    url: str,
+    access: str = "indirect",
+    **kwargs: Any,
+) -> xr.Dataset:
```
betolink, line 482: This superseeds the PR (draft) Julius started working #1135

earthaccess/virtual/core.py
maxrjones: This is the leading proposal for a canonical URL representation of Zarr/Icechunk: https://github.com/jbms/url-pipeline.

It's still a draft, but it's where the ecosystem is converging. It's implemented in TensorStore and Neuroglancer, with a zarr-python PR in progress ([#3369](https://github.com/zarr-developers/zarr-python/pull/3369)). Earthmover's Zarrs.jl already uses it. Its [icechunk: scheme](https://github.com/jbms/url-pipeline/blob/main/schemes/icechunk.md) directly covers our case. `<url>|icechunk:` addresses a repo root (no node), and the spec defines directory-based format auto-detection. I'll check with Ian about the status of the zarr-python PR. I'd avoid inventing yet another URL scheme (e.g. ic+http) when this exists.

betolink:
After testing more, seems like we could go ahead with the current PR and update to more native use of virtualizarr once some of the following issues get addressed. Virtualizarr only supports fsspec for opening a kerchunk store, this is kerchunk to ManifestStore, (Kerchunk parsers) and we still have issues with decoding FillValue. 

One thing we need to define is how CMR will encode the store type, Max suggested following the convention here: https://github.com/jbms/url-pipeline/blob/main/schemes/icechunk.md looks good, although my personal preference would be to use the protocol prefix like git e.g. {store_type}+{net_protocol}:// so one could have `icechunk+https://some_url` and even with branches git already handles it with @branch. e.g. `icechunk+s3://bucket/store@branch`  I'd say there is a parity between git and icechunk in terms of branches nodes etc. Unless I'm missing something. Regardless, we should (and will) follow what the zarr/icechunk community decides and not use a custom schema.


What do you guys think? @maxrjones @danielfromearth

owenlittlejohns:
Sorry - super late comments here:

```
{
  "Description": "Virtual data set reference ",
  "URLContentType": "DistributionURL",
  "URL": "https://archive.podaac.earthdata.nasa.gov/../MUR-JPL-L4-GLOB-v4.1_combined-ref.json",
  "Type": "GET DATA",
  "Subtype": "VIRTUAL COLLECTION"
}
```

I was chatting with the CMR folks late last week to try and work out if this is a suitable related URL type. Apparently "VIRTUAL COLLECTION" has a different intent (grouping together data from multiple collections), and they are keen for us to adopt a different sub type. I've been tasked with getting the right vocabulary from this crowd. I was planning to use @abarciauskas-bgse's next fortnightly co-working session to get the right folks together to decide on the best terminology.

One other thought - I wonder if we should actually move away from stating "VIRTUAL", too? Isn't the access pattern the same if we're pointing at a virtual store, or the entry point of an actual Zarr store? (Probably getting into the conversation we could have at Aimee's session)

Once we have the type and sub type determined, the CMR folks can implement it pretty quickly.

pyproject.toml
```diff
@@ -51,6 +51,11 @@ dependencies = [
   "multimethod >=1.8",
   "importlib-resources >=6.3.2",
   "typing_extensions >=4.10.0",
+  "dask-labextension>=7.0.0",
+  "dask>=2025.7.0",
+  "distributed>=2025.7.0",
+  "holoviews>=1.22.1",
+  "panel>=1.8.10",
```
maxrjones, line 58: these are some heavy dependencies. I recommend moving them to optional dependency groups.

pyproject.toml
```diff
@@ -62,46 +67,35 @@ Changelog = "https://github.com/earthaccess-dev/earthaccess/blob/main/CHANGELOG.
...
   "kerchunk>=0.2.9",
   "dask",
   "h5py >=3.14.0",
+  "h5netcdf",
   "obstore >= 0.8.0",
   "fastparquet >= 2023.4.0",
   "numcodecs >= 0.16.2",
   "xarray",
+  "matplotlib",
+  "hvplot",
```
maxrjones, line 86: virtualizarr capabilities don't require hvplot or matplotlib. I suggest these should be in a separate optional dependency group (e.g., `viz`, following xarray's naming).

earthaccess/virtual/_parser.py
```diff
@@ -133,3 +135,75 @@ def get_urls_for_parser(
...
+                attrs=var.attrs,
+                name=name,
+            )
+            if isinstance(var.data, ManifestArray)
+            else var
+        )
+        for name, var in ds.data_vars.items()
+    }
+
+    return xr.Dataset(new_vars, coords=ds.coords, attrs=ds.attrs)
```
maxrjones, line 209: this seems indicative of an upstream problem. what motivated this patch? what issues were you running into?

earthaccess/virtual/core.py
maxrjones: [url-pipeline](https://github.com/jbms/url-pipeline) is the way to go eventually for both kerchunk and icechunk identification, but I also understand wanting to get this out and think this could be a TODO for later.

FWIW the definitive way to identify an icechunk store is via the [magic bytes](https://icechunk.io/en/stable/reference/spec-v2/#binary-file-format) in the [root file](https://icechunk.io/en/stable/reference/spec-v2/#file-layout) IIUC, but the url-pipeline avoids needing to send an extra request.

earthaccess/virtual/core.py
maxrjones: you could use Icechunk stores with the Zarr API instead of the Xarray API

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+            Ignored when ``uri`` is a ``DataCollection``.
+        force_external: When ``True``, download kerchunk reference files and
+            rewrite ``s3://`` URLs to ``https://``, so the dataset can be
+            opened without direct S3 access.  Only applies to ``.json`` and
+            ``.parquet`` reference files.  Requires authentication.
+        load: When ``True`` (default), returns a concrete lazily-loaded dataset
+            via the kerchunk engine.  When ``False``, returns a virtual dataset
+            backed by ``ManifestArray`` objects via VirtualiZarr's
+            ``open_virtual_dataset``.
+        **kwargs: Additional keyword arguments forwarded to the opener.
```
maxrjones, line 747: since the opener isn't public, it's hard for users to now what kwargs are accepted. I personally prefer smaller, well-typed functions (in this case a separate open_kerchunk and open_icechunk function that could have kwargs aligned with the 3rd party implementation) but recognize that others prefer fewer more feature complete functions. Having separate `open_icechunk` and `open_kerchunk` functions would also resolve the need for url-pipeline resolution or other url based heuristics.

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+
+
+def _sanitize_references_for_external(url: str) -> str:
+    """Download a kerchunk reference file, rewrite ``s3://`` URLs to ``https://``.
+
+    The HTTPS base is inferred from the reference file URL's host, which works
+    for all NASA Earthdata Cloud DAACs (PODAAC, NSIDC, GES DISC, LP DAAC, …).
+
+    Returns the local path to the sanitized file.
+    """
```
maxrjones, line 405: it should be possible to use virtualizarr for this, but that could be a simple TODO to resolve later

earthaccess/virtual/core.py
```diff
@@ -299,3 +301,492 @@ def _load_via_kerchunk(  # noqa: PLR0913
...
+        )
+        raise ValueError(
+            msg,
+        )
+
+    vds = vz.open_virtual_dataset(ref_path, parser=parser, registry=registry, **kwargs)
+
+    for k in kds.coords:
+        vds.coords[k] = kds[k]
+    return vds
```
maxrjones, line 685: I don't understand why you need L667-685 if you already have an xarray dataset opened via L665.

maxrjones (commented):
awesome work, @betolink!

It'd be worth trying to load a small snippet of the virtual Icechunk array in the notebook. I'm surprised you don't need any `authorize_virtual_chunk_access=...` logic for Icechunk stores. It's possible the demo just doesn't get to the point where virtual chunk access authentication is needed. https://www.earthmover.io/blog/secure-virtual-chunks provides helpful context for this configuration setting.

One conceptual correction about the virtual-diagram is that virtualizarr does not require any JSON for interchange. The only JSON usage is if Kerchunk is used as a starting point or output format. Otherwise, the byte ranges and uris are all handled as numpy arrays (passed directly to Rust when using Icechunk), which is more efficient than using dicts + JSON.

earthaccess/virtual/_parser.py
betolink: In some datasets like PODAAC SST, there is a change in the compression level for the codecs, Virtualizarr checks for homogeneity in the codecs along the concatenating dimension so this would fail that. However, in practice level 6 and level 7 do not alter the hashes is basically the same compression for all values in the dataset. This is an experimental feature to pass VZ checks.

earthaccess/virtual/core.py
betolink: this is a key question for this PR, I'm trying to avoid splitting this into `open_icechunk()` and `open_kerchunk()` and follow a bit what xarray does with the engines. Users won't need to be concerned about it and if CMR follows the url pipeline convention or some simpler convention to start with, we should be fine. We can also expose those at the top level and let power users decide.

pyproject.toml
betolink: Good catch, they should be optional.

earthaccess/virtual/core.py
betolink: I came around to think that `open_icechunk` and `open_kerchunk` are better off as their own top level methods. I think all of them should return an xarray dataset object, when you mentioned "you could use Icechunk stores with the Zarr API instead of the Xarray API" is just the way we open it? or the instance it gets returned? @maxrjones

asteiker:
PR #1441 updates the Readme to include related Python packages. We should include a mention of `earthaccess.open_virtual()` within this PR as part of the Virtualizarr summary.

PR #1350, opened 2026-05-22
Enable BLE (flake8-blind-except) ruleset and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling BLE ruleset. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1350.org.readthedocs.build/en/1350/

### DIFF
--- a/tests/unit/test_collection_queries.py
+++ b/tests/unit/test_collection_queries.py
@@ -77,8 +77,6 @@ def test_query_can_parse_single_dates(start, end, expected):
 @pytest.mark.parametrize("start,end,expected", invalid_single_dates)
 def test_query_can_handle_invalid_dates(start, end, expected):
     query = DataCollections()
-    try:
-        query = query.temporal(start, end)
-    except Exception as e:
-        assert isinstance(e, ValueError)
-        assert "temporal" not in query.params
+    assert "temporal" not in query.params
+    with pytest.raises(ValueError):
+        query.temporal(start, end)
--- a/tests/unit/test_granule_queries.py
+++ b/tests/unit/test_granule_queries.py
@@ -48,12 +48,9 @@ def test_query_can_parse_single_dates(start, end, expected):
 @pytest.mark.parametrize("start,end,expected", invalid_single_dates)
 def test_query_can_handle_invalid_dates(start, end, expected):
     granules = DataGranules().short_name("MODIS")
-    try:
-        granules = granules.temporal(start, end)
-    except Exception as e:
-        assert isinstance(e, ValueError)
-        assert "temporal" not in granules.params
-
+    assert "temporal" not in granules.params
+    with pytest.raises(ValueError):
+        granules.temporal(start, end)
 
 @pytest.mark.parametrize("bbox,expected", bbox_queries)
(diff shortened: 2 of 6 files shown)


tests/unit/test_granule_queries.py
```diff
@@ -48,11 +48,9 @@ def test_query_can_parse_single_dates(start, end, expected):
...
 def test_query_can_handle_invalid_dates(start, end, expected):  # noqa: ARG001
     granules = DataGranules().short_name("MODIS")
-    try:
-        granules = granules.temporal(start, end)
-    except Exception as e:
-        assert isinstance(e, ValueError)
-        assert "temporal" not in granules.params
+    with pytest.raises(ValueError):
+        granules.temporal(start, end)
+    assert "temporal" not in granules.params
```
mfisher87:
```suggestion
    assert "temporal" not in granules.params
    with pytest.raises(ValueError):
        granules.temporal(start, end)
```

It's not immediately obvious whether this assertion is reachable without uncommon knowledge about how `pytest.raises` works.

tests/unit/test_collection_queries.py
```diff
@@ -77,8 +77,6 @@ def test_query_can_parse_single_dates(start, end, expected):
...
 def test_query_can_handle_invalid_dates(start, end, expected):  # noqa: ARG001
     query = DataCollections()
-    try:
-        query = query.temporal(start, end)
-    except Exception as e:
-        assert isinstance(e, ValueError)
-        assert "temporal" not in query.params
+    with pytest.raises(ValueError):
+        query.temporal(start, end)
+    assert "temporal" not in query.params
```
mfisher87:
```suggestion
    assert "temporal" not in query.params
    with pytest.raises(ValueError):
        query.temporal(start, end)

```

mfisher87 (commented):
I'm not sure how I feel about listing out the possible exception classes over `noqa`'ing. As much as you know I don't like `noqa`'ing 🙃 do we have high confidence that we've listed out the correct exception classes?

Sherwin-14:
> I'm not sure how I feel about listing out the possible exception classes over `noqa`'ing. As much as you know I don't like `noqa`'ing 🙃 do we have high confidence that we've listed out the correct exception classes?

I tried to understand the functions and come up with possible exceptions that could happen to the best of my understanding. I think we should get some more eyes into this that way we can get an exhaustive list of exception classes if at all I missed something in the very first place.

tests/unit/test_granule_queries.py
Sherwin-14: @mfisher87 Oops, I missed something important here :(

mfisher87:
@Sherwin-14 Hmm... I'm thinking let's actually just `noqa` those. For these PRs I'm leaning more towards let's keep it simple, maximize the protection of our codebase from future changes, and in later PRs we can cut down the `noqa`s. Let's open an issue :)

mfisher87 (approved):
LGTM pending tests!

mfisher87:
Thanks, Sherwin!!! :tada:

PR #1362, opened 2026-06-06
Enable B (flake8-bugbear) ruleset and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling B rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1362.org.readthedocs.build/en/1362/


earthaccess/store.py
```diff
@@ -419,7 +419,7 @@ def get_s3_filesystem(  # noqa: C901
             token=creds["sessionToken"],
         )
 
-    @lru_cache
+    @lru_cache  # noqa:B019
```
mfisher87, line 422: Maybe we should actually fix this one. Is it important for this function to return a unique httpfilesystem per instance of `Store`? I'm not sure at the moment. @betolink

earthaccess/auth.py
```diff
@@ -67,7 +67,7 @@ def __call__(self, r: requests.Response, **kwargs: Any) -> requests.Response:
 
         # Consume content and release the original connection to allow our new
         # request to reuse the same one.
-        r.content
+        r.content  # noqa:B018
```
mfisher87: I think we can resolve this one. What about (if I'm reading the docs right):

```python
r.raw.drain_conn()
```

This is a little more self-explanatory, plus it's faster and safer (it doesn't decode the response).

This isn't as self-explantory, and it's less performant and retains the risk of decoding errors, but it's a smaller change:

```python
_ = r.content
```

I'd choose the former personally!

earthaccess/store.py
mfisher87: Also, unrelated to this PR, but either the docstring or the return type annotation are wrong. Does this return `AbstractFileSystem` or `HTTPFileSystem` (I mean, technically both, but the "correcter" annotation should be the most specific, `HTTPFileSystem`, right?)

We shouldn't keep type information in the docstrings because it leads to this kind of sync problem. Autodoc will read the types from the function signature.

tests/unit/test_store.py
```diff
@@ -272,7 +272,12 @@ def test_session_reuses_token_download(self):
                 # Track cloned sessions
                 cloned_sessions = set()
 
-                def mock_clone_session_in_local_thread(original_session):
+                def mock_clone_session_in_local_thread(
+                    original_session,
+                    store=store,
+                    edl_hostname=edl_hostname,
+                    cloned_sessions=cloned_sessions,
+                ):
```
mfisher87, line 280: I'm curious what bugbear error code does this resolve?

mfisher87 (changes requested):

tests/unit/test_store.py
Sherwin-14: B023 is the code that is resolved

mfisher87: 👍

mfisher87 (approved):
Looks good to me!

PR #1371, opened 2026-06-27
Enable ANN (flake8-annotations) ruleset and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling ANN rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1371.org.readthedocs.build/en/1371/


Sherwin-14:
No clue how to fix the unit test, any ideas?

earthaccess/virtual/_parser.py
```diff
@@ -43,7 +43,7 @@
 def resolve_parser(
     parser: ParserType,
     group: str | None = None,
-) -> Any:
+) -> Any:  # noqa: ANN401
```
mfisher87, line 46: I'd like to fix this one in this PR. I think we can change the `ParserType` type in `earthaccess/virtual/_types.py` to be `ParserName` (and remove `| Any` there as well), and create a new `Parser` type which is `HDFParser | NetCDF3Parser | KerchunkJSONParser | KerchunkParquetParser`. Then this function can accept `parser: ParserName | Parser` and return `Parser`. These two new types would need to be kept in sync because Python has no way to dynamically generate string literal types from object names. But it would be better than using `Any`.

We could then type `_parser_map` as `dict[ParserName, Callable[[], Parser]]`.

earthaccess/virtual/_parser.py
```diff
@@ -106,7 +106,7 @@ def resolve_parser(
 
 def get_urls_for_parser(
     granules: list[earthaccess.DataGranule],
-    parser: Any,
+    parser: Any,  # noqa: ANN401
```
mfisher87, line 109: We could then also use the `Parser` type here.

earthaccess/virtual/core.py
```diff
@@ -210,8 +210,8 @@ def virtualize(  # noqa: PLR0913
 
 def _open_virtual_mfdataset(  # noqa: PLR0913
     granules: list[earthaccess.DataGranule],
-    parser: Any,
-    registry: Any,
+    parser: Any,  # noqa: ANN401
+    registry: Any,  # noqa: ANN401
```
mfisher87: Based on virtualizarr docs, this should be `ObjectStoreRegistry` (https://virtualizarr.readthedocs.io/en/stable/api/virtualizarr.html)

earthaccess/__init__.py
```diff
@@ -76,7 +76,7 @@
 _lock = threading.Lock()
 
 
-def __getattr__(name):  # type: ignore
+def __getattr__(name: str):  # noqa: ANN202 # type: ignore
```
mfisher87: We know the return type here -- `Auth | Store | None` if I'm reading correctly.

earthaccess/auth.py
```diff
@@ -113,7 +113,7 @@ def login(
         strategy: str = "netrc",
         persist: bool = False,  # noqa: FBT001, FBT002
         system: System | None = None,
-    ) -> Any:
+    ) -> Any:  # noqa: ANN401
```
mfisher87: This should be `Auth` right? Based on the docstring.

earthaccess/formatters.py
```diff
@@ -18,7 +18,7 @@ def _repr_collection_html() -> str:
     return "<div></div>"
 
 
-def _repr_granule_html(granule: Any) -> str:
+def _repr_granule_html(granule: Any) -> str:  # noqa: ANN401
```
mfisher87: We know that a granule will always have a specific type, right?

earthaccess/search.py
```diff
@@ -749,7 +749,7 @@ def _valid_state(self) -> bool:
             key in self.params for key in collection_keys
         )
 
-    def _is_cloud_hosted(self, granule: Any) -> bool:
+    def _is_cloud_hosted(self, granule: Any) -> bool:  # noqa: ANN401
```
mfisher87: Same here, do we have a `Granule` type we can use? I don't remember. Or is this just raw CMR JSON (as a dict) at this point? In that case we could at least say it's a `dict`.

earthaccess/store.py
```diff
@@ -94,7 +94,7 @@ def __getattribute__(self, name: str) -> Any:
         proxy = super().__getattribute__("f")
         return getattr(proxy, name)
 
-    def __reduce_ex__(self, protocol: Any) -> Any:
+    def __reduce_ex__(self, protocol: Any) -> Any:  # noqa: ANN401
```
mfisher87: This looks like it's a tuple of a callable and a tuple of objects of known types. We can type this!

earthaccess/store.py
```diff
@@ -163,10 +163,10 @@ def multi_thread_open(data: tuple[str, DataGranule | None]) -> EarthAccessFile:
 
 
 def make_instance(
-    cls: Any,
+    cls: Any,  # noqa: ANN401
     granule: DataGranule,
     auth: Auth,
-    data: Any,
+    data: Any,  # noqa: ANN401
```
mfisher87: Is this a file path? As a `Path` or `str`?

earthaccess/store.py
```diff
@@ -163,10 +163,10 @@ def multi_thread_open(data: tuple[str, DataGranule | None]) -> EarthAccessFile:
 
 
 def make_instance(
-    cls: Any,
+    cls: Any,  # noqa: ANN401
```
mfisher87: I think there is a finite list of classes this could be right?

earthaccess/store.py
```diff
@@ -236,7 +236,7 @@ def _sibling_tempfile(sibling: Path) -> Generator[Path, None, None]:
 class Store:
     """Store class to access granules on-prem or in the cloud."""
 
-    def __init__(self, auth: Any, pre_authorize: bool = False) -> None:  # noqa: FBT001, FBT002
+    def __init__(self, auth: Any, pre_authorize: bool = False) -> None:  # noqa: FBT001, FBT002, ANN401
```
mfisher87: `auth` is an `Auth`, right?

pyproject.toml
```diff
@@ -238,9 +231,10 @@ combine-as-imports = true
 
 [tool.ruff.lint.per-file-ignores]
 # Allow `print` and `pprint` in notebooks
-"*.ipynb" = ["T201", "T203", "N803", "RET504", "ERA001", "PLR2004", "PD002", "PD010", "PD011", "EM", "ARG001", "BLE001", "FBT"]
-"tests/**/*.py" = ["PLR2004", "FBT", "INP001"]
+"*.ipynb" = ["T201", "T203", "N803", "RET504", "ERA001", "PLR2004", "PD002", "PD010", "PD011", "EM", "ARG001", "BLE001", "FBT", "ANN001", "ANN002","ANN003", "ANN201", "ANN202", "ANN204","ANN401",]
+"tests/**/*.py" = ["PLR2004", "FBT", "INP001",   "ANN001","ANN002","ANN003", "ANN201", "ANN202","ANN204","ANN401"]
 "scripts/**/*.py" = ["INP001"]
+"stubs/**/*.pyi" = ["ANN401"]
```
mfisher87, line 234: What kind of errors are we seeing with stubs? Our stubs should be well-typed.

mfisher87 (changes requested):
I think we need to do more in this PR to actually define some of these types instead of using ignores. Many of them appear to be low-hanging fruit at first glance, but I could be wrong. What do you think?

earthaccess/virtual/_parser.py
Sherwin-14:
```py

def test_resolve_parser_invalid_string_raises() -> None:
    """resolve_parser raises ValueError and lists valid names in the message."""
    with pytest.raises(ValueError, match="DMRPPParser"):
        resolve_parser("UnknownParser")

```
This one starts failing, I can delete this yeah?

mfisher87: I need more detail -- is there a type checker error? Or the test fails? What is the error?

I don't think we should delete the test, I'd have to be convinced :)

Sherwin-14: Sorry was in a bit of a hurry that day..
and yeah, it's a type checker error.
```py
tests/unit/test_virtual.py:196: error: Argument 1 to "resolve_parser" has incompatible type "Literal['UnknownParser']";
expected
"Literal['DMRPPParser', 'HDFParser', 'NetCDF3Parser', 'KerchunkJSONParser', 'KerchunkParquetParser'] | HDFParser | NetCDF3Parser | KerchunkJSONParser | KerchunkParquetParser | DMRPPParser"
 [arg-type]
            resolve_parser("UnknownParser")
```

earthaccess/__init__.py
Sherwin-14: Yeah, but we end up with a cascade of errors if we do that.
```py
~
earthaccess/api.py:530: error: Item Auth? of Auth? | Store? | None? has no attribute "authenticated"  [union-attr]
        if earthaccess.__auth__.authenticated:
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Only solution seems to be to add `if not isinstance(..., None)`  where this error shows up.

earthaccess/virtual/_parser.py
mfisher87: I'd say we can type ignore that line. We want to test runtime behavior when the defined input type is violated, and the type checker is getting in the way of that. End-users may not be using a type checker so maybe they could get in this situation in real world usage.

earthaccess/__init__.py
mfisher87: I read a bit more carefully. Looks like the correct solution is to use [typing.overload](https://docs.python.org/3/library/typing.html#typing.overload). If `name` is `"__auth__"` then we return an `Auth`. 

```python
@overload
def __getattr__(name: Literal["__auth__"]) -> Auth:
    ...

@overload
def __getattr__(name: Literal["__store__"]) -> Store:
    ...

def __getattr__(name: Literal["__auth__", "__store__"]) -> Auth | Store:  # Or this could be untyped? I don't know what's conventional, but I normally include the type information. The docs show the implementation function as untyped so maybe better to do that.
    # Actual implementation here
```

(I think we need to do #1004 and eliminate this singleton pattern before 1.0)

Sherwin-14: Tried the suggestions above but `mypy` special-cases module-level `__getattr__` (PEP 562) as one fixed fallback signature, no per-overload dispatch unlike class methods. What should we do in this case? These are the errors I got when I ran tests post applying this suggested change.

```
earthaccess/__init__.py:79: error: Invalid signature "Callable[[Literal['__auth__']], Auth]" for "__getattr__"  [misc]
    def __getattr__(name: Literal["__auth__"]) -> Auth: ...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
earthaccess/__init__.py:81: error: Invalid signature "Callable[[Literal['__store__']], Store]" for "__getattr__"  [misc]
    def __getattr__(name: Literal["__store__"]) -> Store: ...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
earthaccess/__init__.py:92: error: Incompatible return value type (got "Auth | Store | None", expected "Auth | Store") 
[return-value]
        return _auth if name == "__auth__" else _store
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/integration/test_api.py:241: error: Module has no attribute "__store__"  [attr-defined]
            patch.object(earthaccess.__store__, "_download_file", fail_to_download_file),
                         ^~~~~~~~~~~~~~~~~~~~~
tests/integration/test_api.py:260: error: Module has no attribute "__store__"  [attr-defined]
            patch.object(earthaccess.__store__, "_download_file", fail_to_download_file),

```

chuckwondo:
Hi @Sherwin-14. It looks like you and @mfisher87 have been working through this, so I won't muddy the waters with additional comments.  I'm just pinging you to see if you need some assistance working through this.  Alternatively, if this is becoming too complicated to sort through, we could put this on the back burner for now, if there are other Ruff rules remaining for you to work through.

PR #1372, opened 2026-06-29
Enable SLF001 (private-member-access) and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling SLF rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1372.org.readthedocs.build/en/1372/

### DIFF
--- a/earthaccess/auth.py
+++ b/earthaccess/auth.py
@@ -71,7 +71,7 @@ class BasicAuthResponseHook:
         r.close()
 
         prepared_request = r.request.copy()
-        cookies: CookieJar = prepared_request._cookies  # type: ignore
+        cookies: CookieJar = prepared_request._cookies  # noqa: SLF001 # type: ignore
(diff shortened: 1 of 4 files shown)


earthaccess/auth.py
```diff
@@ -71,7 +71,7 @@ def __call__(self, r: requests.Response, **kwargs: Any) -> requests.Response:
         r.close()
 
         prepared_request = r.request.copy()
-        cookies: CookieJar = prepared_request._cookies  # type: ignore
+        cookies: CookieJar = prepared_request._cookies  # noqa: SLF001 # type: ignore
```
mfisher87: This seems to have disabled the type-ignore. The type-ignore should also be scoped to `attr-defined`.

mfisher87 (changes requested):

mfisher87 (approved):
LGTM! Thanks Sherwin :)

PR #1373, opened 2026-06-30
fix(results): derive each S3 link from its own HTTPS link

arpitjain099:
## Description

When a cloud-hosted granule only exposes HTTPS links and we fall back to guessing its S3 paths, `data_links(in_region=True)` comes back with the first file repeated and the others missing. The cause is in `DataGranule._derive_s3_link`: the loop walks each HTTPS link, but it builds every S3 URL from `links[0]` instead of the current link, so a granule with several files loses all but the first. I switched that line to use the loop variable so each HTTPS link maps to its own S3 URL.

I added a unit test in `tests/unit/test_results.py` that builds a granule with two protected HTTPS links and checks they map to two distinct S3 URLs. It fails on the old code (the second link comes back as a duplicate of the first) and passes with the change. There's a CHANGELOG entry under Unreleased / Fixed.

----
📚 Documentation preview 📚: https://earthaccess--1373.org.readthedocs.build/en/1373/


chuckwondo:
Thanks @arpitjain099. Apologies for the delay in looking at your contribution. Your fix looks great, so I kicked off the blocked integration tests.  The int. tests are failing for me as well, even though I'm a maintainer, but I see that is due to a change made to the github checkout action, so I must open a separate PR to fix that first. Then I should be able to unblock this PR, as well as any others that might be stuck for the same reason.

arpitjain099:
> Thanks @arpitjain099. Apologies for the delay in looking at your contribution. Your fix looks great, so I kicked off the blocked integration tests. The int. tests are failing for me as well, even though I'm a maintainer, but I see that is due to a change made to the github checkout action, so I must open a separate PR to fix that first. Then I should be able to unblock this PR, as well as any others that might be stuck for the same reason.

Thank you @chuckwondo for taking the time to review the PR.

chuckwondo:
@arpitjain099, I fixed the CI issue and updated your branch with the fix. I now get int. tests to pass, so I'm going to approve this PR. Thanks again for your contribution!

chuckwondo (approved):
Thanks @arpitjain099!

arpitjain099:
> @arpitjain099, I fixed the CI issue and updated your branch with the fix. I now get int. tests to pass, so I'm going to approve this PR. Thanks again for your contribution!

@chuckwondo thank you for fixing CI and merging the PR.

PR #1403, opened 2026-07-25
Enable PT (flake8-pytest-style) ruleset and fix violations.

Sherwin-14:
Fixed errors and violations as a result of enabling PT rulesets. Related to https://github.com/earthaccess-dev/earthaccess/issues/383.

----
📚 Documentation preview 📚: https://earthaccess--1403.org.readthedocs.build/en/1403/


mfisher87:
The linter is still unhappy with this PR! :heart:

Sherwin-14:
@mfisher87 I think this is ready to be reviewed. What do you think?

tests/unit/test_results.py
```diff
@@ -166,11 +166,10 @@ def test_get_all_less_than_2k(self):
             count=2000,
         )
 
-        # Assert that we performed a hits query, a results query, and a final
-        # query that returns an empty page (to detect the end of the results).
-        self.assertEqual(len(self.cassette), 3)
-        self.assertEqual(len(granules), 163)
-        self.assertTrue(unique_results(granules))
+        # Assert that we performed a hits query and one search results query
+        assert len(self.cassette) == 2
```
mfisher87:
```suggestion
        assert len(self.cassette) == 3
```

tests/unit/test_results.py
```diff
@@ -182,18 +181,15 @@ def test_get_all_more_than_2k(self):
...
         )
 
-        # Assert that we performed a hits query, two results queries, and a
-        # final query that returns an empty page.
-        self.assertEqual(len(self.cassette), 4)
-        self.assertEqual(
-            len(granules),
-            int(self.cassette.responses[0]["headers"]["CMR-Hits"][0]),
+        # Assert that we performed a hits query and two search results queries
+        assert len(self.cassette) == 3
```
mfisher87:
```suggestion
        assert len(self.cassette) == 4
```

tests/unit/test_results.py
mfisher87: I don't understand why this value would change!

tests/unit/test_store.py
```diff
@@ -406,9 +412,7 @@ def test_sibling_tempfile_error(tmp_path):
         pytest.raises(Exception, match="Some error to trigger cleanup"),
         _sibling_tempfile(trg_file) as temp_file,
     ):
-        temp_file.write_text(new_text)
-        msg = "Some error to trigger cleanup"
-        raise RuntimeError(msg)
+        _write_then_raise(temp_file, new_text, msg="Some error to trigger cleanup")
```
mfisher87, line 415: I'm confused about why this was extracted to a function! It's only two lines and only used in one place.

pyproject.toml
```diff
@@ -211,13 +211,7 @@ ignore = [
...
   "E501",
   "FIX002",
-  "PT006",
-  "PT007",
-  "PT009",
-  "PT011",
-  "PT012",
-  "PT017",
-  "PT027",
+  "PGH003",
```
mfisher87: Why did this need to be added back? We shouldn't be doing blanked type-ignores and I didn't see any in this PR!

pyproject.toml
```diff
@@ -230,7 +224,7 @@ combine-as-imports = true
 [tool.ruff.lint.per-file-ignores]
 # Allow `print` and `pprint` in notebooks
 "*.ipynb" = ["T201", "T203", "N803", "RET504", "ERA001", "PLR2004", "PD002", "PD010", "PD011", "EM", "ARG001", "BLE001", "FBT","TRY003"]
-"tests/**/*.py" = ["PLR2004", "FBT", "INP001", "SLF001", "S101", "S108", "S301", "S603", "S607"]
+"tests/**/*.py" = ["PLR2004", "FBT", "INP001", "SLF001", "S101", "S108", "S301", "S603", "S607", "S105"]
```
mfisher87, line 226: Why did this ignore need to be added?

mfisher87 (changes requested):

tests/unit/test_store.py
Sherwin-14: PT012 requires the pytest.raises() block to contain a single statement. Since we needed multiple statements before the raise, I pulled them into a helper function and call it as the single statement inside the block.

pyproject.toml
Sherwin-14: Oops, it seems like I added this accidentally while resolving some merge conflicts that occurred a month ago. I have removed it.

pyproject.toml
Sherwin-14:
```
S105 Possible hardcoded password assigned to: "password"
  --> tests/unit/test_auth.py:75:33
   |
73 |         auth.login(strategy="interactive")
74 |         assert auth.authenticated is True
75 |         assert auth.password == "password"
   |                                 ^^^^^^^^^^
76 |         assert auth.token == json_response
   |

```
I got the above error on rerunning ruff on my local machine, and on reading more about [S105](https://docs.astral.sh/ruff/rules/hardcoded-password-string/) I realized it is a false positive and added it to the ignore list.

PR #1187, opened 2026-01-29
Add methods to results

andypbarrett:
This PR modifies `DataCollection` and `DataGranule` in the following ways.

Addresses #999 
- Attribute-like methods have `@property` decorators and behave like attributes.
- `DataCollection.summary` now includes collection name (`umm.EntryTitle`)
- `DataCollection.summary` returns file_format rather than full data_type entry
- Add `DataCollection.file_format` to return list of file formats in collection
- Modify `DataCollection.data_type` to return list of dicts instead of string
- Adds migration guide to release notes
- Updates docs

<details><summary>Pull Request (PR) draft checklist - click to expand</summary>

- [x] Please review our
      [contributing documentation](https://earthaccess.readthedocs.io/en/latest/contributing/)
      before getting started.
- [x] Populate a descriptive title. For example, instead of "Updated README.md", use a
      title such as "Add testing details to the contributor section of the README".
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [x] Populate the body of the pull request with:
    - A clear description of the change you are proposing.
    - Links to any issues resolved by this PR with text in the PR description, for
      example `closes #1`. See
      [GitHub docs - Linking a pull request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
- [ ] Update `CHANGELOG.md` with details about your change in a section titled
      `## Unreleased`. If such a section does not exist, please create one. Follow
      [Common Changelog](https://common-changelog.org/) for your additions.
      Example PRs: [#763](https://github.com/nsidc/earthaccess/pull/763)
- [ ] Update the documentation and/or the `README.md` with details of changes to the
      earthaccess interface, if any. Consider new environment variables, function names,
      decorators, etc.

</details>

<details><summary>Pull Request (PR) merge checklist - click to expand</summary>

Please do your best to complete these requirements! If you need help with any of these
requirements, you can ping the `@nsidc/earthaccess-support` team in a comment and we
will help you out!

- [ ] Add unit tests for any new features.
- [ ] Apply formatting and linting autofixes. You can add a GitHub comment in this Pull
      Request containing "pre-commit.ci autofix" to automate this.
- [ ] Ensure all automated PR checks (seen at the bottom of the "conversation" tab) pass.
- [ ] Get at least one approving review.

</details>

----
📚 Documentation preview 📚: https://earthaccess--1187.org.readthedocs.build/en/1187/

PR #1420, opened 2026-08-04
Fix deprecation warning

### LINKED ISSUE #1301: Deprecation warning showing from internal call to DataGranule.size()
This deprecation warning is being shown from an internal call to DataGranule.size(). The warning should only display for the user when they are calling that function. 

Image

andypbarrett:
## Description

closes #1301 

Extracts code from `results.DataGranules.size` to a private property `results.DataGranules._size`.
Deprecation warning is only in `results.DataGranules.size` so only appears when `.size()` is called directly. 
`results.DataGranules.size` then calls `results.DataGranules._size`
Internal calls to get granule size are changed to call `results.DataGranules._size`

These were in:
```
earthaccess/formatters.py:    granule_size = round(granule.size(), 2)
earthaccess/results.py:            "As of version 1.0, `DataGranule.size` will be accessed as an "
earthaccess/results.py:            "attribute; e.g. use `DataCollection.size` **not** "
earthaccess/results.py:            "`DataCollection.size()`",
earthaccess/store.py:        total_size = round(sum([granule.size() for granule in granules]) / 1024, 2)
earthaccess/store.py:        total_size = round(sum(granule.size() for granule in granules) / 1024, 2)
```
#### Merge checklist

- [x] PR title is descriptive
- [-] If needed, docs and/or `README.md` updated
- [?] If needed, unit tests added *(unsure how? see below!)*
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1420.org.readthedocs.build/en/1420/


mfisher87:
How do y'all feel about actually making the breaking change now? We need to do it before v1.0.0 anyway.

andypbarrett:
That is true.  I am happy to rip the bandaid off.

mfisher87:
I vote in favor of ripping off the band-aid :)

And burning it

:fire: :adhesive_bandage: :fire:

asteiker:
@mfisher87 Can you remind me what making the breaking change means in terms of a new or existing PR? I have lost sight of whether this is already in work or not. Should we still merge this now? I saw [this reviewer comment](https://github.com/pyOpenSci/software-submission/issues/302#issuecomment-5064346023) on our PyOpenSci submission and noted that this may be continuing to have some negative user impact.

mfisher87:
I'm not sure I follow your question. We could do the breaking change in this PR as a means of "fixing" the deprecation warning, and that's what I'm proposing.

>I have lost sight of whether this is already in work or not

I believe it is not. The change is _very_ simple so I think there was no incentive to get started on it ahead of time.



> I saw [this reviewer comment](https://github.com/pyOpenSci/software-submission/issues/302#issuecomment-5064346023) on our PyOpenSci submission and noted that this may be continuing to have some negative user impact.

There are two problems: 1) We have a deprecation warning that gets printed regardless of whether the user calls the deprecated method, because we use the deprecated method internally, and we shouldn't do that. The original intent of this PR is to fix this. 2) Actually replacing the deprecated method (#1421) is a breaking change, and because of the way Python is designed, we can't reasonably have both the deprecated method and the new method exist at the same time (i.e. we can't have a transition period).

We could fix (1) alone and deal with (2) later but I'd instead advocate to take the hit and deal with (2) now. We've had this spammy deprecation warning in the latest version for some time now, folks have seen it. The breaking change is fairly trivial to resolve (users will have to change `granule.size()` to `granule.size` in their code). I think there's minimal/no benefit to waiting longer. We should just be noisy about the breaking change.

andypbarrett:
The `@property` decorator was added in #1187

betolink:
I think we could merge this PR now and open a new one to change it altogether with the rest of the methods that need to be properties. @mfisher87 @andypbarrett ?

mfisher87:
That's fine! I don't have the highest confidence that we'll get the results class change done in the short term. I think we should do a separate PR that literally only adds `@property` decorators and nothing else

betolink (dismissed):
LGTM, let's merge it and fix it with another PR!

betolink (approved):
Approving and merging, sorry about the bad temporary mere conflict.

PR #1428, opened 2026-08-18
Breaking: Migrate many methods to `@property`s

### LINKED ISSUE #1421: Replace "property" methods (like `DataGranule.size()`) with `@property` (`DataGranule.size`)
I think we should knock these out before v1. These methods are currently producing deprecation warnings. These are breaking changes and they're easy to complete. We probably have several methods like this that need to be updated.

mfisher87:
## Description

Resolves #1421 

There's related discussion somewhere but I'm about to go in to a meeting and don't have them on hand. I'll try and edit that in here ASAP.

---

- [x] Please review our [Pull Request Guide](https://earthaccess.readthedocs.io/en/latest/contributor/howto/pr-guide/)

#### Merge checklist

- [x] PR title is descriptive
- [x] If needed, unit tests added *(unsure how? see below!)*
- [x] All checks passing *(tip: comment `pre-commit.ci autofix` if pre-commit is failing)*
- [ ] At least one approval

----
📚 Documentation preview 📚: https://earthaccess--1428.org.readthedocs.build/en/1428/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,10 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
   `granule.size` instead of `granule.size()`. If you use the old syntax, you'll
   receive errors like `TypeError: 'NoneType' object is not callable`,
   `TypeError: 'str' object is not callable`, etc.
-  
+
   The following methods were changed to read-only fields of the same name,
   except where noted:
-  
+
   - `DataCollection.concept_id()` -> `DataCollection.concept_id`
   - `DataCollection.data_type()` -> `DataCollection.data_type`
   - `DataCollection.doi()` -> `DataCollection.doi`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### Breaking changes
+
+- Change many methods to `@property` methods.
+  Now, for example, you'll get the size of a granule with `granule.size` instead of
+  `granule.size()`.
+  If you use the old syntax, you'll receive `TypeError: 'NoneType' object is not callable`.
+  The following methods were changed:
+  - `DataCollection.summary`
+  - `DataCollection.doi`
+  - `DataCollection.concept_id`
+  - `DataCollection.data_type`
+  - `DataCollection.version`
+  - `DataCollection.landing_page`
+  - `DataCollection.s3_bucket`
+  - `DataCollection.services`
+  - `DataGranule.size`
+- Rename `DataCollection.get_links` to `DataCollection.data_links` and make it a
+  `@property`.
+  The link name was changed because the links are called "GET DATA" in the CMR API, and
+  the old method name, while it matches the property name in the API, can be read as a
+  verb, when it's really an attribute.
+
 ### Fixed
 
 - `search_services` now respects the authenticated system (UAT vs PROD) instead
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,25 +9,33 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
 ### Breaking changes
 
-- Many methods are now `@property` methods.
-  For example, you'll get the size of a granule with `granule.size` instead of
-  `granule.size()`.
-  If you use the old syntax, you'll receive errors like `TypeError: 'NoneType' object is not callable`, `TypeError: 'str' object is not callable`, etc.
-  The following methods were changed:
-  - `DataCollection.summary`
-  - `DataCollection.doi`
-  - `DataCollection.concept_id`
-  - `DataCollection.data_type`
-  - `DataCollection.version`
-  - `DataCollection.landing_page`
-  - `DataCollection.s3_bucket`
-  - `DataCollection.services`
-  - `DataGranule.size`
-- `DataCollection.get_links` is renamed to `DataCollection.data_links` and is now a
-  `@property`.
-  The link name was changed because the links are called "GET DATA" in the CMR API, and
-  the old method name, while it matches the property name in the API, can be read as a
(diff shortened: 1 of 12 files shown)


earthaccess/formatters.py
```diff
@@ -37,7 +37,7 @@ def _repr_granule_html(granule: Any) -> str:
             for link in granule.data_links()
```
chuckwondo, line 37: This line is outside of what you changed, but `data_links` is a property, not a method, so we need to drop the parens:

```suggestion
            for link in granule.data_links
```

CHANGELOG.md
```diff
@@ -7,6 +7,28 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
...
+  - `DataCollection.version`
+  - `DataCollection.landing_page`
+  - `DataCollection.s3_bucket`
+  - `DataCollection.services`
+  - `DataGranule.size`
+- `DataCollection.get_links` is renamed to `DataCollection.data_links` and is now a
+  `@property`.
+  The link name was changed because the links are called "GET DATA" in the CMR API, and
+  the old method name, while it matches the property name in the API, can be read as a
+  verb, when it's really an attribute.
```
chuckwondo: I suggest dropping explicit use of `@property` in the description.  The user-facing concern is that they simply use dot notation to access a read-only "field" (whether that's implemented as an attribute or a property is an implementation detail).

```suggestion
- Many `DataCollection` and `DataGranule` methods are now read-only fields.
  For example, you'll get the size of a `DataGranule` named `granule` via
  `granule.size` instead of `granule.size()`. If you use the old syntax, you'll
  receive errors like `TypeError: 'NoneType' object is not callable`,
  `TypeError: 'str' object is not callable`, etc.
  
  The following methods were changed to read-only fields of the same name,
  except where noted:
  
  - `DataCollection.concept_id()`
  - `DataCollection.data_type()`
  - `DataCollection.doi()`
  - `DataCollection.get_links()` (now `DataCollection.data_links`)
  - `DataCollection.landing_page()`
  - `DataCollection.s3_bucket()`
  - `DataCollection.services()`
  - `DataCollection.summary()`
  - `DataCollection.version()`
  - `DataGranule.size()`

  **NOTE:** The method `DataCollection.get_links()` was replaced with the field
  `DataCollection.data_links`. The method name was `get_links` because in the
  CMR API, the links are described as links of type "GET DATA".  However, while
  the old method name matches the name of the link type in the CMR API, it can
  be read as a verb ("get" the links, rather than links of _type_ "get"), when
  it's an attribute (noun).  Thus, the name `data_links` was chosen to avoid
  confusion (noun instead of verb).
```

chuckwondo (dismissed):
Thanks @mfisher87! I flagged 1 bug, and made a suggestion for clarifying the changelog entry.

mfisher87:
Thanks Chuck! I'm going to add one more change to the changelog that I think will help.

CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
...
+  - `DataCollection.concept_id()`
+  - `DataCollection.data_type()`
+  - `DataCollection.doi()`
+  - `DataCollection.get_links()` (now `DataCollection.data_links`)
+  - `DataCollection.landing_page()`
+  - `DataCollection.s3_bucket()`
+  - `DataCollection.services()`
+  - `DataCollection.summary()`
+  - `DataCollection.version()`
+  - `DataGranule.size()`
```
mfisher87:
```suggestion
  - `DataCollection.concept_id()` -> `DataCollection.concept_id`
  - `DataCollection.data_type()` -> `DataCollection.data_type`
  - `DataCollection.doi()` -> `DataCollection.doi`
  - `DataCollection.get_links()` -> `DataCollection.data_links` (note the name change, details below)
  - `DataCollection.landing_page()` -> `DataCollection.landing_page`
  - `DataCollection.s3_bucket()` -> `DataCollection.s3_bucket`
  - `DataCollection.services()` -> `DataCollection.services`
  - `DataCollection.summary()` -> `DataCollection.summary`
  - `DataCollection.version()` -> `DataCollection.version`
  - `DataGranule.size()` -> `DataGranule.size`
```

mfisher87:
pre-commit.ci autofix

mfisher87:
Hm. I thought your fix was correct, Chuck, but looks like a bit more work needs to be done. I'll look at it now.


```
  File "/home/runner/work/earthaccess/earthaccess/earthaccess/formatters.py", line 37, in _repr_granule_html
    for link in granule.data_links
                ^^^^^^^^^^^^^^^^^^
TypeError: 'method' object is not iterable
```

mfisher87:
OK, `DataGranule.data_links` can't be a property because it has args. This is annoying because `DataCollection.data_links` is now a property, and `DataGranule.data_links` is not. I think that's the right move for now...?

chuckwondo (dismissed):
Thanks @mfisher87!

I see that I was looking at `DataCollection.data_links` (a property), not `DataGranule.data_links` (a method), so thanks for sorting that out.

For future work (not this PR), perhaps we should consider splitting things for `DataGranule` into 2 properties to avoid the method/property inconsistency: `s3_data_links` and `https_data_links`.

That would eliminate the need for the `access` parameter (and the less-than-obvious/helpful "direct" vs "external" values), and the `in_region` parameter should be dropped without an alternative (i.e., just give the user what they ask for without attempting error-prone in-region detection, as we've discussed before).

mfisher87:
> For future work (not this PR), perhaps we should consider splitting things for `DataGranule` into 2 properties to avoid the method/property inconsistency: `s3_data_links` and `https_data_links`.
> 
> That would eliminate the need for the `access` parameter (and the less-than-obvious/helpful "direct" vs "external" values), and the `in_region` parameter should be dropped without an alternative (i.e., just give the user what they ask for without attempting error-prone in-region detection, as we've discussed before).

I think that sounds like a good idea. I would prefer `data_links_https` and `data_links_s3` though just to keep these two methods together when a user `dir()`s an object.

PR #1460, opened 2026-09-05
Enable D401 and D417 rulesets and fix violations

Sherwin-14:
Fixed errors and violations as a result of enabling D401 and D417 rulesets barring D1 and D205.  Related to https://github.com/earthaccess-dev/earthaccess/issues/383.


chuckwondo (approved):
Thanks @Sherwin-14!

